PR #48, opened 2016-08-05
ENH: Add benchmarking with Airspeed Velocity

bocklund:
[Airspeed Velocity (asv)](https://github.com/spacetelescope/asv) is a Python benchmarking tool that can benchmark build matrices to compare the performance of pycalphad over time. The benchmarks here have been confirmed to run using asv 0.2, currently available on GitHub. 

There is a .asv/ directory that is generated during benchmarking that is not currently in the pull request because the benchmarks won't run on my machine (Mac OS X) without customizing the environments so matplotlib plays nicely and the benchmarks on my Linux vm are not representative of performance.


README.rst
```diff
@@ -33,6 +33,9 @@ pycalphad, a library for the CALculation of PHAse Diagrams
     :target: https://pypi.python.org/pypi/pycalphad/
     :alt: License
 
+.. image:: http://img.shields.io/badge/benchmarked%20by-asv-green.svg
+    :target: https://github.com/spacetelescop/asv
```
richardotis: Typo in the URL here

benchmarks/benchmarks.py
```diff
@@ -0,0 +1,524 @@
...
+
+    def setup(self):
+        # create a file to read from
+        with open('alfe_sei.TDB', 'w') as f:
+            f.write(ALFE_TDB)
+        # setup TDBs
+        self.tdb_file =  open('alfe_sei.TDB', 'r')
+        self.tdb_string = ALFE_TDB
+        fec_tdb = FEC_TDB
+        # setup databases
```
richardotis: Per offline conversation, we do this because making a Database from a file on disk follows a different code path from a string, so this difference may be worth capturing in our benchmark.

richardotis:
It  would be good to benchmark Model construction for phases with and without magnetic contributions; the magnetic contribution tends to be trickier, in general. B2_BCC has a magnetic contribution, but phases like LIQUID and simple stoichiometric compounds do not.

richardotis:
Can you add some instructions on how to update the benchmarks as additional steps in RELEASING.rst?

richardotis:
For only running tagged commits, maybe: https://www.kernel.org/pub/software/scm/git/docs/gitrevisions.html

bocklund:
`Can you add some instructions on how to update the benchmarks as additional steps in RELEASING.rst? `

Do you mean how to run the benchmarks for a new release?

richardotis:
Yes.

richardotis:
@bocklund Just go ahead and make the modifications to `.gitignore` we talked about directly to develop on the main repo. No reason to delay the merge.

PR #61, opened 2016-12-13
Add Binder link to README.rst

bocklund:
I was able to go into the examples and run the BinaryExamples in Binder, but it was not picking up and installing pycalphad correctly. There are probably better ways to fix this so that docker detects and installs correctly, but running the following command in the Jupyter Notebook to install it worked fine.
`! conda install -y -p /home/main/anaconda2/envs/python3 -c conda-forge pycalphad`

The performance isn't bad either. I was able to run the Al-Zn example in about 2 minutes on my i7 MacBook and it took about 5 to run in the Docker instance.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -36,6 +36,10 @@ pycalphad, a library for the CALculation of PHAse Diagrams
 .. image:: http://img.shields.io/badge/benchmarked%20by-asv-green.svg
     :target: https://github.com/spacetelescope/asv
 
+.. image:: http://mybinder.org/badge.svg 
+    :target: http://mybinder.org:/repo/pycalphad/pycalphad
+    :alt: Binder
+
 **Note**: Unsolicited pull requests are _happily_ accepted!
 
 pycalphad is a free and open-source Python library for 


richardotis:
This is awesome! I didn't realize it could be this simple to get an interactive example going. The JIT compiler / code generator is going to be an issue for online examples; hopefully there will be some time to improve its performance in the future. (It's hard to justify spending time on it for high-throughput modeling since you pay the codegen cost up front and it's small compared to the total time, but for one-off calculations it's definitely a drag on user experience.)

Is there a way to tell Binder a custom conda environment so that users don't have to type the install command to run the examples?

bocklund:
There are multiple ways (see [here](http://docs.mybinder.org/getting-started)). For anaconda, we should be able to add an `environment.yml`. We could also make a `requirements.txt` or a `Dockerfile`. 

The shell command is easy and avoids adding another file to maintain to the project, but having it install automatically would be nice as well.

bocklund:
Closing this since the pycalphad org has a [repo for Binder](https://github.com/pycalphad/binder) now

Issue #66, opened 2017-01-31
Python 3.6 support


bocklund:
Test are failing for the newly released Python 3.6.

A couple are due to _io.StringIO TypeErrors, but most are `_solve_eq_at_conditions() takes exactly 7 positional arguments (8 given)`

```
======================================================================
ERROR: from_string: Unspecified string format raises ValueError.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/brandon/.virtualenvs/pycalphad36/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/brandon/.virtualenvs/pycalphad36/lib/python3.6/site-packages/nose/tools/nontrivial.py", line 60, in newfunc
    func(*arg, **kw)
  File "/Users/brandon/Projects/pycalphad/pycalphad/tests/test_database.py", line 74, in test_unspecified_format_from_string
    Database.from_string(ALCRNI_TDB)
  File "/Users/brandon/Projects/pycalphad/pycalphad/io/database.py", line 234, in from_string
    return cls.from_file(StringIO(data), **kwargs)
  File "/Users/brandon/Projects/pycalphad/pycalphad/io/database.py", line 186, in from_file
    path, ext = os.path.splitext(fname)
  File "/Users/brandon/.virtualenvs/pycalphad36/bin/../lib/python3.6/posixpath.py", line 120, in splitext
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not _io.StringIO

======================================================================
ERROR: from_file: Unspecified format for file descriptor raises ValueError.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/brandon/.virtualenvs/pycalphad36/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/brandon/.virtualenvs/pycalphad36/lib/python3.6/site-packages/nose/tools/nontrivial.py", line 60, in newfunc
    func(*arg, **kw)
  File "/Users/brandon/Projects/pycalphad/pycalphad/tests/test_database.py", line 94, in test_unspecified_format_from_file
    Database.from_file(StringIO(ALCRNI_TDB))
  File "/Users/brandon/Projects/pycalphad/pycalphad/io/database.py", line 186, in from_file
    path, ext = os.path.splitext(fname)
[...]

PR #67, opened 2017-02-02
BLD/TST: Add support for Python 3.6

bocklund:
**Please check that I updated the Travis and Appveyor files correctly.** 

Closes gh-66.

Add Python 3.6 to Travis and Appveyor

Python 3.6 changed `splitext` to support path-like objects, so passing a StringIO
(file-like) now throws a `TypeError` instead of an `AttributeError`. This was
changed at 3 instances in the format determination.

A test was also written to check that passing a StringIO directly to the Database
constructor (e.g. not using the static method constructors) will proceed
correctly through the try-except block. It's the same test as from_file, it just
did not exist for the direct __new__ constructor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,9 @@ matrix:
         - python: 3.4
           env: DOCKER_TAG=linux-python34
         - python: 3.5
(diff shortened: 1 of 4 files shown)


richardotis:
Looks good. I'm going to have to build a Python 3.6 Docker image before the Travis tag will work. Once I do that, it should be mergeable.

Issue #65, opened 2017-01-13
autowrap fails to import compiled files on macOS


bocklund:
Using `GNU Fortran (Homebrew GCC 6.3.0_1) 6.3.0` compiler on macOS 10.12.2  results in failures to import compiled Fortran files in autowrap. The problem seems to be that the compiler flags passed with the `f2py`backend in custom_autowrap.autowrap generate debug symbol files (`.dSYM`) files alongside the compiled files. 

The quick fix is to just filter out those files, e.g. with a blacklist by appending 
```
    blacklist = ['dSYM']
    npath = [x for x in npath if x.split('.')[-1] not in blacklist]
```
after line 101. 

A more permanent solution would be to explicitly turn off such compiler flags and write a test to confirm. Currently, the tests fail on the above machine because of this bug, but this does not appear to be getting caught by CI.

richardotis:
I'm fine with either approach for patching. We really need to get the CI issues worked out so that there are reliable tests again.

bocklund:
I do want to mention here (so we can reference this) that a viable alternative to a blacklist would be a whitelist with valid file extensions. In principle, whichever list is smaller should be easier to maintain. Here I think excluding special cases is better than including normal cases. Not filtering out files correctly (`len(path)>1`) will raise an error on import, so either approach is technically fine.

PR #70, opened 2017-03-23
FIX: Create a blacklist for compiler files

bocklund:
Compiler settings on certain machines can create files such as debug
symbols. This commit creates a blacklist for filtering those files out,
leaving only the compiled file. We cannot more directly find the compiled
file because only the extensions differ (e.g. '*.so' and '*.so.dysm') and
the extensions of the compiled files change on different platforms.

Fixes gh-65

### DIFF
--- a/pycalphad/core/custom_autowrap.py
+++ b/pycalphad/core/custom_autowrap.py
@@ -99,6 +99,11 @@ def _infer_language(backend):
 def import_extension(path, modname):
     import glob
     npath = glob.glob(os.path.join(path, modname+'.*'))
+    # Blacklist fixes gh-65.
+    # We filter out any files that can be created by compilers which are not our actual compiled file.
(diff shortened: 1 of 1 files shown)


richardotis:
No problems with this, two comments:
(1) Add a short comment pointing to this issue so we don't forget why a blacklist is necessary. You can include your explanatory text there too, if you want.
(2) Any thoughts on a new test? I'm still confused why the OSX tests on Travis have missed this.

bocklund:
(1) I'll add a short comment and amend the commit. 
(2) I believe the tests miss this because of some kind of local compiler setting. Since I'm using a vanilla compiler straight from Homebrew, I assume I wouldn't be the only person with this problem. The compiler used by Travis may be different. The current test suite fails for me without this fix, I've just taken awhile to get around to committing it.

richardotis:
LGTM, merge when ready.

PR #71, opened 2017-03-27
TC TDB Compatibility (symbol length)

bocklund:
This limitation and other limitations as a subset of the different TDB implementations should be handled by pycalphad so that any output database can work on any thermodynamics software. Some of these limitations can be handled automatically, as in the case of line length limits, others cannot. These changes typically break `DB == DB.read(DB.write())`, so they require explicit handling from the user. 

Here, Thermo-Calc TDBs require symbols to be <= 8 characters

This PR:
- Provides a reasonably scalable (rough) framework to optionally sanitize TDBs for compatibility with the following options in the case of an invalid database
    - 'raise' (default): raise an error if the TDB is invalid and cannot be automatically compatible 
    - 'warn': Issue a warning that the database is invalid and write it as is
    - 'ignore': Silently write the invalid database as is
    - 'fix': Make any changes required for compatibility
- Implement optional name mangling of long symbol names to 8 characters using md5 hashing
- Adds documentation to `io.tdb.write_tdb` to clarify the goals and functionality
- Includes tests for the newly added documentation and updates old tests to conform.

Future features:
- Use and improve the framework to add more compatibility fixes (listed in the docstring)
- (Possibly) Add the ability to reconstitute the original database in cases where names have been mangled

### DIFF
--- a/pycalphad/io/tdb.py
+++ b/pycalphad/io/tdb.py
@@ -14,6 +14,7 @@ from sympy.printing.str import StrPrinter
 from sympy.core.mul import _keep_coeff
 from sympy.printing.precedence import precedence
 from pycalphad import Database
+from pycalphad.io.database import DatabaseExportError
 import pycalphad.variables as v
 from pycalphad.io.tdb_keywords import expand_keyword, TDB_PARAM_TYPES
 from collections import defaultdict, namedtuple
@@ -24,6 +25,9 @@ import functools
 import itertools
 import getpass
 import datetime
+import warnings
+import hashlib
+from copy import deepcopy
 
 _AST_WHITELIST = [ast.Add, ast.BinOp, ast.Call, ast.Div, ast.Expression,
                   ast.Load, ast.Mult, ast.Name, ast.Num, ast.Pow, ast.Sub,
@@ -563,9 +567,46 @@ def reflow_text(text, linewidth=80):
     return "\n".join(output_lines)
 
 
-def write_tdb(dbf, fd, groupby='subsystem'):
+def _apply_new_symbol_names(dbf, symbol_name_map):
+    """
+    Push changes in symbol names through the Sympy expressions in symbols and parameters
+
+    Parameters
+    ----------
+    dbf : Database
+        A pycalphad Database.
+    symbol_name_map : dict
+        Map of {old_symbol_name: new_symbol_name}
+    """
+    # first apply the rename to the keys
+    dbf.symbols = {symbol_name_map.get(name, name): expr for name, expr in dbf.symbols.items()}
+    # then propagate through to the symbol SymPy expression values
+    dbf.symbols = {name: S(expr).xreplace({Symbol(s): Symbol(v) for s, v in symbol_name_map.items()}) for name, expr in dbf.symbols.items()}
+    # finally propagate through to the parameters
+    for p in dbf._parameters.all():
+        dbf._parameters.update({'parameter': S(p['parameter']).xreplace({Symbol(s): Symbol(v) for s, v in symbol_name_map.items()})}, eids=[p.eid])
+
+
+def write_tdb(dbf, fd, groupby='subsystem', if_incompatible='warn'):
     """
     Write a TDB file from a pycalphad Database object.
+
+    The goal is to produce TDBs that conform to the most restrictive subset of database specifications. Some of these
+    can be adjusted for automatically, such as the Thermo-Calc line length limit of 78. Others require changing the
+    database in non-trivial ways, such as the maximum length of function names (8). The default is to warn the user when
+    attempting to write an incompatible database and the user must choose whether to warn and write the file anyway or
+    to fix the incompatibility.
+
+    Currently the supported compatibility fixes are:
+    - Line length <= 78 characters (Thermo-Calc)
+    - Function names <= 8 characters (Thermo-Calc)
+
+    The current unsupported fixes include:
+    - Keyword length <= 2000 characters (Thermo-Calc)
+    - Element names <= 2 characters (Thermo-Calc)
+    - Phase names <= 24 characters (Thermo-Calc)
+
+    Other TDB compatibility issues required by Thermo-Calc or other software should be reported to the issue tracker.
 
     Parameters
     ----------
@@ -575,7 +616,32 @@ def write_tdb(dbf, fd, groupby='subsystem'):
         File descriptor.
     groupby : ['subsystem', 'phase'], optional
         Desired grouping of parameters in the file.
+    if_incompatible : string, optional ['raise', 'warn', 'fix']
+        Strategy if the database does not conform to the most restrictive database specification.
+        The 'warn' option (default) will write out the incompatible database with a warning.
+        The 'raise' option will raise a DatabaseExportError.
+        The 'ignore' option will write out the incompatible database silently.
+        The 'fix' option will rectify the incompatibilities e.g. through name mangling.
     """
+    # Before writing anything, check that the TDB is valid and take the appropriate action if not
+    if if_incompatible not in ['warn', 'raise', 'ignore', 'fix']:
+        raise ValueError('Incorrect options passed to \'if_invalid\'. Valid args are \'raise\', \'warn\', or \'fix\'.')
+    # Handle function names > 8 characters
+    long_function_names = {k for k in dbf.symbols.keys() if len(k) > 8}
+    if len(long_function_names) > 0:
+        if if_incompatible == 'raise':
+            raise DatabaseExportError('The following function names are beyond the 8 character TDB limit: {}. Use the keyword argument \'if_incompatible\' to control this behavior.'.format(long_function_names))
+        elif if_incompatible == 'fix':
+            # if we are going to make changes, make the changes to a copy and leave the original object untouched
+            dbf = deepcopy(dbf) # TODO: if we do multiple fixes, we should only copy once
+            symbol_name_map = {}
+            for name in long_function_names:
+                hashed_name = 'F' + str(hashlib.md5(name.encode('UTF-8')).hexdigest()).upper()[:7] # this is implictly upper(), but it is explicit here
+                symbol_name_map[name] = hashed_name
+            _apply_new_symbol_names(dbf, symbol_name_map)
+        elif if_incompatible == 'warn':
+            warnings.warn('Ignoring that the following function names are beyond the 8 character TDB limit: {}. Use the keyword argument \'if_incompatible\' to control this behavior.'.format(long_function_names))
+    # Begin constructing the written database
     writetime = datetime.datetime.now()
     maxlen = 78
     output = ""
--- a/pycalphad/tests/test_database.py
+++ b/pycalphad/tests/test_database.py
@@ -2,9 +2,14 @@
 The test_database module contains tests for the Database object.
 """
 from __future__ import print_function
+import warnings
+import hashlib
+from copy import deepcopy
 from pyparsing import ParseException
-from pycalphad import Database, Model
+from sympy import Symbol, Piecewise, And
+from pycalphad import Database, Model, variables as v
 from pycalphad.io.tdb import expand_keyword
+from pycalphad.io.tdb import _apply_new_symbol_names, DatabaseExportError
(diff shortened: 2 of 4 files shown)


bocklund:
I think I've gotten ahead of myself and didn't propagate the changed symbol names through the other expressions with the fix option. Best way to do that?

richardotis:
First, I appreciate the thorough job you've done here, both with documenting the concern with cross-software TDB compatibility, as well as the fix and tests. A few comments:
(1) I'd like to bikeshed on the `if_invalid` kwarg name for a bit. Closest I can come up with is `if_incompat` but I'm concerned that that's an opaque name for new users. The TDB isn't invalid, exactly, it just isn't universally compatible.
(2) The 'warn' mode needs a more descriptive warning message; it should give the same information on the long function names as the ValueError.

For propagating the changed names, I recommend the following approach:
(1) Create a dict, `long_name_dict` mapping the old (long) name to the mangled name.
(2) Create an empty dict, `new_symbols`. For each key in the `symbols` dict, if the key is in `long_name_dict`, write it to `new_symbols` using the mangled name as the key, otherwise write it to `new_symbols` using the existing name as the key.
(3) `new_symbols = {key: sympy.S(val).xreplace({Symbol(s): Symbol(v) for s, v in long_name_dict.items()}) for key, val in new_symbols.items()}` (or similar)

bocklund:
(1) I agree. I think the full 'is_incompatible' is fine. It's only 5 more characters and more explicit about the intent than either shorter option.

(2) Changed locally. I wanted to try to gather all warnings up and issue just one warning so I each warning will add to the warning string and then issue just one warning with all of the incompatibilities.

This pull and further changes mutate the database 
```
dbf_orig = copy(dbf)
dbf.write()
dbf == dbf_orig  # evaluates to False
```

Is that intended for when we do some calculations, write the database and do some more calculations? I'm thinking forward, such as in the case where we want to try to do name mangling of phase names that are too long (regardless of whether or not we *should* do that). Should we instead make a copy of dbf first if the argument is fix and then write that? For example, currently:

```
test_dbf = Database.from_string(INVALID_TDB_STR, fmt='tdb')
mangled_dbf = Database.from_string(test_dbf.to_string(fmt='tdb', if_incompatible='fix'), fmt='tdb')
test_dbf == mangled_dbf # evaluates to True
```

richardotis:
Writing out a "fixed" database should never mutate the original. That should be one of the tests.

bocklund:
I have two ideas for a deep copy that prevents that. Either 

1. Check if the dbf  is the original database for each fix. If True, make a deep copy and apply the changes. If False, apply the changes. 

2. Gather a list of the fixes as callables and apply them all to a copy at the end


1. Becomes dirtier than 2. as the number of fixes gets larger. What do you think?

pycalphad/io/tdb.py
```diff
@@ -575,7 +594,31 @@ def write_tdb(dbf, fd, groupby='subsystem'):
...
+            raise ValueError('The following functions names are beyond the 8 character TDB limit: {}. Use the kwarg \'if_invalid\' to control this behavior.'.format(long_function_names))
+        elif if_invalid == 'fix':
+            for name in long_function_names:
+                hashed_name = 'F' + str(hashlib.md5(name.encode('UTF-8')).hexdigest()).upper()[:7] # this is implictly upper(), but it is explicit here
+                dbf.symbols[hashed_name] = dbf.symbols.pop(name)
+        elif if_invalid == 'warn':
+            warn_invalid_tdb = True
+    if warn_invalid_tdb:
+        warnings.warn('Writing an incompatible TDB.')
+    # Begin contstructing the written database
```
richardotis: typo in comment

richardotis:
If you're in 'fix' mode, just rebind the `dbf` name to a deepcopy of itself: `dbf = deepcopy(dbf)`. Then mutate `dbf` (which no longer points to your original object) to your heart's content and you can leave the rest of the code unchanged.

pycalphad/io/tdb.py
bocklund: Thanks

bocklund:
I think I've got this now. Let me do the whole test suite locally then I'll push

bocklund:
@richardotis Pushed fixes for the above comments:

- Not mutating a database on write+fix (deepcopy)
- Change `if_invalid` to `if_incompatible`
- A function to apply symbol changes to the symbol names, symbol expressions, and parameter expressions
- Improve tests to include the above

One weakness of my tests is that I don't actually test fixes on a complete database. In principle, based on the tests it should work, but do you think I should do a full scale test to read a database, then create a new database from fixed and then do a calculation? I could modify an existing DB to be incompatible, but I don't have a good example for this that I can publicly share.

pycalphad/tests/test_model.py
```diff
@@ -35,7 +35,7 @@ def test_model_ne():
 
 def test_export_import():
     "Equivalence of Model using re-imported database."
-    test_model = Model(Database.from_string(ALNIPT_DBF.to_string(fmt='tdb'), fmt='tdb'), ['PT', 'NI', 'VA'], 'FCC_L12')
+    test_model = Model(Database.from_string(ALNIPT_DBF.to_string(fmt='tdb', if_incompatible='ignore'), fmt='tdb'), ['PT', 'NI', 'VA'], 'FCC_L12')
```
richardotis: Does this database contain a long symbol name, so now the kwarg is required? I'm slightly concerned about regressions now. Should `warn` be the default setting instead, at least for a while?

bocklund: Yes, this one does and so does one of the others in the `test_database.py`. You could change to pass warn or raise and see the warning for the names. Do you care to elaborate on what regressions you're expecting? If you are more comfortable with warn being the default, that's fine.

pycalphad/tests/test_database.py
```diff
@@ -64,9 +77,59 @@ def test_load_from_string():
...
+def test_tc_incompatible_ignores_with_kwarg_ignore():
+    "Symbol names too long for Thermo-Calc are ignored the database written as given."
+    test_dbf = Database.from_string(INVALID_TDB_STR, fmt='tdb')
+    with warnings.catch_warnings(record=True) as w:
+        invalid_dbf = test_dbf.to_string(fmt='tdb', if_incompatible='ignore')
+        assert len(w) == 0
+    assert test_dbf == Database.from_string(invalid_dbf, fmt='tdb')
+
+def test_tc_incompatible_mangles_names_with_kwarg_fix():
+    "Symbol names too long for Thermo-Calc are mangled and put in a JSON in the same dir as the TDB."
```
bocklund: The JSON is some cruft and should be removed. I was thinking that a possible solution would be to write a JSON file with the mangled names (e.g. symbol_name_ map to a JSON in the same directory as the TDB) then we could reconstitute with possibly a special keyword arg, but I didn't implement it. Another idea I had for reconstituting was to write somewhere constant relative to the package/os, but that wouldn't be portable. 

Ideally we could just use some comments, but comments are filtered in the read_tdb, so we can't automatically constitute.

pycalphad/io/tdb.py
```diff
@@ -575,7 +615,32 @@ def write_tdb(dbf, fd, groupby='subsystem'):
         File descriptor.
     groupby : ['subsystem', 'phase'], optional
         Desired grouping of parameters in the file.
+    if_incompatible : string, optional ['raise', 'warn', 'fix']
+        Strategy if the database does not conform to the most restrictive database specification.
+        The 'raise' option (default) will raise a ValueError.
```
bocklund: Should other write-related `ValueError` exceptions also use this?

pycalphad/io/tdb.py
```diff
@@ -575,7 +615,32 @@ def write_tdb(dbf, fd, groupby='subsystem'):
...
+        The 'fix' option will rectify the incompatibilities e.g. through name mangling.
     """
+    # Before writing anything, check that the TDB is valid and take the appropriate action if not
+    if if_incompatible not in ['raise', 'warn', 'ignore', 'fix']:
+        raise ValueError('Incorrect options passed to \'if_invalid\'. Valid args are \'raise\', \'warn\', or \'fix\'.')
+    # Handle function names > 8 characters
+    long_function_names = {k for k in dbf.symbols.keys() if len(k) > 8}
+    if len(long_function_names) > 0:
+        if if_incompatible == 'raise':
+            raise ValueError('The following function names are beyond the 8 character TDB limit: {}. Use the kwarg \'if_invalid\' to control this behavior.'.format(long_function_names))
```
bocklund: Fixed in 294a4fe

pycalphad/io/tdb.py
```diff
@@ -563,10 +566,47 @@ def reflow_text(text, linewidth=80):
...
+    Currently the supported compatibility fixes are:
+    - Line length <= 78 characters (Thermo-Calc)
+    - Function names <= 8 characters (Thermo-Calc)
+
+    The current unsupported fixes include:
+    - Keyword length <= 2000 characters (Thermo-Calc)
+    - Element names <= 2 characters (Thermo-Calc)
+    - Phase names <= 24 characters (Thermo-Calc)
+
+    Other TDB compatibility issues required by Thermo-Calc or other software should be reported to the issue tracker
```
bocklund: Fixed in 294a4fe

pycalphad/tests/test_database.py
```diff
@@ -2,9 +2,14 @@
 The test_database module contains tests for the Database object.
 """
 from __future__ import print_function
+import warnings
+import hashlib
+from copy import deepcopy
 from pyparsing import ParseException
-from pycalphad import Database, Model
+from sympy import Symbol, Piecewise, And
+from pycalphad import Database, Model, variables
```
bocklund: Fixed in 294a4fe

pycalphad/tests/test_database.py
bocklund: Fixed in f6b7711

pycalphad/tests/test_model.py
bocklund: This call is changed to `warn` in 294a4fe. `raise` is still the default for `write_tdb`.

richardotis: `if_incompatible=ignore` is fine here since the warning is expected and safe to ignore; I'm just wondering about the default setting. If someone has their own TDB cleaning/whatever script for personal use and it happens to deal with TDBs with long symbol names, then updating to the new release of pycalphad will break their workflow until they read the documentation. A default setting of `if_incompatible=warn` for a few releases would give them an opportunity to react without the abrupt change.

richardotis: I think the test should use `ignore` and the signature of `write_tdb` should use `warn`.

bocklund: Fixed in  d949079

pycalphad/io/tdb.py
bocklund: @richardotis Just need a response on this last item in this review

pycalphad/io/tdb.py
```diff
@@ -563,10 +567,47 @@ def reflow_text(text, linewidth=80):
...
+        dbf._parameters.update({'parameter': S(p['parameter']).xreplace({Symbol(s): Symbol(v) for s, v in symbol_name_map.items()})}, eids=[p.eid])
+
+
+def write_tdb(dbf, fd, groupby='subsystem', if_incompatible='warn'):
     """
     Write a TDB file from a pycalphad Database object.
 
+    The goal is to produce TDBs that conform to the most restrictive subset of database specifications. Some of these
+    can be adjusted for automatically, such as the Thermo-Calc line length limit of 78. Others require changing the
+    database in non-trivial ways, such as the maximum length of function names (8). The default is to raise an error
```
richardotis: The default is now to warn the user.

pycalphad/io/tdb.py
```diff
@@ -575,7 +616,32 @@ def write_tdb(dbf, fd, groupby='subsystem'):
         File descriptor.
     groupby : ['subsystem', 'phase'], optional
         Desired grouping of parameters in the file.
+    if_incompatible : string, optional ['raise', 'warn', 'fix']
+        Strategy if the database does not conform to the most restrictive database specification.
+        The 'raise' option (default) will raise a DatabaseExportError.
+        The 'warn' option will write out the incompatible database with a warning.
```
richardotis: warn is now the default.

richardotis:
This looks good. Update the docstring with the new default as noted in my comments, squash, rebase and merge.

PR #72, opened 2017-03-28
Linting set and dict comprehensions 

bocklund:
- Changed list comprehensions -> set() to set comprehensions
- Did the same for list comp -> tuple list -> dict dict comprehensions


richardotis:
LGTM, rebase and merge when ready.

PR #74, opened 2017-04-05
Fix filtering of phases with composition >1

bocklund:
As discussed in #4, conditions where the total composition is > 1 should be skipped, but instead were referencing the `phases` variable which does not exist.

Also a small fix some duplicate code in `eqplot` is included because line 58 ensures that `ax` will never be `None` here.


richardotis:
Looks good to me. Can you rewrite the commit messages to have FIX and MAINT prefixes, respectively, before rebasing and merging?

PR #75, opened 2017-04-05
ENH: TDB compatibility: all chars after command delimiters should be neglected

bocklund:
Thermo-Calc will not parse commands or other strings after the command delimiter. Previously pycalphad only removed comments after command delimiters. This commit makes pycalphad conform  to the behavior of parsing out everything on the line after the delimiter and includes a test for this behavior.

Another approach might be to only strip characters after the last delimiter on a line, which would be an overall improved approach in the case that multiple commands were on a single line (or an entire TDB on a single line) even though I have never encountered such a case in the wild. I'm not sure how other commercial software handles multiple commands per line.

### DIFF
--- a/pycalphad/io/tdb.py
+++ b/pycalphad/io/tdb.py
@@ -826,13 +826,12 @@ def read_tdb(dbf, fd):
     splitlines = [' '.join(k.split()) for k in splitlines]
     # Remove comments
     splitlines = [k.strip().split('$', 1)[0] for k in splitlines]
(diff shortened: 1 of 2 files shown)


richardotis:
This seems like a reasonable change. Merge when ready.

I find it weird that this is TC behavior since, from what I've heard, the historical reason for the `!` line delimiter was that TDBs were often copied and pasted into plaintext emails (this was before MIME encoding for attachments) and line breaking was unpredictable.

PR #77, opened 2017-04-18
Minimizer core improvements

richardotis:
There is a lot going on in this PR, but some of that is unavoidable due to how tightly coupled the current solver is to the rest of the system. The robustness is still not where I want it to be, particularly for the chemical potentials of dilute species but, as we pass the 200 commit mark since the last release, I worry about perfect becoming the enemy of good.

There are three key objects the new solver concerns itself with:

1. `Model` / `CompiledModel`: These contain the callable implementations of each phase's energy function, gradient and Hessian. Each phase in the system has its own `Model` or `CompiledModel` object. In the case of `Model`, we use SymPy's Fortran code generation to create the callables and do the differentiation symbolically inside of SymPy. For `CompiledModel`, the entire class is implemented in Cython and the functions and derivatives are all implemented by hand, with the database parameters filled in at initialization time. The advantage of `CompiledModel` is that there is no code generation time (which can be minutes of wall clock time in real systems); the disadvantages are that it is much less flexible than `Model` in terms of the functional forms accepted, e.g., `exp(T)*T` terms do not work. Right now, `CompiledModel` is still missing the Xiong magnetic model and the two-state model for liquids. By default, the solver will fall back to `Model` for unsupported phases. These objects are pickleable and immutable after initialization, i.e., their member functions do not have side effects or mutate the object state. (`CompiledModel` has its own memory manager and *does* actually mutate internal state when calling functions but, since we hold the GIL, this is an implementation detail that can be ignored by the rest of the system.)
2. `PhaseRecord`: This object exposes a common API to the solver so it doesn't need to know about the differences between `Model` and `CompiledModel`. Each `PhaseRecord` holds a reference to its own `Model` or `CompiledModel`; these objects are pickleable. `PhaseRecord`s are immutable after initialization.
3. `CompositionSet`: This is the primary object the solver interacts with. It keeps the state of a phase (P, T, y...) at a particular solver iteration and can be updated using the `update()` member function. Every `CompositionSet` has a reference to a particular `PhaseRecord` which describes the prototype of the phase. These objects can be created and destroyed by the solver as needed to describe the stable set of phases. Multiple `CompositionSet`s can point to the same `PhaseRecord` for the case of miscibility gaps. `CompositionSet`s are not pickleable.


pycalphad/core/eqsolver.pyx
```diff
@@ -54,13 +36,14 @@ def remove_degenerate_phases(object phases, double[:,:] mole_fractions,
     cdef double[:,:] comp_distances
     cdef double phfsum = 0
     cdef object redundant_phases, kept_phase, removed_phases, saved_indices
-    cdef int num_phases = len(phases)
+    cdef int num_phases = len(composition_sets)
     cdef int phase_idx, sidx
     cdef int[:] indices
+    cdef CompositionSet compset
     # Group phases into multiple composition sets
     cdef object phase_indices = defaultdict(lambda: list())
```
bocklund: isn't `defaultdict(lambda: list())` == `defaultdict(list)`?

pycalphad/core/phase_rec.pyx
```diff
@@ -47,56 +47,11 @@ cdef public class PhaseRecord(object)[type PhaseRecordType, object PhaseRecordOb
     @cython.boundscheck(False)
```
bocklund: (GitHub is not very nice about where it lets me put comments) Can you document the purpose of PhaseRecord (line 20). Mostly w.r.t abstracting the out the difference between Model and CompiledModel API. 

I'm actually not sure if things should be documented in the pyx or the pxd. In principle, someone who is going to use it should just have to look at the definition in the pxd and not worry too much about the implementation, but it looks like the .pyx are the ones that show up in the docs.

pycalphad/core/composition_set.pyx
```diff
@@ -0,0 +1,104 @@
+from pycalphad.core.phase_rec cimport PhaseRecord
+cimport numpy as np
+import numpy as np
+
+cdef public class CompositionSet(object)[type CompositionSetType, object CompositionSetObject]:
```
bocklund: Add documentation for the CompositionSet addressing its purpose. Maybe highlight use in miscibility gaps.

pycalphad/core/eqsolver.pyx
```diff
@@ -88,140 +73,131 @@ def remove_degenerate_phases(object phases, double[:,:] mole_fractions,
...
-        site_fractions[phase_idx, :] = site_fractions[sidx, :]
-        phase_idx += 1
-    phases[saved_indices.shape[0]:] = <unicode>''
-    phase_fractions[saved_indices.shape[0]:] = np.nan
-    mole_fractions[saved_indices.shape[0]:, :] = np.nan
-    site_fractions[saved_indices.shape[0]:, :] = np.nan
+        if (composition_sets[phase_idx].NP <= MIN_SITE_FRACTION) and (not allow_negative_fractions):
+            composition_sets[phase_idx].NP = np.nan
+        elif abs(composition_sets[phase_idx].NP) <= MIN_SITE_FRACTION:
+            composition_sets[phase_idx].NP = MIN_SITE_FRACTION
```
richardotis: Should be fixed in another PR

pycalphad/core/eqsolver.pyx
```diff
@@ -447,66 +368,51 @@ def _solve_eq_at_conditions(dbf, comps, properties, phase_records, conds_keys, v
...
+            changed_phases |= remove_degenerate_phases(composition_sets, allow_negative_fractions, verbose)
+            num_phases = len(composition_sets)
+            total_dof = sum([compset.phase_record.phase_dof for compset in composition_sets])
             if num_phases == 0:
-                raise ValueError('Zero phases are left in the system', cur_conds)
-            zero_dof = np.all(
-                (prop_Y_values[it.multi_index] == 1.) | np.isnan(prop_Y_values[it.multi_index]))
-            if (num_phases == 1) and zero_dof:
+                print('Zero phases are left in the system: {}'.format(cur_conds))
+                converged = False
```
richardotis: Good catch, but I'm going to argue for leaving it as-is because the control flow isn't obvious and this makes our intentions clearer.

pycalphad/core/constants.py
```diff
@@ -6,7 +6,7 @@
 # Force zero values to this amount, for numerical stability
 MIN_SITE_FRACTION = 1e-12
 # For each phase pair with composition difference below tolerance, eliminate phase with largest index
```
richardotis: Fixed in a future PR

pycalphad/core/phase_rec.pyx
```diff
@@ -204,6 +160,7 @@ cpdef PhaseRecord PhaseRecord_from_f2py(object comps, object variables, double[:
         int var_idx, subl_index
         PhaseRecord inst
     inst = PhaseRecord()
+    # XXX: Missing inst.phase_name
```
richardotis: Fixed in #84

pycalphad/core/equilibrium.py
```diff
@@ -22,8 +22,6 @@
 from collections import namedtuple, OrderedDict
 from datetime import datetime
 
-def delayed(func, *fargs, **fkwargs):
```
richardotis: I definitely agree. Thank you for stepping up as a reviewer.

richardotis:
@bocklund I responded to all your comments with either a comment or a code change. Can you take a look and see if you have any other thoughts?

pycalphad/core/phase_rec.pyx
richardotis: I added a short explanation to the docstrings of `CompositionSet` and `PhaseRecord`, essentially just copying from the text explaining this PR.

pycalphad/core/eqsolver.pyx
```diff
@@ -23,23 +19,9 @@ MIN_SOLVE_ENERGY_PROGRESS = 1e-3
...
-        self.dof = np.zeros(len(prx.variables)+2)
-        self.dof[0] = pressure
-        self.dof[1] = temperature
-        self.dof[2:] = site_fracs
-        self.NP = phase_amt
 
-def remove_degenerate_phases(object phases, double[:,:] mole_fractions,
-                             double[:,:] site_fractions, double[:] phase_fractions, bint allow_negative_fractions):
+cdef bint remove_degenerate_phases(object composition_sets, bint allow_negative_fractions, bint verbose):
     """
```
bocklund: Fixed.

pycalphad/core/eqsolver.pyx
```diff
@@ -1,15 +1,11 @@
-#cython: profile=True
-#cython: linetrace=True
-#cython: binding=True
 from collections import defaultdict, OrderedDict
```
bocklund: Fixed in c3b08b579781dd40e9898736c8419817dd0f02b4

pycalphad/core/constants.py
bocklund: I think I was unclear. The comment in line 8 should be updated because we don't alway use COMP_DIFFERENCE_TOL to remove phases, we use it to add phases as well.

pycalphad/core/composition_set.pyx
bocklund: Fixed in c3b08b579781dd40e9898736c8419817dd0f02b4

bocklund:
Looks good. Just need to update the comment in line 8 of pycalphad/core/contstants (per the review comment) and this is ready to merge

richardotis:
Thoughts on squash versus rebase for the merge?

richardotis:
I decided on squash since many of the intermediate commits on this branch were broken.

PR #79, opened 2017-04-19
Drop asv benchmarking

bocklund:


richardotis:
Some elements of the benchmark suite may eventually make their way into the test suite (once reworked to also preserve timing information), so it's good that this will remain in the history.

PR #81, opened 2017-04-19
Warn when unused keyword arguments are passed to equilibrium

### LINKED ISSUE #52: Warn user when specifying invalid kwargs to equilibrium and calculate
Example: when specifying raw pdens as kwarg to equilibrium() -- this is incorrect

bocklund:
Closes #52 

- Calling equilibrium() with keyword arguments that are unused raises a UserWarning
- Includes a test for the behavior
- removes outdated progress bar API in test suite
- Changes other tests for warnings to assert that `len(warnings) == 1` rather than `len(warnings) > 0` in case we add new warnings in the future that could cause test to falsely pass

### DIFF
--- a/pycalphad/core/equilibrium.py
+++ b/pycalphad/core/equilibrium.py
@@ -3,6 +3,7 @@ The equilibrium module defines routines for interacting with
 calculated phase equilibria.
 """
 from __future__ import print_function
+import warnings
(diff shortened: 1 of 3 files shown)


richardotis:
Good thinking on the use of `warnings`.

PR #82, opened 2017-04-19
sympydiff_utils: Switch from Fortran to C

richardotis:
Going to C-based codegen for `Model` will work around some issues with getting a working Fortran toolchain on Windows.

Note: This PR depends on #77


bocklund:
Don't forget to remove the try: import cymem that got pulled in setup.py today.

richardotis:
That will have to get removed during rebase. This will be a mildly tricky GitHub operation, so my recommendation is to review #77 first so we can get it merged into develop. Then I'll change the base branch of this PR from bfgs_hessian to develop.

richardotis:
Fun fact: [Older versions of MSVC do not support the C99 standard ](http://stackoverflow.com/questions/21696983/error-on-declaring-variable-in-for-statement)

richardotis:
@bocklund It looks like all the tests are passing on Windows now. The pycalphad story for Windows users is still not going to be great because they will need to separately install the multi-GB Visual Studio Express to get a working C compiler (I don't think GCC will be compatible), and there is likely to be some PATH environment variable issue to resolve, but at least it's a free download.

richardotis:
For Python 3.5 and 3.6, the fastest route appears to be the [Visual C++ 2015 Build Tools](http://landinghub.visualstudio.com/visual-cpp-build-tools).

Background: http://matthew-brett.github.io/pydagogue/python_msvc.html

pycalphad/core/custom_autowrap.py
```diff
@@ -128,58 +127,274 @@ def import_extension(path, modname):
 class CodeWrapError(Exception):
     pass
 
+class C89CodePrinter(CCodePrinter):
```
bocklund: C89 due to MSVC correct? Should put a comment in for that

pycalphad/core/custom_autowrap.py
```diff
@@ -128,58 +127,274 @@ def import_extension(path, modname):
...
+            open_lines.append(loopstart % {
+                'var': self._print(i.label),
+                'start': self._print(i.lower),
+                'end': self._print(i.upper + 1)})
+            close_lines.append("}")
+        return open_lines, close_lines
+
+class CustomCCodeGen(CCodeGen):
+    def get_prototype(self, routine):
+        """Returns a string for the function prototype of the routine.
```
bocklund: I might change the word `prototype` in the docstring to `declaration` and remove the wikipedia link, but this is fine.

pycalphad/core/compiled_model.pyx
```diff
@@ -39,12 +39,8 @@ cdef inline int _intsum(int[:] arr) nogil:
 
 # From https://gist.github.com/pv/5437087
 cdef void* f2py_pointer(obj):
```
bocklund: Should this still be called `f2py_pointer`?

pycalphad/core/eqsolver.pyx
```diff
@@ -112,7 +112,7 @@ cdef bint add_new_phases(composition_sets, phase_records, current_grid, chemical
     if largest_df > minimum_df:
```
bocklund: Should like 418 also use `np.ascontiguousarray` or continue to use `asfortranarray`?

pycalphad/core/phase_rec.pyx
```diff
@@ -1,19 +1,14 @@
...
 import numpy as np
 cimport numpy as np
-from cpython cimport (PY_VERSION_HEX, PyCObject_Check,
-    PyCObject_AsVoidPtr, PyCapsule_CheckExact, PyCapsule_GetPointer, Py_INCREF, Py_DECREF)
+from cpython cimport PyCapsule_CheckExact, PyCapsule_GetPointer
 from pycalphad.core.compiled_model cimport CompiledModel
 import pycalphad.variables as v
 
 # From https://gist.github.com/pv/5437087
 cdef void* f2py_pointer(obj):
```
bocklund: Same here, do we still keep the `f2py` name?

setup.py
```diff
@@ -21,13 +21,13 @@ def read(fname):
...
     license='MIT',
     long_description=read('README.rst'),
     url='https://pycalphad.org/',
-    install_requires=['matplotlib', 'pandas', 'xarray!=0.8', 'sympy', 'pyparsing', 'tqdm', 'Cython>=0.24',
-                      'autograd', 'tinydb', 'scipy', 'numpy>=1.9', 'dask[complete]>=0.10', 'dill', 'cymem'],
+    install_requires=['matplotlib', 'pandas', 'xarray!=0.8', 'sympy', 'pyparsing', 'Cython>=0.24',
+                      'tinydb', 'scipy', 'numpy>=1.9', 'dask[complete]>=0.10', 'dill'],
     classifiers=[
         # How mature is this project? Common values are
         #   3 - Alpha
```
bocklund: Add support of `'Programming Language :: Python :: 3.6'` to line 52 of setup.py

pycalphad/core/custom_autowrap.py
```diff
@@ -128,58 +127,274 @@ def import_extension(path, modname):
...
+        else:
+            return str(arg.name)
 
     def wrap_code(self, routine, helpers=None):
         helpers = helpers if helpers is not None else []
         workdir = self.filepath
         if not os.access(workdir, os.F_OK):
             os.mkdir(workdir)
         try:
+            print(os.path.join(workdir, self.filename))
```
bocklund: Remove print statement

pycalphad/core/custom_autowrap.py
```diff
@@ -102,7 +101,7 @@ def import_extension(path, modname):
     # Blacklist fixes gh-65.
     # We filter out any files that can be created by compilers which are not our actual compiled file.
     # We cannot more directly search for our files because of differing platforms.
-    blacklist = ['dSYM']
+    blacklist = ['dSYM', 'c', 'pyx']
```
richardotis: Yes. There may be a way to control it as a command-line option, but I liked having everything in one directory for debugging and it doesn't seem like there's harm in leaving it like this.

conda_recipe/meta.yaml
```diff
@@ -48,11 +47,11 @@ requirements:
     - sympy
```
richardotis: Yes.

pycalphad/core/compiled_model.pyx
```diff
@@ -642,14 +638,14 @@ cdef public class CompiledModel(object)[type CompiledModelType, object CompiledM
                 out[out_idx] += disordered_energy
         if self._debug:
             debugout = np.zeros_like(out)
-            self._debug_energy(debugout, np.asfortranarray(dof), np.ascontiguousarray(parameters))
+            self._debug_energy(debugout, np.ascontiguousarray(dof), np.ascontiguousarray(parameters))
```
richardotis: No, it only matters for the energy function because the evaluation function is vectorized, so the memory layout in 2-D matters. For the gradient, everything is 1-D so the memory layouts are equivalent. For the Hessian, we should change it (since it's symmetric, the memory layout doesn't technically matter, which it why it still works)

pycalphad/core/eqsolver.pyx
richardotis: You are correct

pycalphad/core/custom_autowrap.py
richardotis: Because we're overriding a method from SymPy, we're limited in our ability to rename the method.

pycalphad/core/custom_autowrap.py
bocklund: What you have here is fine, just checking

setup.py
richardotis: Done

pycalphad/core/custom_autowrap.py
richardotis: Added

richardotis:
@bocklund The response to your review is up to date and ready for another look.

bocklund:
* The documentation in line 525 of pycalphad/core/custom_autowrap should be updated to reflect that in our custom_autowrap, Cython is the default. Nice job cleaning up all of the other `f2py` instances

Otherwise this looks good and is ready for rebase and merge

bocklund:
Ready to merge

Issue #76, opened 2017-04-17
errors in line calculation


broshe:
Hello, I am trying to do a line caluclation with pycalphad.
Al-Fe chosen as model system.
I am using the latest pycalphad and anaconda python on linux.
The attached script caused warning and error messages and did not run correctly.
The error message is attached below.
Also, I attached the two log files mentioned in the error message.
What's wrong?

[ebrosh@ebrosh-laptop Al-Fe-solidification]$ python try_pycalphad_linemap.py
Initialize (1/3): 100%|████████████████████████| 2/2 [00:00<00:00,  2.61phase/s]
/home/ebrosh/anaconda3/lib/python3.5/site-packages/pycalphad/core/equilibrium.py:109: RuntimeWarning: invalid value encountered in less_equal
  unstable_phases = np.nonzero(properties['NP'].values[multi_index] <= MIN_SITE_FRACTION)
Preserved Files:  []
Preserved Files:  ['/tmp/ufuncify_btb9xrf9.log', '/tmp/ufuncify_efrlo334.log']
Traceback (most recent call last):
  File "try_pycalphad_linemap.py", line 26, in <module>
    eq = equilibrium(db, ['AL', 'FE', 'VA'], my_phases, {v.X('FE'): 1.5e-3, v.T: Tvector, v.P: 101325})
  File "/home/ebrosh/anaconda3/lib/python3.5/site-packages/pycalphad/core/tempfilemanager.py", line 64, in wrapper
    return func(*args, **kwargs)
  File "/home/ebrosh/anaconda3/lib/python3.5/site-packages/pycalphad/core/equilibrium.py", line 886, in equilibrium
    results = dask.compute(*res, get=dask.multiprocessing.get)
  File "/home/ebrosh/anaconda3/lib/python3.5/site-packages/dask/base.py", line 179, in compute
    results = get(dsk, keys, **kwargs)
  File "/home/ebrosh/anaconda3/lib/python3.5/site-packages/dask/multiprocessing.py", line 86, in get
    dumps=dumps, loads=loads, **kwargs)
  File "/home/ebrosh/anaconda3/lib/python3.5/site-packages/dask/async.py", line 493, in get_async
    raise(remote_exception(res, tb))
dask.async.IndexError: index 1 is out of bounds for axis 0 with size 1

Traceback
---------
  File "/home/ebrosh/anaconda3/lib/python3.5/site-packages/dask/async.py", line 268, in execute_task
    result = _execute_task(task, data)
[...]

PR #83, opened 2017-04-24
Solver stability improvements

richardotis:
Note: This PR depends on #82.

This PR adds a few stability fixes to the solver:
1. Keep track of previously removed CompositionSets and prevent them from being re-added.
2. If the active set of phases has not changed, just calculate the change in chemical potentials for that iteration, rather than computing the full chemical potentials. This seems to confer a stability improvement.
3. Remove the unwieldy `obj_weight` scheme, which was full of magic numbers and didn't seem to work reliably.

~~I have marked this as WIP since there's some obsolete code related to `obj_weight` that still needs to be removed, but it's otherwise in shape for review.~~


richardotis:
Cool fact: For the first time, the phase diagrams of Al-Zn and Al-Fe in BinaryExamples will calculate with zero convergence failures. (I have not yet checked the chemical potentials to make sure, e.g., #76 is fixed.) For Al-Ni, I am still getting a singular matrix crash (which I suspect is related to numerical instability from magnetism), but all the other convergence failures are gone.

pycalphad/core/eqsolver.pyx
```diff
@@ -424,13 +448,20 @@ def _solve_eq_at_conditions(comps, properties, phase_records, grid, conds_keys,
             master_hess[:num_vars, num_vars:] = -constraint_jac.T
             master_hess[num_vars:, :num_vars] = constraint_jac
             master_grad = np.zeros(l_hessian.shape[0] + l_constraints.shape[0])
-            master_grad[:l_hessian.shape[0]] = -np.array(gradient_term)
+            master_grad[:l_hessian.shape[0]] = -np.array(gradient_term) + np.dot(constraint_jac.T, l_multipliers)
```
bocklund: Was this an implementation error or what is this?

pycalphad/core/eqsolver.pyx
```diff
@@ -8,7 +8,7 @@ import scipy.spatial
 import collections
 from pycalphad.core.composition_set cimport CompositionSet
 from pycalphad.core.phase_rec cimport PhaseRecord, PhaseRecord_from_f2py
-from pycalphad.core.constants import MIN_SITE_FRACTION, COMP_DIFFERENCE_TOL, BIGNUM
+from pycalphad.core.constants import MIN_SITE_FRACTION, MIN_PHASE_FRACTION, MIN_PHASE_FRACTION, COMP_DIFFERENCE_TOL, BIGNUM
```
bocklund: `MIN_PHASE_FRACTION` is imported twice

richardotis:
@bocklund This is ready for another look.

bocklund:
LGTM

PR #84, opened 2017-04-28
FIX: Set phase name when using Model-based JIT

richardotis:
The `PhaseRecord.phase_name` attribute is now properly set when using `Model` based code generation.


bocklund:
Note that this fixes a breakage in eqplot (binplot) (and everything that need phase names) where the Model was used class

PR #86, opened 2017-05-02
FIX: equilibrium: Unary calculation and property calculation

### LINKED ISSUE #78: Properties of unary component sets cannot be calculated with equilibrium
Calculate works fine. 

Any database should reproduce this with the following code:

```

db_file = '/Users/brandon/Box Sync/databases/nist-databases/Ag-Al.tdb'
dbf = Database(db_file)
eq_1 = equilibrium(dbf, ['AG', 'AL', 'VA'], phases, {v.X('AL'): 0, v.T: 500, v.P: 101325, })
eq_2 = equilibrium(dbf, ['AG', 'VA'], phases, {v.T: 500, v.P: 101325})
```

eq_1 and eq_2 should produce the same results, but eq_2 fails with the following error:

```
  File "unary_equilibrium.py", line 7, in <module>
    eq_2 = equilibrium(dbf, ['AG', 'VA'], phases, {v.T: 500, v.P: 101325})
  File "/Users/brandon/Projects/pycalphad/pycalphad/core/equilibrium.py", line 322, in equilibrium
    properties = delayed(lower_convex_hull, pure=False)(grid, properties, verbose=verbose)
  File "/Users/brandon/Projects/pycalphad/pycalphad/core/lower_convex_hull.py", line 422, in lower_convex_hull
    comp_idx = np.ravel_multi_index(it.multi_index[len(indep_conds):], comp_coord_shape)
ValueError: At least one iterator operand must be non-NULL
```

richardotis:
This PR does the following:

- Fixes #78 
- Fixes the `output` keyword argument so it actually works again (this was a gap in the test suite)
- Fixes the `parameters` keyword argument so it works with `CompiledModel` and is propagated everywhere. Also a `str` to `sympy.Symbol` conversion issue is properly handled now.

There is still a bug where undefined FUNCTIONs in PARAMETERs will raise an error for `CompiledModel` instead of just setting to zero and warning the user like `Model` does. There's some additional work in figuring out missing parameters for `CompiledModel` since `sympy.atoms` is unavailable in that mode, so I'm not sure the best way yet.

### DIFF
--- a/pycalphad/core/calculate.py
+++ b/pycalphad/core/calculate.py
@@ -461,7 +461,7 @@ def calculate(dbf, comps, phases, mode=None, output='GM', fake_points=False, bro
             continue
         if (not isinstance(mod, CompiledModel)) or (output != 'GM'):
             if isinstance(mod, CompiledModel):
-                mod = Model(dbf, comps, phase_name)
+                mod = Model(dbf, comps, phase_name, parameters=parameters)
             # Construct an ordered list of the variables
             variables, sublattice_dof = generate_dof(phase_obj, mod.components)
             # Build the "fast" representation of that model
--- a/pycalphad/core/compiled_model.pyx
+++ b/pycalphad/core/compiled_model.pyx
@@ -73,13 +73,20 @@ cdef public class CompiledModel(object)[type CompiledModelType, object CompiledM
             for comp in sublattice:
                 self.variables.append(v.Y(phase_name, idx, comp))
         parameters = parameters if parameters is not None else {}
+        renamed_params = []
+        for param, val in parameters.items():
+            if not isinstance(param, Symbol):
+                parameters[Symbol(param)] = val
+                renamed_params.append(param)
+        for param in renamed_params:
+            parameters.pop(param)
+        if isinstance(parameters, dict):
+            parameters = OrderedDict(sorted(parameters.items(), key=str))
+        param_symbols = tuple(parameters.keys())
         self._debug = _debug
         if _debug:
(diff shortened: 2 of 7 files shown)


richardotis:
The P, T assumption made in line 418 is actually pervasive in pycalphad. We would need to decide where to store the state variable information in both `Model` and `CompiledModel`, and then change all the hard-coded indexing in the phases' degrees of freedom to use a lookup table. Then we would need to write really good tests for the new functionality to convince ourselves that we didn't break anything.

bocklund:
Maybe that should define the 0.6 release

richardotis:
After re-running the test suite with caching enabled for calculate(), I'm reminded why it's disabled: it causes unpredictable breakage when changing calculation options. I think it has to do with the way certain objects are hashed when checking for cache hits. Anyway, I don't think it's a significant performance hit to keep it disabled since the actual performance bottleneck, sampling and (for `Model`) JIT code generation, are and will continue to be cached.

richardotis:
I take back my previous comment: it looks like the test error was coming from my failure to recompile after a branch change

richardotis:
However, my conclusion is the same: I'd like to leave caching disabled because I'm spooked about memory consumption and I'd want to see new profiling done demonstrating that calculate() is a bottleneck.

richardotis:
Any other comments before approval?

bocklund:
I didn't see a response to 

* 'parameters = parameters or dict()' in compiled_model.pyx line 74.

* Renaming the phase record 'prn' variable

richardotis:
Ready for another look.

Issue #54, opened 2016-08-25
Invalid number of arguments error when energy simplifies to zero


richardotis:
It's possible to get a difficult-to-debug error concerning the number of arguments passed to the energy function. This happens if the energetic contribution simplifies to zero. For example:

```
PHASE M7C3_D101 %  2 7 3 !
CONSTITUENT M7C3_D101 :MN:C: !
PARAMETER G(M7C3_D101,MN:C;0)  1 VV22+VV23*T**2+VV24*T**3
 +VV25*T**4+VV26*T**5; 6000 N !
```

where all of the above `VV` variables are zero. If the user happens to typo the name of the phase in the TDB, they may be scratching their head about this error (perhaps parameters without a matching phase should throw their own error, but that's a separate issue).

richardotis:
`Model` seems to work fine now, but `CompiledModel` will raise a `TypeError` complaining about undefined symbols. This is definitely an easier to understand error message, but for consistency we expect `CompiledModel` to set all undefined symbols to zero while warning the user.

richardotis:
I'm thinking to defer this issue (of undefined symbols in `CompiledModel`) to after 0.5 to think more about solutions.

PR #90, opened 2017-05-05
FIX: Undefined symbols in CompiledModel are automatically set to zero

richardotis:
Fixes gh-54.

The warning mechanism here (using the `logging` module) is the same as what is used in `calculate` to emit the same warning for `Model`, but it is different from the mechanism (using the `warnings` module) in `equilibrium`. I think there may be some kind of reason why I originally picked the logging module (perhaps some ability to redirect or disable messages selectively), but it did not seem to catch on with me for whatever reason. I am also not sure how to do the equivalent of `warnings.catch_warnings` with the `logging` module to programmatically verify we are emitting the correct warnings in the test.


bocklund:
You can set up filtering with warnings similarly, I think. IMO logging should be reserved for looking at what happened in history and shouldn't be used to let the user know that something went wrong at runtime. 

Also I believe logging output has historically been more tricky to make visible in jupyter notebooks without manually importing logging and setting the log level. So from that and a binder or jupyter hub perspective, I'd like to see this in `warnings` unless there were a compelling reason to use `logging`. 

All that being said, is this worth issuing a warning in logging or otherwise?

richardotis:
This particular warning (undefined symbol) has saved me a few times when I typo'd a FUNCTION name or forgot to define it. It helped me figure out why I was getting nonsensical results. *Usually* an undefined symbol is a mistake, and if I were designing TDBs from scratch I would make it an error, but setting it to zero matches the behavior in TC and is sometimes convenient when building new databases.

I'll take a look at changing the warning method from logging to the warnings module here and also in calculate.

richardotis:
For the record, the way to programmatically capture logging output is discussed here: http://alanwsmith.com/capturing-python-log-output-in-a-variable

With the key part of it reproduced here
```python
import logging
import io

### Create the logger
logger = logging.getLogger('basic_logger')
logger.setLevel(logging.DEBUG)

### Setup the console handler with a StringIO object
log_capture_string = io.StringIO()
ch = logging.StreamHandler(log_capture_string)
ch.setLevel(logging.DEBUG)

### Optionally add a formatter
formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
ch.setFormatter(formatter)

### Add the console handler to the logger
logger.addHandler(ch)


### Send log messages. 
logger.debug('debug message')
logger.info('info message')
logger.warn('warn message')
logger.error('error message')
logger.critical('critical message')


### Pull the contents back into a string and close the stream
log_contents = log_capture_string.getvalue()
log_capture_string.close()

### Output as lower case to prove it worked. 
print(log_contents.lower())
```

richardotis:
Good idea.

richardotis:
@bocklund Trying to figure out why the Linux py27 tests are hanging. I thought maybe it's a dependency issue but the previous commit runs fine when restarted.

richardotis:
I cannot reproduce this in a clean local py27 environment.

richardotis:
The coverage check looks bugged but otherwise everything passes, so I'll merge.

PR #98, opened 2017-06-07
Support triangular plots

bocklund:
Major points:

* Add `triangular` matplotlib projection
* Add `ternplot` API that functions similarly to binplot
* Both `ternplot` and `binplot` use `eqplot`. The behavior of `eqplot` is controlled by the `projection`. When 3D plotting etc. is added, the constraints for different projections can simply be modified

Minor points:
* ternplot uses similar API and semantics as binplot
* Fix binplot error when only single phase regions exist (e.g. only liquid, even though it's not an interesting diagram)
* Add _axis_label function to more generally handle axis labels
* Small docs fixes

Pay attention to:
* Is there a better way to transform the y axis label in the triangular plot? (eqplot line 98)
* Any ideas on eqplot line 124? I think we could do that and it would make things a little cleaner. Especially because we could just append all of the found phase stuff together. 
* Anything seem wrong in my ternary colors? If you make the symbol sizes larger, you can see that the three phase ones are miscolored
* Should we sort the independent component list in the ternplot API to get more consistent (alphabetic) plotting or just let it be unless the user specifies with `x=` and `y=` in ternplot? 

Example code:

```python
from matplotlib import pyplot as plt
from pycalphad import Database, ternplot, variables as v
dbf = Database(database_file)
comps = ['AL','CU','ZR','VA']
fig = plt.figure(figsize=(20,20))
conditions = {v.X('ZR'): (0,1,0.01), v.X('AL'): (0,1,0.01), v.P: 101325, v.T: 1000}
ternplot(dbf, comps, phases, conditions, x=v.X('ZR'), y=v.X('AL')) 
```


bocklund:
@richardotis This was updated comments made offline:

* Removed projection kwarg from eqplot. Plot type (projection) is determined by the degrees of freedom in the passed equilibrium. 
This might make isopleths more confusing to call with the eqplot API, but also more explicit by passing the eq object after the correct number of DOF have been selected with xarray. 

* Remove the phases argument from eqplot. The phases argument doesn't make sense because you wouldn't ever want to plot a phase that may be in stable equilibrium in eqplot.

richardotis:
FYI regarding one of the failing tests: https://github.com/numpy/numpy/issues/9251

pycalphad/plot/eqplot.py
```diff
@@ -55,62 +66,116 @@ def eqplot(eq, ax=None, x=None, y=None, z=None, phases=None, **kwargs):
...
-    two_phase_indices = np.nonzero(np.sum(eq.Phase.values != '', axis=-1, dtype=np.int) == 2)
-    found_phases = eq.Phase.values[two_phase_indices][..., :2]
+    # Select all two- and three-phase regions
+    three_phase_idx = np.nonzero(np.sum(eq.Phase.values != '', axis=-1, dtype=np.int) == 3)
+    two_phase_idx = np.nonzero(np.sum(eq.Phase.values != '', axis=-1, dtype=np.int) == 2)
 
     legend_handles, colorlist = phase_legend(phases)
-    # position the phase legend
+
+    # TODO: can we concatenate the last dimension of these axes to have just one plot command for tielines and another for points?
```
richardotis: This probably isn't important.

pycalphad/plot/eqplot.py
```diff
@@ -55,62 +66,116 @@ def eqplot(eq, ax=None, x=None, y=None, z=None, phases=None, **kwargs):
...
+        # As the figure size gets very large, the label approaches ~0.55 on the yaxis
+        # 0.55*cos(60 deg)=0.275, so that is the xcoord we are approaching.
+        # TODO: it would be nice to handle this automatically
+        ax.yaxis.label.set_va('baseline')
+        fig_x_size = ax.figure.get_size_inches()[0]
+        y_label_offset = 1 / fig_x_size
+        ax.yaxis.set_label_coords(x=(0.275 - y_label_offset), y=0.5)
+
+    # get the active phases and support loading netcdf files from disk
+    phases = map(str, sorted(set(np.array(eq.Phase.values.ravel(), dtype='U')) - {''}, key=str))
```
richardotis: There should probably be a test which verifies that `eqplot(eq)` == `eqplot(deserialize(serialize(eq))`. This could probably be done by writing the output of the LHS and RHS to files which should compare bitwise identical

pycalphad/plot/eqplot.py
```diff
@@ -55,62 +66,116 @@ def eqplot(eq, ax=None, x=None, y=None, z=None, phases=None, **kwargs):
...
     legend_handles, colorlist = phase_legend(phases)
-    # position the phase legend
+
+    # TODO: can we concatenate the last dimension of these axes to have just one plot command for tielines and another for points?
+    #ax.scatter(two_phase_x.reshape((-1, 1)), two_phase_y.reshape((-1, 1)), c=two_phase_plotcolors.reshape(-1))
+    #ax.scatter(three_phase_x.reshape((-1, 1)), three_phase_y.reshape((-1 ,1)),  c=three_phase_plotcolors.reshape(-1))
+
+    # For both two and three phase, cast the tuple of indices to an array and flatten
+    # If we found two phase regions:
+    if len(np.array(two_phase_idx).flatten()) > 0:
```
richardotis: You can simplify this with `two_phase_idx.size > 0`

pycalphad/plot/eqplot.py
```diff
@@ -55,62 +66,116 @@ def eqplot(eq, ax=None, x=None, y=None, z=None, phases=None, **kwargs):
...
+        tielines = np.rollaxis(tielines,1)
+        lc = mc.LineCollection(tielines, zorder=1, colors=[0,1,0,1], linewidths=[0.5, 0.5])
+        # plot two phase points and tielines
+        two_phase_plotcolors = np.array(list(map(lambda x: [colorlist[x[0]], colorlist[x[1]]], found_two_phase)), dtype='U') # from pycalphad
+        ax.scatter(two_phase_x[..., 0], two_phase_y[..., 0], s=3, c=two_phase_plotcolors[:,0], edgecolors='None', zorder=2, **kwargs)
+        ax.scatter(two_phase_x[..., 1], two_phase_y[..., 1], s=3, c=two_phase_plotcolors[:,1], edgecolors='None', zorder=2, **kwargs)
+        ax.add_collection(lc)
+
+    # If we found three phase regions:
+    if len(np.array(three_phase_idx).flatten()) > 0:
```
richardotis: You can simplify this with `three_phase_idx.size > 0`

pycalphad/plot/eqplot.py
```diff
@@ -55,62 +66,116 @@ def eqplot(eq, ax=None, x=None, y=None, z=None, phases=None, **kwargs):
...
+        three_phase_x = eq.X.sel(component=x.species).values[three_phase_idx][..., :3]
+        three_phase_y = eq.X.sel(component=y.species).values[three_phase_idx][..., :3]
+        # three phase tielines
+        three_phase_tielines = np.array([np.concatenate((three_phase_x[..., 0][..., np.newaxis], three_phase_y[..., 0][..., np.newaxis]), axis=-1),
+                                     np.concatenate((three_phase_x[..., 1][..., np.newaxis], three_phase_y[..., 1][..., np.newaxis]), axis=-1),
+                                     np.concatenate((three_phase_x[..., 2][..., np.newaxis], three_phase_y[..., 2][..., np.newaxis]), axis=-1)])
+        three_phase_tielines = np.rollaxis(three_phase_tielines,1)
+        three_lc = mc.LineCollection(three_phase_tielines, zorder=1, colors=[1,0,0,1], linewidths=[0.5, 0.5])
+        # plot three phase points and tielines
+        # TODO: the colors seem to be ordered wrong on the plots
```
richardotis: What does this mean?

pycalphad/plot/ternary.py
```diff
@@ -0,0 +1,57 @@
...
+    Returns
+    -------
+    A phase diagram as a figure.
+
+    Examples
+    --------
+    None yet.
+    """
+    eq_kwargs = eq_kwargs if eq_kwargs is not None else dict()
+    indep_comps = [key for key, value in conds.items() if isinstance(key, v.Composition) and len(np.atleast_1d(value)) > 1]
```
richardotis: I recommend `sorted(conds.items())` so that the plot is deterministic

pycalphad/plot/eqplot.py
richardotis: The TODO comment, I mean

pycalphad/plot/eqplot.py
```diff
@@ -55,62 +66,116 @@ def eqplot(eq, ax=None, x=None, y=None, z=None, phases=None, **kwargs):
...
-    phases = phases if phases is not None else \
-        map(str, sorted(set(np.array(eq.Phase.values.ravel(), dtype='U')) - {''}, key=str))
+    ax = plt.gca(projection=projection) if ax is None else ax
+
+    # Handle cases for different plot types
+    if projection is None:
+        x = indep_comps[0] if x is None else x
+        y = indep_pots[0] if y is None else y
+        # plot settings
+        ax.set_xlim([np.min(conds[x]) - 1e-2, np.max(conds[x]) + 1e-2])
```
bocklund: eqplot returns the Axes object so if someone wanted to customize anything I think they should just do it in matplotlib directly. IMO we should choose a few sane defaults for the common path and allow users to do fine adjustments in matplotlib or an image editor.

pycalphad/plot/eqplot.py
bocklund: The coloring of the markers seemed wrong on the tie triangles I was looking at. The three marker colors were usually correct, but they were on the wrong vertex of the triangle.

richardotis: I think you are selecting the first, second and third rows when you should be selecting the columns.

pycalphad/plot/eqplot.py
```diff
@@ -55,62 +66,116 @@ def eqplot(eq, ax=None, x=None, y=None, z=None, phases=None, **kwargs):
...
+            two_phase_y = np.array([two_phase_y, two_phase_y]).swapaxes(0, 1)
+
+        # construct tielines
+        tielines = np.array([np.concatenate((two_phase_x[..., 0][..., np.newaxis], two_phase_y[..., 0][..., np.newaxis]), axis=-1),
+                     np.concatenate((two_phase_x[..., 1][..., np.newaxis], two_phase_y[..., 1][..., np.newaxis]), axis=-1)])
+        tielines = np.rollaxis(tielines,1)
+        lc = mc.LineCollection(tielines, zorder=1, colors=[0,1,0,1], linewidths=[0.5, 0.5])
+        # plot two phase points and tielines
+        two_phase_plotcolors = np.array(list(map(lambda x: [colorlist[x[0]], colorlist[x[1]]], found_two_phase)), dtype='U') # from pycalphad
+        ax.scatter(two_phase_x[..., 0], two_phase_y[..., 0], s=3, c=two_phase_plotcolors[:,0], edgecolors='None', zorder=2, **kwargs)
```
richardotis: Note the use of `two_phase_plotcolors[:,0]` -- selected by column

pycalphad/plot/eqplot.py
```diff
@@ -55,62 +66,116 @@ def eqplot(eq, ax=None, x=None, y=None, z=None, phases=None, **kwargs):
...
+        # three phase tielines
+        three_phase_tielines = np.array([np.concatenate((three_phase_x[..., 0][..., np.newaxis], three_phase_y[..., 0][..., np.newaxis]), axis=-1),
+                                     np.concatenate((three_phase_x[..., 1][..., np.newaxis], three_phase_y[..., 1][..., np.newaxis]), axis=-1),
+                                     np.concatenate((three_phase_x[..., 2][..., np.newaxis], three_phase_y[..., 2][..., np.newaxis]), axis=-1)])
+        three_phase_tielines = np.rollaxis(three_phase_tielines,1)
+        three_lc = mc.LineCollection(three_phase_tielines, zorder=1, colors=[1,0,0,1], linewidths=[0.5, 0.5])
+        # plot three phase points and tielines
+        # TODO: the colors seem to be ordered wrong on the plots
+        three_phase_plotcolors = np.array(list(map(lambda x: [colorlist[x[0]], colorlist[x[1]], colorlist[x[2]]], found_three_phase)), dtype='U') # from pycalphad
+        ax.scatter(three_phase_x[..., 0], three_phase_y[..., 0], s=3, c=three_phase_plotcolors[0], edgecolors='None', zorder=2, **kwargs)
```
richardotis: It should be `three_phase_plotcolors[:,0]` and so on

bocklund:
I addressed everything except for the test. I think the proposed test would be hard to maintain. 

If there are some specific things you want to test, I can break those out into functions and test them. Or possibly just do an integration test to see that eqplot returns an Axes object.

richardotis:
An integration test would be fine. What needs to happen to fix the tests on AppVeyor?

bocklund:
Looks like the NumPy issue from above and a failure in a warning test because we are also getting a warning for using dask.async.get_sync which is deprecated in favor of dask.local.get_sync

bocklund:
@richardotis this is ready for another look once tests pass

bocklund:
@richardotis looks like the AppVeyor build of on py34 that uses sympy<1.1 (#105) is passing. 

This is ready to be squashed and merged with review approval.

PR #99, opened 2017-06-08
Fix CementiteAnalysis.ipynb - by adding np.squeeze()

jan-janssen:
It has to be "fig.gca().plot(result['T'], np.squeeze(4.0 * result['heat_capacity']))" rather than "fig.gca().plot(result['T'], 4.0 * result['heat_capacity'])"


bocklund:
Thanks!

`result['heat_capacity'].sel(P=101325, points=0)` would also work in cases where the other dimensions do not have a length of 1 (e.g. you want to call calculate once for many pressures, but make multiple plots).

PR #104, opened 2017-06-29
DOC:cache.py/equilibrium.py/database.py/tdb.py:Edit docstrings for co…

olivia-higgins:
…nsistent format


olivia-higgins:
@bocklund

bocklund:
It looks like the AppVeyor continuous integration build failed. Can you take a look at the Details for one of the builds and see if one of your changes caused the failure?

bocklund:
@olivia-higgins anything I can help with on this?

olivia-higgins:
I did what you said to do up to the forced push. I was getting error messages saying that I needed to set an upstream for my branch. I followed the recommendations the message gave and then continued with the forced push. Let me know if this was not the correct thing to do, thanks.

bocklund:
Your changes pushed correctly, just waiting on the tests now. 

On the setting upstream: the first time you pushed, you probably typed something like `git push origin fix-typos` then just now to force push maybe you only did `git push -f`? If that was the case, it would give you those error messages to set an upstream because it doesn't know where to push to. If you did the suggested `git push -f -u origin fix-typos` then the `-u` option set the upstream to the fix-typos branch and from now on `git push` or `git push -f` will correctly push to the correct branch.

bocklund:
@olivia-higgins I don't think we are actually using the log.py file anywhere. Do you want to just remove it and push those changes and see if tests pass?

olivia-higgins:
@bocklund This time there is an ImportError, the error message reads "no module named log". Any suggestions on what else to try?

bocklund:
We are importing and using logging in a couple places it looks like: https://github.com/pycalphad/pycalphad/search?utf8=%E2%9C%93&q=logger&type=

I think the debug in model could just be removed, the logger.warning should be made into a real warning with the Python standard library [`warnings` module](https://docs.python.org/3.6/library/warnings.html), which is already imported in `calculate.py`.

bocklund:
> @olivia-higgins (via email): For changing the logger.warning into a warning, when using the warn() function is there a way to keep the message format as it is in logger.warning, so that phase_name is inserted into the printed message? 

logging has the format string stuff built in (`fmt = '%(name)s...'` in line 9 of `log.py`). Now you'll have to format things manually instead of depending on the logger to do it for you. https://pyformat.info/ is a really good reference.

In pycalphad we mix some of the '%' (old) style strings with the `.format()` (new) style. Use the new style in all new code.

PR #106, opened 2017-07-18
DOC: Add ternary examples

bocklund:


docs/examples/TernaryExamples.rst
```diff
@@ -0,0 +1,84 @@
...
+
+
+
+.. image:: TernaryExamples_files/TernaryExamples_2_1.png
+
+
+triangular projection
+---------------------
+
+Importing the ``pycalphad.plot.traingular`` module automatically
```
richardotis: spelling of module name

docs/examples/TernaryExamples.rst
```diff
@@ -0,0 +1,84 @@
...
+
+    <matplotlib.axes._subplots.TriangularAxesSubplot at 0x10d907518>
+
+
+
+
+.. image:: TernaryExamples_files/TernaryExamples_2_1.png
+
+
+triangular projection
```
richardotis: This example was confusing the first time I saw it because I didn't understand the plot was supposed to be nonsense. This would be improved with a contour plot of a fictive custom property model.

bocklund:
@richardotis this is ready for another look

bocklund:
I ended up using mixing enthalpy and plotting a scatter plot. 

Although not too interesting, I think this most clearly demonstrates it's utility.  It also avoids mesh gridding and interpolating with scipy to get the right inputs for a contour.

A better, separate example for custom property models helps to keep the purpose of the example distinct from property models and allows users to compose the ideas themselves based on clear examples.

richardotis:
Can you rebase against develop so we can verify all the tests pass?

Issue #105, opened 2017-07-10
Update sympy version to 1.1 because of moved CCodePrinter. Make required changes.


bocklund:
See https://github.com/sympy/sympy/issues/12687 and https://github.com/sympy/sympy/issues/12220

PR #108, opened 2017-07-20
FIX: sympy 1.1 compatibility

richardotis:
Fixes gh-105.

### DIFF
--- a/pycalphad/core/custom_autowrap.py
+++ b/pycalphad/core/custom_autowrap.py
@@ -69,7 +69,7 @@ from sympy.core.compatibility import iterable
 try:
     from sympy.utilities.codegen import CCodePrinter
 except ImportError:
-    from sympy.printing.ccode import C99CodePrinter as CCodePrinter
+    from sympy.printing.ccode import C89CodePrinter as CCodePrinter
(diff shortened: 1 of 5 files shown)


bocklund:
Can you remind me why we are import checking rather than just `sympy >= 1.1`?

Something about python 3.4 support, right?

richardotis:
Why break backcompat when you can keep it with a two line check?

And I think Python 3.4 support will be going away soon when IPOPT starts getting merged.

On Jul 20, 2017 2:23 PM, Brandon Bocklund <notifications@github.com> wrote:

Can you remind me why we are import checking rather than just sympy >= 1.1?

Something about python 3.4 support, right?

bocklund:
That's fine. Still some failing tests here

bocklund:
Looks like Mac builds are failing on travis, but everything else looks ok.

richardotis:
`TypeError: argument is not an mpz` is an indication that gmpy2 is installed. It's purposefully excluded from the Linux-based Docker images but it's slipping into the Mac builds.

richardotis:
Tracking issue: sympy/sympy#12895

richardotis:
Based on separate discussion, we think this is suitable for merge. We'll work with upstream to get the issue with gmpy2 fixed.

PR #109, opened 2017-07-23
FIX: Improve tests that check warnings.

bocklund:
Warning testing is independent of the number of warnings raised and depends instead on specific string fragments.


bocklund:
Tests passed locally, but this will have to wait for a rebase onto #108 to see if all the CI tests pass.

PR #115, opened 2017-08-01
fixed some of the old string formatting

mxf469:


bocklund:
Thanks for this! Your changes look good. Can you use git to amend your commit and update it to a similar [style that we use in the repo] (https://github.com/pycalphad/pycalphad/commits/develop)? Then force push those changes up here and I will approve and merge.

PR #117, opened 2017-08-06
FIX: Database.write should raise if one of the specified values is not entered.

### LINKED ISSUE #94: Passing a nonsense string to the if_exists arg of Database.write results in overwrite as default
Line 285 in database.py

```
if os.path.exists(fname) and if_exists != 'overwrite':
	if if_exists == 'raise':
		raise FileExistsError('File {} already exists'.format(fname))
	elif if_exists == 'rename':
		writetime = datetime.now()
		fname = os.path.splitext(fname)
		fname = fname[0] + "." + writetime.strftime("%Y-%m-%d-%H-%M") + fname[1]
with open(fname, mode='w') as fd:
    format_registry[fmt].write(self, fd, **write_kwargs)
```

If any text at all is passed that is not `raise` or `rename` (e.g. a typo), an overwrite will occur. Should we keep this as the default?

bocklund:
Closes #94.

This handles cases where `if_exists` is passed a string that is not in ['raise', 'overwrite', 'rename']. In that case, we should raise rather than overwrite.

### DIFF
--- a/pycalphad/io/database.py
+++ b/pycalphad/io/database.py
@@ -284,12 +284,13 @@ class Database(object): #pylint: disable=R0902
             format_registry[fmt].write(self, fname, **write_kwargs)
         else:
             if os.path.exists(fname) and if_exists != 'overwrite':
-                if if_exists == 'raise':
-                    raise FileExistsError('File {} already exists'.format(fname))
-                elif if_exists == 'rename':
+                if if_exists == 'rename':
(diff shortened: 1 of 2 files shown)


bocklund:
+1 for tests revealing breakages. FileExistsError was introduced in Python 3.3.

richardotis:
Looks good to merge pending tests.

PR #122, opened 2017-08-11
WIP: Automated deployment to PyPI

richardotis:
1.  A whitelist of GPG public keys will be maintained in the repo.
2. Before deployment, the GPG signature of the tag on master is verified against the whitelist.
3. Travis handles deployment to PyPI.


richardotis:
It doesn't look like there's a way to do this without giving the ability to dump deployment secrets to everyone with push access. For now, deployment of new releases will continue to be done manually.

PR #135, opened 2017-10-09
MAINT: Cleanup refdata, fitting, and core.eqresult

bocklund:
These modules appear to be dead code and unused and are removed.
Potentially useful is the `pycalphad.refdata`, which has been moved to ESPEI because it fits more into the scope of ESPEI than pycalphad.

refdata will be released in ESPEI with ESPEI version 0.3.1


bocklund:
# DO NOT MERGE UNTIL ESPEI 0.3.1 IS RELEASED WITH REFDATA

I've replaced the code in the refdata module with an import error saying that it has been moved to ESPEI. @richardotis do you think this is necessary? Should we include a date that the pycalphad.refdata module will be removed from pycalphad, e.g. pycalphad 0.7?

bocklund:
This is ready for a review

PR #136, opened 2017-10-12
ENH: Plot tielines option

bocklund:
Add a tieline bool option to turn off tieline plotting in eqplot

The keyword argument is propagated from binplot/ternplot kwargs and there
is not an explicit keyword argument. Should there be?

This also changes the plot colorlist by removing the awful yellow and changing the order. By changing the order we don't guarantee that to like colors are not next to each other on the phase diagram, but when the number of phases is small we can ensure that only unique colors show up.


richardotis:
Tests on Travis are failing but it looks like it's due to `gmpy2` sneaking in again as a dependency and triggering https://github.com/sympy/sympy/issues/12895 . Since the tests on the other platforms pass I'm going to ignore that issue for the purposes of the review.

richardotis:
The explicit keyword argument is binplot and ternplot is unnecessary.

richardotis:
Since this is difficult to test for in CI, could you add an example here in this issue showing the tielines option turned on and off for a binary and ternary phase diagram?

Issue #123, opened 2017-08-13
TDB parse errors with unary 5.0


richardotis:
```python
import requests
from pycalphad import Database
pure_elements = requests.get('http://www.crct.polymtl.ca/sgte/unary50.tdb')
dbf = Database(pure_elements.text)
```
```
Failed while parsing:  ELEMENT N 1/2_MOLE_N2(G) 1.4007E+01 4.3350E+03 9.5751E+01 
Tokens: None
Failed while parsing:  ELEMENT O 1/2_MOLE_O2(G) 1.5999E+01 4.3410E+03 1.0252E+02 
Tokens: None
Failed while parsing:      SPECIES N2 N2 
Tokens: None
Failed while parsing:  SPECIES O2 O2 
Tokens: None
```
(I commented out the raise command so it would finish parsing. Incidentally, this would probably be a useful feature, but that's another issue.)

richardotis:
@bocklund Is this fixed now?

bocklund:
Yes

Issue #110, opened 2017-07-24
Species support


bocklund:
Tasks:

- [x] determine and implement a representation for aliasing species and storing charge information
- [x] add support for parsing species in tdbs, modifying tdb parsing element parsing to match the above
- [ ] use species in the model (including calculating charge?)
- [ ] add charge balance as a constraint in the solver class

richardotis:
The first two tasks are done now, right?

richardotis:
Opinions (open to discussion) on how users manipulate species entries in Models:
1. All element/species strings entered in the `comps` should be internally promoted to `Species` objects (or a new `core` object if we want to enforce separation between `core` and `io`; alternatively we promote `Species` to a top-level object (TLO)). All components in the constituent arrays of `Database`s should also be promoted to objects.
2. Following (1), string-based comparison of components should go away everywhere. This may require Cythonization of `Species` (another point in favor of promotion to TLO) since string-based comparison of species names is in the hot path in several places.
3a. If only pure elements are specified in `comps`, all species which can be formed from the set of active elements should be entered.
3b. If at least one species is specified, no additional species should be automatically entered, but the corresponding pure elements (if not specified) should be automatically entered.
3c. We should also support "negation notation" for species where a user can write `'~CO2` or similar, and the system will ensure that the species is suspended, even if it would be automatically entered. I can also imagine supporting this for pure elements, so you could enter a species without a corresponding pure element, but how this feature would interact with 3a and 3b is unclear.

I am not really happy with the solution proposed in 3a-c; hoping we can come up with something better.

bocklund:
What don't you like about 3? 3c might be kind of clumsy, but a and b seem like reasonable defaults. Assume all the species are active unless otherwise specified.

richardotis:
This is just a situation where it seems like TC's stateful workspaces allow a more elegant expression of user intent.

With some examples showing users how to apply `set` operations to `dbf.species` to enter exactly what they want, it's probably fine.

bocklund:
OTOH, the method proposed in 3 means you never have to (interactively or programmatically) inspect the state or find where the state is set in a script/program.

PR #137, opened 2017-10-14
ENH: Species support in Database and TDB read/write

bocklund:
Species in a TDB are parsed in the `Specie` class and added to the `species` set attribute of the Database.

The Specie class has several attributes:

- `name`: the name of the specie
- `constituents`: a dictionary of `{element: composition}` of the specie
- `charge`: integer charge. Defaults to zero if none specified.

As discussed in the Thermo-Calc database manager guide, elements entered via the ELEMENT command are automatically entered as species. We match that behavior here. The constituents of an element are just that element with a fraction of 1 and the charge is 0.

In principle, the constituents dictionary will be used in the solver to link the specie to how it contributes to the overall mass balance conditions.

Tests are included that test the basic functionality. I think the examples I chose were representative, and I'm unaware of any edge cases that I missed.

This could be merged immediately to better support parsing of species, but no changes were made to the solver and the species defined here should not affect calculations in any way.

@richardotis comments? suggestions?

### DIFF
--- a/pycalphad/io/database.py
+++ b/pycalphad/io/database.py
@@ -43,10 +43,13 @@ class Specie(object):
         Name of the specie
     constituents : dict
         Dictionary of {element: quantity} where the element is a string and the quantity a float.
+    charge : int
+        Integer charge. Can be positive or negative.
     """
-    def __init__(self, name, constituents):
+    def __init__(self, name, constituents, charge):
         self.name = name
         self.constituents = constituents
+        self.charge = charge
 
     def __eq__(self, other):
         """Two species are the same if their names and constituents are the same."""
--- a/pycalphad/io/database.py
+++ b/pycalphad/io/database.py
@@ -33,6 +33,31 @@ def _to_tuple(lst):
     return tuple(_to_tuple(i) if isinstance(i, list) else i for i in lst)
 
 
+class Specie(object):
+    """
+    A specie in the database.
+
+    Attributes
+    ----------
+    name : string
+        Name of the specie
+    constituents : dict
+        Dictionary of {element: quantity} where the element is a string and the quantity a float.
+    """
+    def __init__(self, name, constituents):
+        self.name = name
+        self.constituents = constituents
+
+    def __eq__(self, other):
+        """Two species are the same if their names and constituents are the same."""
+        if isinstance(other, self.__class__):
+            return (self.name == other.name) and (self.constituents == other.constituents)
+        else:
+            return False
+
+    def __hash__(self):
+        return hash(self.name)
+
 class Phase(object): #pylint: disable=R0903
     """
     Phase in the database.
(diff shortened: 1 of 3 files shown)


bocklund:
This would move towards closing #123 (the Unary 50 TDB parses, but fails in writing, in what seems to be an issue unrelated to species) and check off the first two tasks in #110

richardotis:
What about using SPECIES in PARAMETERs? Does that parse correctly?

bocklund:
IIRC,it already does, and we don't enforce the need for the values to be elements. But I can write a test to verify

bocklund:
Parsed correctly, but did not write correctly. Fixed with test in newest changes.

richardotis:
Since this is backwards-compatible with the existing solver implementation I'm okay with not solving this in this issue, but I'd like to do some thinking about how species should be supported in `Model`, `calculate` and ultimately `equilibrium`.

bocklund:
I agree. I think the species class might also be the correct place to keep our assumptions about the species, e.g. vacancies, so we can get away from implicit assumptions throughout the codebase.

For the most part, I envision that everything in `Model`, `calculate` and `equilibrium` will be based on species instead of elements. The only part that will depend on elements is the mass balance. 

I think most of the changes are relatively straightforward replacement of element with species in `Model` and `calculate`, but I imagine we'll have to make more complicated changes for supporting stepping in species space rather than element space in `equilibrium` and our visualization.

pycalphad/io/database.py
```diff
@@ -33,6 +33,34 @@ def _to_tuple(lst):
     return tuple(_to_tuple(i) if isinstance(i, list) else i for i in lst)
 
 
+class Specie(object):
```
richardotis: The internet seems to say that "species" is spelled and pronounced the same way for the singular and plural case (like "sheep" or "deer"). I think we should compromise on clarity for the sake of grammatical correctness here.

pycalphad/io/database.py
```diff
@@ -33,6 +33,34 @@ def _to_tuple(lst):
...
+
+    def __eq__(self, other):
+        """Two species are the same if their names and constituents are the same."""
+        if isinstance(other, self.__class__):
+            return (self.name == other.name) and (self.constituents == other.constituents)
+        else:
+            return False
+
+    def __hash__(self):
+        return hash(self.name)
```
richardotis: This is not a good implementation of hash because it will break putting species into sets and frozensets when some species have the same name but different charges. Or is the name an unambiguous descriptor?

pycalphad/io/database.py
```diff
@@ -33,6 +33,34 @@ def _to_tuple(lst):
...
+        Dictionary of {element: quantity} where the element is a string and the quantity a float.
+    charge : int
+        Integer charge. Can be positive or negative.
+    """
+    def __init__(self, name, constituents, charge):
+        self.name = name
+        self.constituents = constituents
+        self.charge = charge
+
+    def __eq__(self, other):
```
richardotis: See comment for `__hash__`

pycalphad/io/database.py
bocklund: Well that's embarrassing... I'll fix it.

pycalphad/io/database.py
bocklund: The TDB format differentiates species (for example Al+2 and Al+3) by their names, which are arbitrary but unique and unambiguous. In that case, the names are conventionally `AL+3` and `AL+2` (but again, are arbitrary). 

If I am understanding your meaning correctly, the tests should demonstrate this case for the species 
```
SPECIES O-2                         O1/-2!
SPECIES O1                          O!
```

which have the same chemical stoichometry but are differentiated by name.

richardotis: That answers my question, so it should be fine.

PR #141, opened 2017-11-09
ENH: Add a function to filter inactive phases

### LINKED ISSUE #100: Filter inactive phases in calculate and equilibrium
Phases containing sublattices with all inactive components are currently handled differently in calculate and equilibrium. In equilibrium a DoFError is raised and in calculate, the DoFError is caught and the phase is suspended with a warning.

Options:
1. Make equilibrium catch the error and give a similar warning (kind of ugly for higher-order systems with many phases)
2. Filter out phases on the front ends and just raise an error
    a. either with just some code
    b. provide a database convince function that gives active phases given a compositions

A sample piece of code for (currently) finding active phases looks like:
```python
active_phases = [phase for phase in dbf.phases.keys() if all(
                 len(set(comps).intersection(subl)) > 0 for subl in dbf.phases[phase].constituents)] 
```

Architecture for future species support (how primitive components of species are represented) may change the implementation of this.

### LINKED ISSUE #129: more descriptive error messages, if possible
When inputting a phase into the phase list that does not exist in the TDB, attempting to use this list in, e.g., binplot() results in the not-terribly-useful error message: KeyError: <wrong phase>.  It's fairly easy to figure out, but it would be nice if there were a check or a message like "this phase <wrong phase> does not exist in the TDB."

If I come across additional places for more helpful messages, I'll update this until it's closed.

Update: Also, there should be an error message for the bug in issue #116, instead of silently calculating the wrong free energy.

bocklund:
Closes #100 and closes #129

I use this all the time. @richardotis what do you think about including this in pycalphad proper?

Any thoughts on doing this automatically in equilibrium calculations? That way users could always pass in phases=dbf.phases.keys() without getting any error messages. In large databases, your only other choice is to tediously select phases in the system you want to calculate.

### DIFF
--- a/pycalphad/core/equilibrium.py
+++ b/pycalphad/core/equilibrium.py
@@ -205,7 +205,7 @@ def equilibrium(dbf, comps, phases, conditions, output=None, model=None,
     if len(list_of_possible_phases) == 0:
         raise ConditionError('There are no phases in the Database that can be active with components {0}'.format(comps))
     if len(active_phases) == 0:
-        raise ConditionError('None of the passed phases ({0}) are active. List of active phases: {1}.'.format(phases, list_of_possible_phases))
+        raise ConditionError('None of the passed phases ({0}) are active. List of possible phases: {1}.'.format(phases, list_of_possible_phases))
     comps = sorted(comps)
     if len(set(comps) - set(dbf.elements)) > 0:
         raise EquilibriumError('Components not found in database: {}'.format(','.join(set(comps) - set(dbf.elements))))
--- a/pycalphad/core/equilibrium.py
+++ b/pycalphad/core/equilibrium.py
@@ -6,7 +6,7 @@ from __future__ import print_function
 import warnings
 import pycalphad.variables as v
 from pycalphad.core.utils import unpack_kwarg
-from pycalphad.core.utils import unpack_condition, unpack_phases
+from pycalphad.core.utils import unpack_condition, unpack_phases, filter_phases
 from pycalphad import calculate, Model
 from pycalphad.core.lower_convex_hull import lower_convex_hull
 from pycalphad.core.sympydiff_utils import build_functions as compiled_build_functions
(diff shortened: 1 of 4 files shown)


bocklund:
Having users do this (by updating examples) and applying this closes #129

richardotis:
This sounds like a good idea. Adding this to `equilibrium` (but not `calculate`) would be useful.

richardotis:
If the result of `filter_phases` is empty, we should give the user a nice error message like "No phase with the specified components could be found in the database."
The user probably made a typo somewhere, so the error message should give a hint to check the component list and/or the database.

Semi-related comment: I would be okay with creating a base `EquilibriumError` exception class and then basing all "logical" errors (as opposed to errors from bugs) returning that exception or a subclass of it.

pycalphad/core/equilibrium.py
```diff
@@ -193,7 +193,14 @@ def equilibrium(dbf, comps, phases, conditions, output=None, model=None,
...
     from pycalphad import __version__ as pycalphad_version
-    active_phases = unpack_phases(phases) or sorted(dbf.phases.keys())
+    phases = unpack_phases(phases) or sorted(dbf.phases.keys())
+    # remove phases that cannot be active
+    possible_active_phases = filter_phases(dbf, comps)
+    active_phases = sorted(set(possible_active_phases).intersection(set(phases)))
+    if len(possible_active_phases) == 0:
+        raise ConditionError('There are no phases in the Database that can be active with components {0}'.format(comps))
+    if len(active_phases) == 0:
+        raise ConditionError('None of the passed phases ({0}) are active. List of active phases: {1}.'.format(phases, possible_active_phases))
```
richardotis: Should be "List of possible phases"

pycalphad/tests/test_utils.py
```diff
@@ -0,0 +1,23 @@
...
+The utils test module contains tests for pycalphad utilities.
+"""
+
+from pycalphad import Database
+from pycalphad.core.utils import filter_phases
+
+from pycalphad.tests.datasets import ALNIPT_TDB
+
+ALNIPT_DBF = Database(ALNIPT_TDB)
+
```
richardotis: Needs an integration test to show the equilibrium logic is working. Ideally for both ConditionErrors

Issue #143, opened 2017-11-09
Database float parsing does not work for values with no numbers after a period


bocklund:
Failing test case:

```python
from pycalphad import Database

def test_database_parsing_of_floats_with_no_values_after_decimal():
    """Floating points with no values after the decimal should be properly parsed"""
    tdb_string = """$ The element has no values after the decimal in 5004. causing a failure
ELEMENT CU   FCC_A1           63.546           5004.             33.15      !"""
    dbf = Database.from_string(tdb_string, fmt='tdb')
    assert "CU" in dbf.elements
```

PR #144, opened 2017-11-10
FIX: tdb: Update float parsing regex

richardotis:
FIX: tdb: Update float parsing regex to detect floats with no trailing digits. Fixes gh-143.


richardotis:
Useful site for testing: https://regex101.com/

richardotis:
I think this change caused a 5-10% performance regression in creating Database objects. I'm not sure if it matters that much.

richardotis:
Never mind, I'm getting a lot of scatter in my local unit testing. I think we're fine.

PR #145, opened 2017-11-11
BLD/FIX: appveyor: Bump miniconda to fix broken conda

bocklund:


richardotis:
Looks like conda has dropped py34 support altogether, so the py34 environment is getting updated to py36. The compiler environment is then incompatible.

bocklund:
I updated travis to install and remove gmpy2 so the Mac builds pass. Since this will be documented and we are waiting for sympy to be released for a permanent fix, I think it's a reasonable work around. 

@richardotis ok to merge? If so, I'll rebase my other PRs on this

Issue #139, opened 2017-11-05
Documentation nits for 0.6


richardotis:
Due to a more complicated upstream packaging environment and new dependencies in 0.6, users are more likely than ever to encounter snags when trying to install and use pycalphad. Since upstream patches take a while to make it through a largely volunteer-driven QA process, the best thing we can do is try to mitigate the issue in our installation instructions.

Below is a list of issues users are likely to encounter, and the corresponding workaround. These need to make it into the documentation with the release of 0.6.

- **Problem** (during calculation): `TypeError: argument is not an mpz`
- **Solution**: Once sympy/sympy#13530 is merged and released, users should update sympy. Until then, users can remove the `gmpy2` package from their Python environment to fix the issue. Alternatively, users can set the environment variable `MPMATH_NOGMPY` to a non-zero value.

- **Problem** (during installation via pip): `src/cyipopt.c:239:29: fatal error: IpStdCInterface.h: No such file or directory`
- **Solution**: The library IPOPT is a new dependency in pycalphad 0.6 and requires installation of an external library not available on PyPI. This error message means IPOPT is either not installed or not in your PATH. IPOPT installation instructions are available at https://www.coin-or.org/Ipopt/documentation/node10.html - however, users (especially users on Windows) are strongly encouraged to use the Anaconda installation instructions for pycalphad instead. `conda install pycalphad` and `conda update pycalphad` will automatically install IPOPT on all platforms.

@bocklund We need to wrap up what you started in https://github.com/pycalphad/pycalphad/commit/b67675fad68364224931eea5d7e91fbaaa0a44e3 and include the feedback from #126

Issue #134, opened 2017-09-27
Document the pycalphad xarray datasets


bocklund:

richardotis:
Closest attempt at it so far: https://openresearchsoftware.metajnl.com/articles/10.5334/jors.140/#F3

amjokisaari:
It's definitely helpful (I've been referring to the paper, etc before
asking questions) but it's not complete.  The biggest thing missing here
(for me as a newbie to the project) is that "X" and "Y" are not also
described there.

Issue #128, opened 2017-09-21
Graphs are sometimes cut off when saved


amjokisaari:
The right-hand information (like a legend) or the bottom axis title is sometimes cut off when saving the figure to a png.  Saving as an svg seems to include the information so that it can be moved/edited, but that's not ideal.

bocklund:
Looks like [this SO answer](https://stackoverflow.com/a/25344713/5106932) might provide some insight. Basically in `plt.savefig` use the `bbox_extra_artists` and `bbox_inches='tight'` arguments. Not clear if we can do this automatically, but it's a starting point.

amjokisaari:
Thanks for this.  With Python 2.7.13 and pycalphad 0.5.2, all I need to do is add the bbox_inches='tight' to the savefig command.

bocklund:
I think the only options for enforcing this would be to patch the figure object so that the `bbox_inches='tight'` argument is automatically applied. Since that's not very predictable behavior, the upcoming documentation updates will have some kind of troubleshooting/faq page with this in it.

richardotis:
Is it possible to modify the matplotlib options in a supported way via some global settings object?

bocklund:
Yeah you can technically use the [matplotlib rc](https://matplotlib.org/users/customizing.html):

```
import matplotlib as mpl
mpl.rcParams['savefig.bbox'] = 'tight'
```

The question is whether we *should* do this or not.

richardotis:
How about check the value and warn the user if it isn't set?

bocklund:
On import? I think import spam is kind of annoying

bocklund:
Hard to anticipate if users will be using `savefig`. If they aren't we probably don't want to warn them

richardotis:
I mean, inside eqplot is it possible?

bocklund:
We could warn inside eqplot, but I'm thinking that I call savefig only a small fraction of the time I use eqplot because I work in the notebook so often.

amjokisaari:
Maybe it's not worth trying to fix, but add it into documentation somewhere?  ("Exporting figures" section?)

PR #146, opened 2017-11-12
Update installation and add FAQ. Handle warnings

### LINKED ISSUE #126: Introduction/Installation/Getting Started documentation
Various suggestions to improve the documentation for introduction/installation/getting started:

- The conda installation instructions work for Linux as well as Windows.  This could be slightly confusing for new users.

- May want to indicate that conda may install in root and that users may need to sudo chown -r <usr> /path/to/conda in order to actually create new conda environments

- May wish to include a note about the possible gmpy2 incompatibility (though that seems to be fixed in version 0.5.2)

- May wish to indicate that the stable version does not ship with tests, or add a test suite (to keep someone from wasting time looking for how to test...); or indicate that you should at least run all the examples

- May wish to indicate that the dask warning for v0.5.2 is going to dealt with

- Indicate to users that mpicc may not be appropriate; and to use standard gcc compiler (for example, MOOSE will default load moose-dev-gcc module; this should be module purged before using pycalphad)

- Include information as to whether pycalphad can/should run in parallel or not; and if so, how to do so

- May want to include information as to how long computations for the example problems take (give reasonable expectations for performance)

bocklund:
Fixes gh-126, fixes gh-128, fixes gh-134, and fixes gh-139

Regarding points in #126:

* Update installation to first suggest anaconda installation for all platforms. Also update the development mode, because conda-build is not well maintained for using development mode in environments. Suggested to use pip or setup.py in conda/conda-build#1992

This PR updates the installation guide with step by step instructions for all platforms, but emphasizing Anaconda. It is now more clear that you can use either distribution to install pycalphad. 

The installation includes the caveats of the Ipopt installation situation for pip installation and develop mode, as well as a Anaconda alternative to develop mode. 

* Add page describing parallelism with dask

This is in the newly added FAQ

* Estimate of computation time in the examples (roughly ~1 minute accuracy)

A brief statement of computation time and the implications for binplot/ternplot were added to the FAQ.

* (Possible) mention which warnings are normal to see, including convergence failures. Where should this go?

I did not handle the convergence failures (Ipopt should make them more rare). This commit explicitly adds code to filter out the NumPy warnings and fixes the dask import warning that users have been seeing and bumps the dask version accordingly. (Python 3.7 will make async a keyword (see https://bugs.python.org/issue30406). This means dask had to rename their async module to local to avoid having a module name be a keyword and users have been seeing warnings about the import. This commit updates the import from dask.async.get_sync to dask.local.get_sync and bumps the requirements to dask>=0.15.). 

#139: Addressed simultaneously with #126 

#128: added to FAQ

#134: Added a UsingCalculationResults example notebook (and generated docs) that goes through the calculate and equilibrium datasets and describes them fully. Also provides some brief examples of how one would select and mask using xarray.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -3,6 +3,14 @@
 FAQ
 ===
 
+What units does pycalphad use?
+------------------------------
+
+* All units are SI units.
+* Molar quantities are used for state variables, e.g. energy has units ``J/mol``.
+* Composition and site occupancies are mole fractions.
+
+
 Is any parallelism supported in pycalphad?
 ------------------------------------------
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -54,3 +54,15 @@ The fix has been copmleted, but not yet released. While the fix is not released,
 removing the gmpy2 package from their Python environment (e.g.
 ``conda remove --force gmpy2``) will fix the error. Alternatively, setting the
 environment variable ``MPMATH_NOGMPY`` to a non-zero value will fix the error.
+
+
+Text is sometimes cut off when saving figures
+---------------------------------------------
+
+Occasionally when saving images with the matplotlib function ``plt.savefig``, axis titles and legends are cut off.
+
+This can be fixed
+
+* Per function call by passing ``bbox_inches='tight'`` keyword argument to ``plt.savefig``
+* Locally by running ``import matplotlib as mpl; mpl.rcParams['savefig.bbox'] = 'tight'``
+* Permanently by adding ``savefig.bbox : tight`` to your `matplotlibrc file <https://matplotlib.org/users/customizing.html>`_.
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -1,0 +1,35 @@
+.. title:: FAQ
+
+FAQ
+===
+
+Is any parallelism supported in pycalphad?
+------------------------------------------
+
+Equilibrium calculations in pycalphad can be parallelized using `dask <http://dask.pydata.org/en/latest/>`_ out of the box.
+Several schedules are supported in `dask <http://dask.pydata.org/en/latest/scheduler-overview.html>`_
+and some have been `benchmarked in pycalphad <https://github.com/pycalphad/pycalphad/issues/101>`_,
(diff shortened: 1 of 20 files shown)


bocklund:
There were some minor code changes, but mostly docs here. I suggest checking out this branch locally and `cd docs`, then building the docs and inspecting them with something like ``sphinx-build . _build/html && cd _build/html && python3 -m http.server && cd ../..``

docs/faq.rst
```diff
@@ -0,0 +1,68 @@
...
+are mapped by the tieline points of the two phase regions, so unless there are
+two phase regions in a very small composition range, only coarse composition
+grids are required for phase diagram calculations.
+
+
+``TypeError: argument is not an mpz`` during a calculation
+----------------------------------------------------------
+
+This is an upstream bug in sympy, where floats are unable to be pickled.
+The fix has been copmleted, but not yet released. While the fix is not released,
```
richardotis: spelling

richardotis:
- Al-Fe phase diagram picture on the index page is a broken link. (It's been that way a while.)
- `Anaconda is a scientific Python distribution by Continuum Analytics.` They call themselves `Anaconda, Inc.` now.
- `This can be fixed` in the FAQ needs a trailing colon.
- All the examples need new titles; shorter would be better. "Calculation" is in almost every one and doesn't help the user very much. (Almost every feature of pycalphad involves calculation.)
- Using Calculation Results: Headings should be `calculate() Results` and `equilibrium() Results` to be unambiguous
-- Masking (mentioned at bottom) needs an example
-- It would also be nice to mention `.values` for conversion to a NumPy array

bocklund:
I think with the state of the solver, Al-Ni is a better looking example to show rather than Al-Fe. As far as the masking, the `where` is what that is: everything that does not match the condition is `nan`.

richardotis:
Either phase diagram is fine. We just need to make sure the link on the front page is correct.

bocklund:
TODO: Add notes on units in the FAQ

bocklund:
This is ready for another look

richardotis:
Merge when ready.

PR #149, opened 2017-11-30
FIX: tdb: Allow TEMPERATURE_LIMITS and commas in low-temperature limits.

richardotis:


bocklund:
Does the test have a `TEMPERATURE_LIMIT` command in it? It's not obvious to me where it is

richardotis:
1. Fixed
2. Already supported, updated the test to show this
3. TC will use 298.15 as a low-T default, but seems to ignore the limit anyway and extrapolate outside of them. pycalphad using 0 will give results more in line with user expectations

PR #156, opened 2017-12-12
ENH: JIT performance improvement

richardotis:
The `Model` class is JIT-compiled by an autowrap-based utility, which is primarily located in `core\sympydiff_utils.py` and `core\custom_autowrap.py`. A substantial portion of the startup time for a new `Model` used in a calculation is in the JIT. Here the performance is improved via profiling, where redundant or unnecessary computations are eliminated. In particular, since the new IPOPT-based solver does not require direct Hessian computation, Hessian codegen is disabled. This, and a few other tweaks,lead to a 50-70% reduction in the first-run time for `Model`-based computations.

### DIFF
--- a/pycalphad/core/custom_autowrap.py
+++ b/pycalphad/core/custom_autowrap.py
@@ -160,6 +160,8 @@ class C89CodePrinter(CCodePrinter):
             close_lines.append("}")
         return open_lines, close_lines
 
+    _print_ComplexInfinity = CCodePrinter._print_Infinity
+
 
 class CustomCCodeGen(CCodeGen):
     def get_prototype(self, routine):
--- a/pycalphad/core/custom_autowrap.py
+++ b/pycalphad/core/custom_autowrap.py
@@ -161,6 +161,7 @@ class C89CodePrinter(CCodePrinter):
         return open_lines, close_lines
 
     _print_ComplexInfinity = CCodePrinter._print_Infinity
+    _print_NegativeComplexInfinity = CCodePrinter._print_NegativeInfinity
 
 
 class CustomCCodeGen(CCodeGen):
--- a/pycalphad/core/custom_autowrap.py
+++ b/pycalphad/core/custom_autowrap.py
@@ -73,6 +73,8 @@ except ImportError:
 from sympy.utilities.codegen import (AssignmentError, InOutArgument, InputArgument, OutputArgument,
                                      ResultBase, Result, CodeGenArgumentListError,
                                      CCodeGen, CodeGenError, Routine, Variable)
+from sympy.codegen.ast import Assignment
+from sympy.printing.precedence import precedence
 from sympy.core import Symbol, S, Tuple, Equality
 from sympy.core.compatibility import is_sequence
 from sympy.tensor import Idx, Indexed, IndexedBase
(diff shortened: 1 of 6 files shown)


pycalphad/core/sympydiff_utils.py
```diff
@@ -143,18 +143,17 @@ class f(Function):
         restup.append(AutowrapFunction(args, Eq(y[i], f(*args_with_indices))))
     if include_grad or include_hess:
         diffargs = (inp_nobroadcast, params)
-        # Replacing zoo's is necessary because sympy's CCodePrinter doesn't handle them
```
bocklund: Does this set a new minimum required SymPy version?

pycalphad/core/custom_autowrap.py
```diff
@@ -151,6 +160,9 @@ def _get_loop_opening_ending(self, indices):
             close_lines.append("}")
         return open_lines, close_lines
 
+    _print_ComplexInfinity = CCodePrinter._print_Infinity
```
bocklund: Do we need a `_print_NegativeComplexInfinity = CCodePrinter._print_NegativeInfinity`?

pycalphad/core/custom_autowrap.py
```diff
@@ -276,6 +288,137 @@ def dump_h(self, routines, f, prefix, header=True, empty=True):
 
     dump_fns = [CCodeGen.dump_c, dump_h]
 
+    def routine(self, name, expr, argument_sequence, global_vars):
```
bocklund: Would it be possible to add some commentary on why we need to override this and what the major changes are compared to the CodeGen base class?

pycalphad/core/equilibrium.py
```diff
@@ -267,7 +267,9 @@ def equilibrium(dbf, comps, phases, conditions, output=None, model=None,
                 undefs = list(out.atoms(Symbol) - out.atoms(v.StateVariable) - set(param_symbols))
                 for undef in undefs:
                     out = out.xreplace({undef: float(0)})
-                cf, gf, hf = build_functions(out, tuple([v.P, v.T] + site_fracs), parameters=param_symbols)
+                cf, gf = build_functions(out, tuple([v.P, v.T] + site_fracs),
```
bocklund: @richardotis can you clarify: we aren't building Hessians, but handling for Hessian code are still around. Can you be more specific on your comments in the main PR comment?

pycalphad/core/custom_autowrap.py
richardotis: I don't think so, but it couldn't hurt. This code is necessary because SymPy differentiation can sometimes produce complex infinity symbols. I think if we were consistent in enforcing assumptions those piecewise paths wouldn't be generated, but ignoring the imaginary part seems to work in this application.

pycalphad/core/custom_autowrap.py
richardotis: Added. For reference: http://linediff.com/?id=5a3be237687f4b2d2a8b4567

pycalphad/core/sympydiff_utils.py
richardotis: No.

pycalphad/core/equilibrium.py
richardotis: I hesitated to remove the Hessian handling code altogether, but since history lives on in Git, I should push forward and clean it up.

Issue #159, opened 2018-02-02
Error when Calculating Enthalpy


mxf469:
When calculating the enthalpy over multiple phases "ValueError: shapes (10,4) and (3,) not aligned: 4 (dim 1) != 3 (dim 0)"

```
from pycalphad import equilibrium, Database
from pycalphad.plot.utils import phase_legend
from pycalphad import Model
import pycalphad.variables as v

db = Database('Cr-Fe-Ni.tdb')

res = equilibrium(db, ['CR', 'FE', 'VA'], ['BCC_A2', 'SIGMA'], {v.X('FE'): [0.5], v.T: (400, 500, 21), v.P: 101325}, model=Model, output=['HM'])
for composition in res.X_FE:
    print(res.sel(X_FE=composition))
```
[New folder.zip](https://github.com/pycalphad/pycalphad/files/1690956/New.folder.zip)

--
Edited to fix formatting on code block - Brandon

richardotis:
I can reproduce this error on e50b618d35c3db6a5d880aa44b158db9c7963dcb (develop head)

PR #160, opened 2018-02-02
FIX: eqcalculate: Property calculation for "partially loaded" phase models

richardotis:
Compute degrees of freedom correctly when only a subset of constituents in a phase's sublattice model are used in a calculation. Fixes gh-159.

@bocklund Do you think it needs a test?


richardotis:
The AppVeyor tests are failing for an unrelated reason, so I'm going to merge this.

PR #166, opened 2018-03-20
FIX: fix Species Python 2 unicode support

### LINKED ISSUE #165: Py27 unicode strings break Species parsing
For unicode strings, the following `if` is not True and species parsing is broken because we end up trying to index single character strings

```python
        if isinstance(arg, str):
            try:
                parse_list = chemical_formula.parseString(arg.upper())
            except ParseException:
                return None
        else:
            parse_list = arg
        new_self.name = name
        new_self.charge = parse_list[1]
        parse_list = parse_list[0]
        new_self.constituents = {parse_list[i]: parse_list[i+1] for i in range(0, len(parse_list), 2)}
        return new_self
```

Solution is to use `six` (which is currently an implicit dependency) or to directly solve the problem by:

```python
import sys
if sys.version_info[0] >= 3:
    string_type = str
else:
    string_type = basestring
```

and replace `if isinstance(arg, str):` with `if isinstance(arg, string_type):`

bocklund:
Closes #165

### DIFF
--- a/pycalphad/tests/test_variables.py
+++ b/pycalphad/tests/test_variables.py
@@ -1,0 +1,10 @@
+"""
+Test variables module.
+"""
+
+from pycalphad import variables as v
+
+
+def test_species_parse_unicode_strings():
+    """Species should properly parse unicode strings."""
+    s = v.Species(u"MG")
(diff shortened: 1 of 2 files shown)


bocklund:
I don't have a py2 environment set up, so I am just making sure CI fails. I have a commit waiting with the proposed fix

bocklund:
This is ready for a look

PR #176, opened 2018-06-05
ENH: Parallelize calculation of objective function

bocklund:
These changes define a parallel wrapper around PhaseRecord.obj that uses OpenMP to parallelize calculation of the objective function and uses that by default in `_compute_phase_values`.

I tested two main cases and some scaling between them

1. Large case (same as #175, ~30M points): gained a 2x speedup of the same benchmark from #175 (on an Intel i7)
2. Small case: running calculate with a 1 sublattice phase and `pdens=1`, giving 4 total points. The parallel and serial versions of `calculate` are within 2x for small cases . As the number of points grows and sampling becomes more expensive, they eventually switch and parallel seems to be about 2x faster. Note that on the curve there is some constant time that grows due to copying arrays, so the speedup of the objective is actually more exaggerated than shown here.

scaling

I chose to have parallelism on by default, but I also put in an option to turn off parallelization. My reasoning for default is that the time difference is only a few ms. The use case with the biggest potential pitfall is if users are running `calculate` for small point densities in tight loops (instead of using broadcasting features). This does not seem too likely to me. There's also an option to have some heuristic so that having parallelism turned on still uses serial mode unless there are more than 100k point sets (for example).

I added an option for parallelization to be turned off because uses may not want it if they are performing equilibrium calculations in parallel, such as in ESPEI or using a dask scheduler, or for other special cases like the above.

This also knocks off a chunk of time for large calculate calls and can go a long way depending on the machine without having to fully Cythonize `_compute_phaes_values` yet. That still might be worth doing in the future.

@richardotis what do you think? I have some reservations about parallelism as a default.


bocklund:
Might need to fix some compiler issues on Windows https://cython.readthedocs.io/en/latest/src/userguide/parallelism.html

bocklund:
The tests seem to be passing on Windows, so maybe not.

richardotis:
Could you try some of these results with smaller chunk sizes?

bocklund:
I can look. My initial guess is that it will actually hurt performance.

The benefit of chunking this way is that each thread has only one chunk to process. We are making the assumption that the execution time of each vectorized chunk is roughly equal (which I think is fair). If you have potentially unequal chunks, then you incur some possible looping overhead for too small chunk sizes and run the risk of having an odd number of chunks and you have multiple workers waiting.

bocklund:
FWIW: line profiling indicates that the execution of the objective calls for the large test case went from 35.9s to 12.2s, so about 3x speedup.

bocklund:
Turns out I was not right... Still gathering data.

bocklund:
chunking


Here's some data I collected on my system with the case of 30M points and 8 threads. Each data point is an average of 20 runs

richardotis:
Amateur computer science incoming.

My guess is this has to do with what we can fit in the CPU caches. If we have N points and the model has to perform K "operations" (not sure how to define this formally), if size(K) >> size(N) then vectorization has reduced benefit, since we have to do a lot of copying. By breaking the N points up into smaller chunks, we're able to get better data locality on each of the cores for performing all the operations, since we can fit both the input arrays and the intermediate results on the cache.

richardotis:
NERSC has a good explanation here: http://www.nersc.gov/users/computational-systems/edison/programming/vectorization/

"Performance improvement maximizes when all data is in caches. If data has to be moved from memory to caches frequently, performance gain with vectorization can become obscured by the cost for data movement...For even larger array dimensions, the cost associated with data movement dominates so much that vectorization effect is almost non-existent or vectorization has an adverse effect."

bocklund:
I think we should not proceed with this work for now.

1. This has possible conflicts with other forms of parallelization and it's not clear how to best partition the work without inspecting the state of available workers.
2. The API in the calculate function seems so far from where this actually occurs. It's not needed for the vast majority of cases and not very discoverable for when it is needed.
3. There are other places in this path that could be optimized to gain some performance on a single thread. Specifically, not broadcasting the points and state variables arrays and copying chunks of memory.
4. This could see a bigger boost from paralleling on a GPU, so it might be worth waiting for some time to work on this to develop a more general solution.

I'm going to close this because the changes in this PR are fairly trivial and will be easy to reproduce if we go down this path later.

PR #177, opened 2018-06-08
ENH: Enable custom solver

### LINKED ISSUE #174: changing constants in equilibrium calculation without reinstaling
In the past (version 0.5) it was not possible to change constants such as 
MAX_SOLVE_DRIVING_FORCE  without reinstalling pycalphad.
Now, I find that there is a constants.py module that contains several constants.
I tried to run at the head of my program:
import pycalphad.core.constants
pycalphad.core.constants.MIN_SITE_FRACTION = 1e-14
pycalphad.core.constants.MIN_PHASE_FRACTION = 1e-7
pycalphad.core.constants.COMP_DIFFERENCE_TOL = 1e-6
pycalphad.core.constants.MAX_SOLVE_DRIVING_FORCE = 1e-6
pycalphad.core.constants.MAX_SOLVE_ITERATIONS = 4000
pycalphad.core.constants.MIN_SOLVE_ENERGY_PROGRESS = 1e-6

but could not see a change in the performance.
Is this the correct method of changing constants?
Can it now be done without reinstalling pycalphad?

Best regards,
Eli

bocklund:
* Makes a new `SolverBase` class that `InteriorPointOptimzer` subclasses. All subclasses of `SolverBase` must implement a `solve` method. 
* `solve` must take a `pycalphad.core.problem.Problem` class and return a `pycalphad.core.solver.SolverResult` named tuple.
* `equilibrium` takes a `solver` argument that takes an instance of `SolverBase` or a subclass, which is passed to the `_solve_eq_at_conditions` function. 
* Pull out instantiating solver objects from inside a while loop nested in `_solve_eq_at_conditions` to being instantiated either outside `equilibrium` or at the top level of `_solve_eq_at_conditions`.

The idea here is that users can

1. Instantiate their own `InteriorPointOptimizer` with custom IPOPT options (with an easy interface to set them - just a dict)
2. Create subclasses of `SolverBase` that use different solvers.

Can close #174 

This also removes some dead code and adds/improves documentation. `CompiledModel` and the now unused global constants in `pycalphad.core.constants` are removed.


bocklund:
@richardotis made a comment in #174:

> User-developed custom Solver subclasses can have any signature they want, of course, depending on what makes sense for their work. 

The current method proposed here is that the `solve` cannot have any signature, but must take some kind of Problem instance. Instead of having any signature for a `solver.solve` method, users should encode any additional information options in their `solver.__init__` method at instantiation time or in a custom `Problem` class, which is passed directly to `solver.solve`. 

> I think the configuration API for the builtin Solver should be very limited, to avoid users hard-coding solver constants for an optimization package which is subject to change in the future.

IMO we don't provide guarantees that any options will work, but I think it's within scope of pycalphad as a tool for method development to allow users to do this. We should make it easy for users (particularly developers) to tinker and possibly optimize for their use case (e.g. maybe you are developing a high-throughput simulation and want to turn down tolerance to meet performance constraints). I also hope that our test suite continues to develop into a benchmark of solver correctness that can be part of validating solvers and global minimization algorithms.

PR #182, opened 2018-08-13
MAINT: Remove dead constraint calculation code

bocklund:
This constraint calculation code has been moved to the Problem class

PR #184, opened 2018-08-17
FIX: Max phases by Gibbs phase rule accommodated 

### LINKED ISSUE #170: Error for too many phases during equilibrium calculations
Some conditions trigger an error where too many phases are found during equilibrium calculations.

I don't have a traceback or example for this right now, but I wanted to open this issue as a reminder that this exists and add the example next time I come across the error.

bocklund:
* Mostly straightforward fix that closes #170 by padding the `vertex` to be of size `C+2` instead of `C` (for C components). 
* Includes a test of a eutectic point in Pb-Sn. 
* Modestly improved hyperplane documentation

This requires a hack in `hyperplane` because hyperplane constructs a C-simplex hyperplane corresponding to the chemical potentials, however the phase fraction and point index arrays are now padded to accommodate C+2 phases instead of C phases. So I set the point index to zero (which pulls in the phase, site fracs, etc. associated with that point) and the phase fraction of those points to zero. I'm not sure if this was correct or what implications there are for doing this, but the tests pass...?


bocklund:
I'm not expecting the tests to pass until this is rebased on #182 that updates SymPy requirements

richardotis:
Does this still work if you set the maximum number of phases to c+1 instead of c+2? I support this change, but I want to note that the current implementation already respects the Gibbs phase rule, which is `c + 2 - 2 = c` phases can coexist at once. We lose two degrees of freedom from fixing T and P.

The reason we want to artificially increase the maximum phases allowed is because we want to accommodate the degenerate degree of freedom at the invariant reactions. (The solution this change gives will be "correct," but the phase fractions won't be meaningful since multiple values would give the same Gibbs energy on the invariant line.)

bocklund:
I changed C+2 to C+1 and tests pass locally.

pycalphad/core/hyperplane.pyx
```diff
@@ -84,6 +84,12 @@ cpdef double hyperplane(double[:,::1] compositions,
     Examples
     --------
     None yet.
+
+    Notes
+    -----
+    M: number of energy points that have been sampled
+    N: number of components
+    P: N+2, max phases by gibbs phase rule
```
richardotis: Needs to be changed to N+1. Maybe choose another letter besides P to represent this quantity, so it isn't easy to confuse with a thermodynamic variable

pycalphad/core/hyperplane.pyx
```diff
@@ -156,7 +162,10 @@ cpdef double hyperplane(double[:,::1] compositions,
...
         idx = best_guess_simplex[i]
         out_energy += fractions[saved_trial, i] * energies[idx]
-    result_fractions[:] = fractions[saved_trial, :]
+    result_fractions[:num_components] = fractions[saved_trial, :]
     chemical_potentials[:] = candidate_potentials
-    result_simplex[:] = best_guess_simplex
-    return out_energy
\ No newline at end of file
+    result_simplex[:num_components] = best_guess_simplex
+    # Hack to enforce Gibbs phase rule, shape of result is comp+2, shape of hyperplane is comp
```
richardotis: `comp+1` now

pycalphad/core/hyperplane.pyx
richardotis: (though we do already use N, so...)

bocklund: I think defining them prevents confusion for overloading (but not having any labels was confusing).

PR #187, opened 2018-08-28
ENH: Allow solution refinement by the Ipopt solver to be disabled.

bocklund:


richardotis:
Does it need a test?

bocklund:
Maybe an integration test to validate that a certain point density gives a certain result? Do you have another suggestion?

richardotis:
An integration test that exercises the code path and makes sure we return an answer without raising is fine.  We don't make any accuracy promises with this mode.

bocklund:
I have a sketch of a test written, but this has made me realize that the chemical potentials are all 0.0. The other properties seem correct.

bocklund:
Py3.5 tests pass due to some dependency luck, so this should just need a rebase on #192 to fix other platforms

Issue #189, opened 2018-09-25
Refactor the creation of callables in calculate and equilibrium


bocklund:
ESPEI uses this code to generate callables.

We could refactor `equilibrium` and `calculate` to call this function, making them easier to read and reason about. This would also allow users to write tighter loops through calculate and equilibrium, avoiding relatively expensive calls to sympy atoms.

This issue is mostly a reminder to do this for the next release. It just needs to be integrated and cleaned up.

```python
def build_callables(dbf, comps, phases, model=None, param_symbols=None, output='GM', build_gradients=True):
    """
    Create a dictionary of callable dictionaries for phases in equilibrium

    Parameters
    ----------
    dbf : pycalphad.Database
        A pycalphad Database object
    comps : list
        List of component names
    phases : list
        List of phase names
    model : dict or type
        Dictionary of {phase_name: Model subclass} or a type corresponding to a
        Model subclass. Defaults to ``Model``.
    param_symbols : list
        SymPy Symbol objects that will be preserved in the callable functions.
    output : str
        Output property of the particular Model to sample
    build_gradients : bool
        Whether or not to build gradient functions. Defaults to True.

    Returns
    -------
    dict
        Dictionary of keyword argument callables to pass to equilibrium.

    Notes
    -----
    Based on the pycalphad equilibrium method for building phases as of commit 37ff75ce.

    Examples
    --------
    >>> dbf = Database('AL-NI.tdb')
    >>> comps = ['AL', 'NI', 'VA']
    >>> phases = ['FCC_L12', 'BCC_B2', 'LIQUID', 'AL3NI5', 'AL3NI2', 'AL3NI']
    >>> eq_callables = eq_callables_dict(dbf, comps, phases)
    >>> equilibrium(dbf, comps, phases, conditions, **eq_callables)
    """
    comps = sorted(unpack_components(dbf, comps))
    pure_elements = get_pure_elements(dbf, comps)

    eq_callables = {
        'massfuncs': {},
        'massgradfuncs': {},
        'callables': {},
        'grad_callables': {},
        'hess_callables': {},
    }

    models = unpack_kwarg(model, default_arg=Model)
    param_symbols = param_symbols if param_symbols is not None else []
    # wrap param symbols in Symbols if they are strings
    if all([isinstance(sym, string_types) for sym in param_symbols]):
        param_symbols = [Symbol(sym) for sym in param_symbols]
    param_values = np.zeros_like(param_symbols, dtype=np.float64)

[...]

PR #192, opened 2018-10-29
MAINT: Refactor callables creation in equilibrium() and calculate()

richardotis:
Fixes gh-189.

- Creates a new `codegen` subdirectory; move some existing codegen functionality there
- New `build_callables` function

FIX: equilibrium/calculate: Properly filter phases that cannot be built with build_callables()
MAINT: Remove Hessian callable support and clean up PhaseRecord pickling
FIX: solver: Tweak to attempt fix for stochastic failure in test_eq_issue43_chempots_misc_gap
BLD: appveyor: Build fixes
BLD: New implicit dependency on numpy>=1.13 from xarray
MAINT: dask>=0.18 compatibility
MAINT: core.utils: Add get_pure_elements from ESPEI
MAINT: codegen: Move sympydiff_utils and custom_autowrap to separate subdirectory


richardotis:
One possible TODO might be to turn the `callables` nested dict object and turn it into a real class, to make it a little more self-documenting

richardotis:
Thank you for this. I am now working to resolve a state issue with `build_callables` and `parameters`, but we need to discuss some of it before the fix can be completed.

`Check that passing callables should skip the build phase, but use the values from 'VV0000' in the tdb`
I am not sure that the behavior here is wrong but let's discuss it. `callables` now keeps a dict of PhaseRecords by default, and there the specified parameter values are stored. If I call `build_callables(dbf, comps, phases, parameters={'VV0000': 20000})`, and use that object to do calculations, I expect the value of `VV0000` to be treated as 20000, right?

richardotis:
> The ideal case (IMO) is that `build_callables` should be able to resolve parameters just in time without triggering a recompile.

I think that's how things work now (as of 52fe6c9 at least). We only disagree on the resolution order. This is how it looks as implemented in 52fe6c9:
1. Any parameters passed to `calculate` or `equilibrium`
2. The parameter values passed to `build_callables`

PhaseRecords don't know which symbols (or even which Database) their parameters correspond to, so if we wanted to change the default parameter values, we'd need to create additional state somewhere. But using the values passed into `build_callables` seem like a natural default to me; otherwise, what's the point of passing in default values at all?

Alternatively, we could add support for passing in a list of symbols to the parameters kwarg (instead of a dict), and choose defaults based on the values in the Database -- _however_, entries in `Database.symbols` are not guaranteed to be single, real values; in general, they may be functions. The logic for making this work would get messy. I think a good workaround for this functionality would be to choose your default values in `build_callables` to be the same as those in `Database.symbols`.

bocklund:
I agree with your resolution order and I think your current solution is fine. Any looking into the Database in `build_callables` would just be a convenience to passing them programmatically or by hand. 

I have this working for ZPF error in ESPEI, so there are no blockers for my review of this PR.

pycalphad/model.py
```diff
@@ -111,7 +112,13 @@ def wrap_symbol(obj):
             else:
                 return Symbol(obj)
         if parameters is not None:
-            symbols.update([(wrap_symbol(s), val) for s, val in parameters.items()])
+            if isinstance(parameters, dict):
+                symbols.update([(wrap_symbol(s), val) for s, val in parameters.items()])
+            else:
+                # Lists of symbols that should remain symbolic
+                for s in parameters:
+                    symbols.pop(s)
```
bocklund: This should probably be `symbols.pop(wrap_symbol(s))` in case strings are passed.

bocklund:
Your recent changes to the build callables seem to work, but there's a couple chances for you to shoot yourself in the foot.

1. If you instantiate your models before passing them to `build_callables`, then they won't have the symbolic parameters popped (unless you explicitly pass parameter symbols in the `Model(dbf, comps, phases, parameters=my_symbols_list`), but the catch is now you *have* to pass in parameters to your `calculate` or `equilibrium`, otherwise it gets evaluated to zero. 

I think this is the current behavior:

I. If no Model instance is passed to `build_callables` (e.g. the class `Model` or a subclass are passed), then things are replaced as expected.
II. If Model instances with `Model(..., parameters=None)` or `Model(..., parameter=<dict>)`, the values are essentially hardcoded and any parameters passed to `calculate`, `equilibrium` and `build_callables` don't take effect
III. If the `Model(..., parameters=<list>)`, then `build_callables` works as intended

These cases are demonstrated in the script at the end.

2. If `callables` are passed to equilibrium or calculate, anything passed to the `model` kwarg will be disregarded. Perhaps we  should warn the user. 


```python
DO_ASSERTS = False

from pycalphad.tests.datasets import ALFE_TDB
from pycalphad import Database, calculate, Model, variables as v
from pycalphad.codegen.callables import build_callables
import numpy as np
from sympy import Symbol
comps = ["AL", "NI"]
OUTPUT = 'HM_MIX'

TDB = """
ELEMENT AL   FCC_A1                    26.981539   4577.296    28.3215!
ELEMENT NI   FCC_A1                    0   0    0!
FUNCTION VV0000     298.15   5000;              6000.00 N !
FUNCTION VV0001     298.15   10;              6000.00 N !

PHASE FCC_A1  %  1 1 !
   CONSTITUENT FCC_A1  :AL,NI:  !
   PARAMETER G(FCC_A1,AL;0)      298.15   -10000;       2900 N !
   PARAMETER G(FCC_A1,NI;0)      298.15   -10000;       2900 N !
   PARAMETER L(FCC_A1,AL,NI;0)      298.15   4*VV0000#+4*VV0001#*T;       2900 N !
"""
dbf = Database(TDB)
phases = ['FCC_A1']
conds = {v.P: 101325, v.T: 500}

print('Case 0: (no callables built) preinstantiating the Model without passing parameters substitutes correctly in calculate...')
models = {'FCC_A1': Model(dbf, comps, phases[0])}

calc_res = calculate(dbf, comps, phases, conds, output=OUTPUT, points=[[0.5, 0.5]], parameters={'VV0000': 10000})
if DO_ASSERTS: assert calc_res.HM_MIX.values.squeeze() == 10000.0
if not DO_ASSERTS: print(np.isclose(calc_res.HM_MIX.values.squeeze(), 10000.0))
print("Parameter value we got: ", calc_res.HM_MIX.values.squeeze())


print('Case 1: preinstantiating the Model without passing parameters substitutes correctly in build_callables...')
models = {'FCC_A1': Model(dbf, comps, phases[0])}

# build callables with a parameter of 20000.0
callables = build_callables(dbf, comps, phases, model=models, parameters={'VV0000': 20000}, output=OUTPUT, build_gradients=False)

calc_res = calculate(dbf, comps, phases, conds, output=OUTPUT, points=[[0.5, 0.5]], callables=callables, parameters={'VV0000': 10000})
if DO_ASSERTS: assert calc_res.HM_MIX.values.squeeze() == 10000.0
if not DO_ASSERTS: print(np.isclose(calc_res.HM_MIX.values.squeeze(), 10000.0))
print("Parameter value we got: ", calc_res.HM_MIX.values.squeeze())


print('Case 2: preinstantiating the Model with popping the parameters we\'ll substitute are substituted correctly in build_callables...')
models = {'FCC_A1': Model(dbf, comps, phases[0], parameters=[Symbol('VV0000')])}

# build callables with a parameter of 20000.0
callables = build_callables(dbf, comps, phases, model=models, parameters={'VV0000': 20000}, output=OUTPUT, build_gradients=False)

calc_res = calculate(dbf, comps, phases, conds, output=OUTPUT, points=[[0.5, 0.5]], callables=callables, parameters={'VV0000': 10000})
if DO_ASSERTS: assert calc_res.HM_MIX.values.squeeze() == 10000.0
if not DO_ASSERTS: print(np.isclose(calc_res.HM_MIX.values.squeeze(), 10000.0))
print("Parameter value we got: ", calc_res.HM_MIX.values.squeeze())

print('If no parameters are called, we adopt what was in the Database (or in build_callables, depending on the decided resultion order)...')
# What actually happens is that if we don't pass any parameters, the value just goes to zero instead of taking on the parameter in the database.
# Possible way to shoot yourself in the foot, probably not intended behavior
calc_res = calculate(dbf, comps, phases, conds, output=OUTPUT, points=[[0.5, 0.5]], callables=callables)
if DO_ASSERTS: assert calc_res.HM_MIX.values.squeeze() == 5000.0
if not DO_ASSERTS: print(np.isclose(calc_res.HM_MIX.values.squeeze(), 5000.0))
print("Parameter value we got: ", calc_res.HM_MIX.values.squeeze())


print('Case 3: preinstantiating the Model with parameter dict substities correctly...')
models = {'FCC_A1': Model(dbf, comps, phases[0], parameters={'VV0000': 8000})}

# build callables with a parameter of 20000.0
callables = build_callables(dbf, comps, phases, model=models, parameters={'VV0000': 20000}, output=OUTPUT, build_gradients=False)

print('Override the parameters with calculate works...')
calc_res = calculate(dbf, comps, phases, conds, output=OUTPUT, points=[[0.5, 0.5]], callables=callables, parameters={'VV0000': 10000})
if DO_ASSERTS: assert calc_res.HM_MIX.values.squeeze() == 10000.0
if not DO_ASSERTS: print(np.isclose(calc_res.HM_MIX.values.squeeze(), 10000.0))
print("Parameter value we got: ", calc_res.HM_MIX.values.squeeze())

print('With no parameters passed, we adopt what was passed to the Model...')
calc_res = calculate(dbf, comps, phases, conds, output=OUTPUT, points=[[0.5, 0.5]], callables=callables)
if DO_ASSERTS: assert calc_res.HM_MIX.values.squeeze() == 8000.0
if not DO_ASSERTS: print(np.isclose(calc_res.HM_MIX.values.squeeze(), 8000.0))
print("Parameter value we got: ", calc_res.HM_MIX.values.squeeze())


```

With output

```
Case 0: (no callables built) preinstantiating the Model without passing parameters substitutes correctly in calculate...
True
Parameter value we got:  10000.0
Case 1: preinstantiating the Model without passing parameters substitutes correctly in build_callables...
False
Parameter value we got:  5000.0
Case 2: preinstantiating the Model with popping the parameters we'll substitute are substituted correctly in build_callables...
True
Parameter value we got:  10000.0
If no parameters are called, we adopt what was in the Database (or in build_callables, depending on the decided resultion order)...
False
Parameter value we got:  20000.0
Case 3: preinstantiating the Model with parameter dict substities correctly...
Override the parameters with calculate works...
False
Parameter value we got:  8000.0
With no parameters passed, we adopt what was passed to the Model...
True
Parameter value we got:  8000.0
```

pycalphad/codegen/callables.py
```diff
@@ -0,0 +1,173 @@
...
+
+    Parameters
+    ----------
+    dbf : Database
+        A Database object
+    comps : list
+        List of component names
+    phases : list
+        List of phase names
+    model : dict or type
```
bocklund: See comment in the issue about instances of `Model` or subclasses here.

pycalphad/codegen/callables.py
```diff
@@ -0,0 +1,173 @@
...
+    comps = sorted(unpack_components(dbf, comps))
+    pure_elements = get_pure_elements(dbf, comps)
+
+    callables = callables if callables is not None else {}
+    _callables = {
+        'massfuncs': {},
+        'massgradfuncs': {},
+        'callables': {},
+        'grad_callables': {},
+        'hess_callables': defaultdict(lambda: None),
```
bocklund: Can/should we remove the Hessian callables?

richardotis:
It seems like the easiest way to resolve the footguns with Model instances is to remove the ability to pass your own Model instances at all, that is, only allow the class/subclass itself to be passed and require `build_callables` to do the instantiation. Would doing that break any important functionality?

pycalphad/codegen/callables.py
richardotis: Yes but I think its usage is annoyingly hardcoded into a few function signatures. Is it used in ESPEI?

richardotis:
Alright, I'll leave that alone for now then.

pycalphad/codegen/callables.py
bocklund: Everything in ESPEI just passes around the callables from `build_callables` now

richardotis:
@bocklund If you want to do one more run against ESPEI's test suite, I think this is the version I will merge once tests pass.

bocklund:
ESPEI works. Looks good to me.

Issue #197, opened 2018-11-14
Blockers for 0.7.1


richardotis:
- [x] Update FAQ
- [x] Fix scaling on Al-Ni phase diagram on front page of website
- [x] Update Changelog

@bocklund Please chime in with any items

bocklund:
* Stop removing gmpy2 in travis tests (line 45-47 of `.travis.yaml`)

PR #198, opened 2018-11-14
Release 0.7.1

richardotis:
Fixes gh-197.

### DIFF
--- a/docs/api/pycalphad.rst
+++ b/docs/api/pycalphad.rst
@@ -6,23 +6,33 @@ Subpackages
 
 .. toctree::
 
+    pycalphad.codegen
     pycalphad.core
     pycalphad.io
     pycalphad.plot
+    pycalphad.tests
 
 Submodules
 ----------
 
-pycalphad\.model module
------------------------
+pycalphad.model module
+----------------------
 
 .. automodule:: pycalphad.model
     :members:
     :undoc-members:
     :show-inheritance:
 
-pycalphad\.variables module
----------------------------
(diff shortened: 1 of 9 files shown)


docs/faq.rst
```diff
@@ -57,11 +58,14 @@ grids are required for phase diagram calculations.
...
 
-This is an upstream bug in sympy, where floats are unable to be pickled.
-The fix has been completed, but not yet released. While the fix is not released,
-removing the gmpy2 package from their Python environment (e.g.
-``conda remove --force gmpy2``) will fix the error. Alternatively, setting the
-environment variable ``MPMATH_NOGMPY`` to a non-zero value will fix the error.
+This bug should now be fixed. Please update to pycalphad 0.7.1.
+
+
+```RecursionError``` during a calculation
```
bocklund: One set of ticks too many.

docs/api/pycalphad.rst
```diff
@@ -6,23 +6,33 @@ Subpackages
...
+----------------------
 
 .. automodule:: pycalphad.model
     :members:
     :undoc-members:
     :show-inheritance:
 
-pycalphad\.variables module
----------------------------
+pycalphad.refdata module
```
bocklund: We should probably remove this documentation and module. It's not a blocker, but either remove or open an issue to put it in the deprecation cycle so we don't forget.

richardotis: Defer to 0.8

PR #205, opened 2019-03-07
ENH: Model reference states

### LINKED ISSUE #55: Mixing properties are incorrect for phases with non-chemical energy contributions
The reason is, e.g., `GM_MIX`, is only subtracting out the contribution from `self.models['ref']`, which only includes the chemical energy contribution to each end-member. For phases with magnetic, Einstein, or some custom end-member contributions, the result will not be correct.

Standard properties like `GM`, `CPM`, etc., appear to be fine.

bocklund:
* Adds a `ReferenceState` class which allows the definition of a reference phase for each pure element in the system and any fixed state variables, such as temperature or pressure.
* `Model` now has a `shift_reference_state` method which can take these reference states and create new properties on the Model object to shift the AST for `XYZ` contributions `XYZR` (defaults to `GM`, `HM`, `SM`, `CPM`)
* `XYZ_MIX` properties have been fixed in the `Model` class to remove all endmember contributions (closes #55). For order-disorder models, this cannot be done properly in a way that sets the reference endmember energies to zero and preserves the disordered state mixing terms, so the `_MIX` energies have been set to `nan` in that case. 

* `test_energy.py` has been slightly refactored to be able to test arbitrary model properties.


bocklund:
The main ideas have all been implemented and could be reviewed.
There are several outstanding items that prevent this from being merged:


- [ ] An API hook in calculate/equilibrium to pass in ReferenceStates (see [1])
- [ ] Tests for the calculate/equilibrium API
- [ ] A documentation example showing usage of the different ReferenceState objects in calculate and equilibrium,
- [ ] Potential issue where if a user changes the AST of a model after instantiation (e.g. `mod.models['idmix'] = 0`) that the GMR will not update automatically. We need to set the GMR properties dynamically, e.g. `setattr(self, 'GMR', property(lambda self: self.GMR - ref.GMR))`, but that doesn't work because the `property` needs to be bound to the class. I'm pretty sure there's a way to do this, but I need to look into it.


[1] I want to ask for input for two things on this. 
1. My main blocker here is that I would like to be able to specify ReferenceStates either as a list of `[ReferenceState(...), ReferenceState(...)]`... that apply to all phases in the calculation or as a dict of `{'FCC_A1': [ReferenceState(...), ReferenceState(...), ...}`, basically following the format of unpack_kwargs. My hangup is that I'm not sure where to apply this, since the Models in calculate/equilibrium are just default dicts of either a `Model` subclass or a `Model` instance until they are pulled out and constituted in `build_callables`. I hesitate to propagate `Model` instances down to `build_callables` because that function shouldn't have to know anything about the internal representation of the phase or property it's building. 
2. My other question is whether the calculate/equilibrium should create copies of Model instances before applying the reference state, so that the user-passed Model objects are not modified, otherwise 

```python
# raises attribute error b/c GMR does not exist
equilibrium(dbf, comps, phases, conds, model=model_dict, output='GMR)
```

```python
equilibrium(dbf, comps, phases, conds, model=model_dict, refstate=reference_states, output='GMR)

# produces the same output as above because the objects in model_dict have been changed
equilibrium(dbf, comps, phases, conds, model=model_dict, output='GMR)
```

If users explicitly want this behavior, they should change the reference states manually:
```python
for phase_name in model_dict.keys():
    model_dict[phase_name].shift_reference_state(reference_states, dbf)
equilibrium(dbf, comps, phases, conds, model=model_dict, output='GMR)

# produces the same output as above
equilibrium(dbf, comps, phases, conds, model=model_dict, refstate=reference_states, output='GMR)
```

richardotis:
This API is great.

1. I think all phases in a given calculation should accept a common list of `ReferenceState`s. The meaning of the `R` properties is unclear in a mixed reference state environment. My thought is that `Model.reference_model` shouldn't affect `Model.GM` at all, so the solver should be independent of a reference state shift. When the user specifies `GMR` or whatever `R` property, we should leverage the property calculation machinery to compute the reference state shift after the fact and subtract it from the SER property value. I also think if the user specifies the `refstate` kwarg, we should assume they want `GMR` implicitly (and not return `GM` to avoid footguns).
2. The `refstate` kwarg should not mutate anything in the `models` kwarg.

Also, I know this isn't finished, but I notice that there is no test for `fixed_statevars` in `ReferenceState` yet.

bocklund:
Since Models are lazily instantiated via `unpack_kwarg`, it's tricky to know where they will be instantiated and we want to avoid mutating them. The latest they could possibly be instantiated is in`build_callables`, but I don't think the logic for copying `Model` instances and then calling `shift_reference_state` belongs in that function.

Can we defer calculate/equilibrium API to a later issue and merge this when the tests are passing?

richardotis:
Okay.

pycalphad/model.py
```diff
@@ -19,6 +19,51 @@
...
+        else:
+            self.species = v.Species(species)
+        self.phase_name = reference_phase
+        self.fixed_statevars = fixed_statevars if fixed_statevars is not None else {}
+
+    def __repr__(self):
+        if len(self.fixed_statevars.keys()) > 0:
+            s = "<ReferenceState('{}', '{}', {})>".format(self.species.name, self.phase_name, self.fixed_statevars)
+        else:
+            s = "<ReferenceState('{}', '{}')>".format(self.species.name, self.phase_name)
```
richardotis: same here

pycalphad/tests/test_energy.py
```diff
@@ -33,22 +40,35 @@ def calculate_energy(model, variables, mode='sympy'):
...
+
     mode, ['numpy', 'sympy'], optional
         Optimization method for the abstract syntax tree.
     """
-    # Generate a callable energy function
+    # Generate a callable function
     # Normally we would use model.subs(variables) here, but we want to ensure
     # our optimization functions are working.
-    energy = make_callable(model.ast, list(variables.keys()), mode=mode)
+    prop = make_callable(getattr(model, output), list(variables.keys()), mode=mode)
```
richardotis: `variables` is not guaranteed to have stable ordering for keys and values on all supported platforms (at least until py27 rolls off). It should probably be wrapped to `OrderedDict` at this time.

pycalphad/model.py
```diff
@@ -855,6 +949,80 @@ def atomic_ordering_energy(self, dbe):
...
+        Parameters
+        ----------
+        reference_states : Iterable of ReferenceState
+            Pure element ReferenceState objects. Must include all the pure
+            elements defined in the current model.
+        dbe : Database
+            Database containing the relevant parameters.
+        output : Iterable, optional
+            Parameters to subtract the ReferenceState from, defaults to ('GM', 'HM', 'SM', 'CPM').
+        contrib_mods : Mapping, optional
```
richardotis: Is this functionality (`contrib_mods`) tested anywhere?

pycalphad/model.py
```diff
@@ -19,6 +19,51 @@
...
+        if isinstance(species, v.Species):
+            self.species = species
+        else:
+            self.species = v.Species(species)
+        self.phase_name = reference_phase
+        self.fixed_statevars = fixed_statevars if fixed_statevars is not None else {}
+
+    def __repr__(self):
+        if len(self.fixed_statevars.keys()) > 0:
+            s = "<ReferenceState('{}', '{}', {})>".format(self.species.name, self.phase_name, self.fixed_statevars)
```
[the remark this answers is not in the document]
bocklund: Thanks. I'm really bad at these. I was actually looking for a PEP document specifying the standard recently. The `eval(repr(obj))` rule is nice and easy to remember.

pycalphad/tests/test_energy.py
bocklund: According to this [SO answer](https://stackoverflow.com/a/835430):

>> If `items()`, `keys()`, `values()`,  `iteritems()`, `iterkeys()`, and  `itervalues()` are called with no intervening modifications to the dictionary, the lists will directly correspond.

> On [2.x documentation](https://docs.python.org/2/library/stdtypes.html#dict.items) and [3.x documentation](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects)

So I think in this case, the lines

```python
    prop = make_callable(getattr(model, output), list(variables.keys()), mode=mode)
    # Unpack all the values in the dict and use them to call the function
    return prop(*(list(variables.values())))
```

Should preserve a consistent ordering, even though they are not guaranteed to be in insertion order, as OrderedDict gives. If this didn't work `zip(d.keys(), d.values())` would be unstable.

pycalphad/model.py
bocklund: It's currently tested in ESPEI 😉 I can write one for pycalphad's test suite.

bocklund:
An example notebook with three examples, including explaining the difference between `R` and `_MIX` properties. Added a simple test for `contrib_mods`. 

I think this is ready for another look.

richardotis:
Please merge when all the tests except Windows py37 (which is deferred for this issue) pass.

Issue #101, opened 2017-06-16
equilibrium: dask scheduler default incorrectly set to synchronous


richardotis:
The default setting of the `scheduler` kwarg in `equilibrium` is `dask.async.get_sync`, which means that the computation is being performed sequentially on a single core. This is suboptimal for any line/map calculation.

It looks like that change slipped in about 9 months ago in an unrelated commit, which is one of many reasons why we now do code review.

bocklund:
Is multiprocessing enough or should we set up a client by default?

richardotis:
I think `scheduler` should be `dask.multiprocessing.get` by default, but if the number of calculations is "small", we should just fall back to `dask.async.get_sync` (or whatever it's called now in newer versions).

richardotis:
We compute the number of calculations toward the bottom of `equilibrium()`

bocklund:
Some examples from NIMS databases. I assumed that parallelization in dask was independent of the shape of the equilibrium calculation. That is, for calculation shape of (P,T,X) that the scaling of (1,1,N) is similar to (1,sqrt(N),sqrt(N)). Each point is sampled twice and averaged.

It appears that multiprocessing would be a poor choice for default. 

What are your thoughts then on client vs sync, where with client we will have to implicitly start a distributed client. I think maybe sync as the default is okay given that the biggest speedup was less than 2x even for 10k equilibrium calculations on 8 cores.

![al-ni](https://user-images.githubusercontent.com/7681751/28586738-c3a52f52-7142-11e7-8275-f9111aa19998.png)
![cu-zr_sau](https://user-images.githubusercontent.com/7681751/28586735-c39d07d2-7142-11e7-8351-75b41cc2f3cb.png)
![la-ti](https://user-images.githubusercontent.com/7681751/28586736-c39e46e2-7142-11e7-9d87-6eb7b8386fa2.png)
![mg-zn](https://user-images.githubusercontent.com/7681751/28586737-c39eee76-7142-11e7-8dcb-594f0ef69ca4.png)

bocklund:
Log-log because why not 
![al-ni-log](https://user-images.githubusercontent.com/7681751/28586791-e3d00a4a-7142-11e7-95f0-b50ce5528d7c.png)
![cu-zr_sau-log](https://user-images.githubusercontent.com/7681751/28586790-e3ced224-7142-11e7-8af3-ca67bb821ee2.png)
[...]

PR #209, opened 2019-04-18
ENH: Binary phase diagram mapping

bocklund:
~This is a first look at an improved method of doing phase diagram mapping that I'm opening for some initial feedback.~

This improves on the efficiency of the brute force equilibrium calculations that are implemented now for binary phase diagrams.
 
### Main features

* No regressions. Visually, this should produce exactly the same results as the current brute force method. It's just faster.
* In theory (not implemented), every calculation temperature can be computed in parallel
* Some (trivial and optional) connecting of ZPF boundaries from the same two phase region is supported. Includes ability to close miscibility gaps (with a straight line) if the compositions are within a tolerance.
* Mappings in different regions (with different conditions) can be easily composed through the `ZPFBoundarySets` object.
* We can (in principle) test the plotting algorithm now (by comparing ZPFBoundarySets or matplotlib line collections). ~No tests are implemented here.~

~Currently, all the code is in the `pycalphad.mapping` module, however~ since there are no theoretical regressions, I think we can just drop in replace the mapping calculation and plotting for `binplot`.
~If we decide to move forward, I will reorganize the modules under `pycalphad.plot`. It just made sense to keep things separate while I was still developing the approach.~

#### Performance

There are two main sources for the performance bump compared to the brute force method, where every broadcasted condition is calculated:
1. We avoid almost all single phase calculations by finding two phase regions using a convex hull
2. Each two phase region only is calculated one time

Therefore, the most significant speed bumps are going to come from systems that have wide regions of single phase or two phases present. The slowest will have many small two phase regions. Assuming convex hull calculations are much slower than equilibrium calculations, this should never be slower than the brute force method.

~Since equilibrium calculations are looped, there is some small performance bump on the table to run `build_callables` and pass around the results (Al-Zn goes from 19s to 15s). ~

There are some performance improvements and optimizations in this branch, with help from @richardotis. 

1. `EquilibriumResult` instead of Datasets are implemented now. There are no slow xarray objects in the core, and EquilibriumResult wraps the Dataset constructor so we can just replace it at the end. A `to_xarray=True` argument is added to `equilibrium` in case users want to control whether or not an xarray dataset is created. 
2. Removes chunking code (it was part of the above, it's basically never used since #101.  
3. lower_convex_hull now users EquilibriumResult, removes `force_indep_align` (@richardotis anything we need to do for this?)
4. Overhead of calling equilibrium in a tight loop is bypassed by calling `starting_point` and `equilibrium` directly, using the grid that was used for the convex hull

#### Examples
All images from the new binplot_map. 

Timings do not include the JIT time and are dependent on hardware. Callables are not pre-built and passed for these.

##### Al-Zn
19s for binplot_map
4min 35s (275s) for binplot

Al-Zn-pts

##### Al-Fe
1m, 41s for binplot_map
Al-Fe-lines

###### Cu-Mg
32s for binplot_map
Cu-Mg-lines

##### Cr-Fe
1m, 14s for binplot_map
Cr-Fe

### Further discussion

The main idea of the algorithm is the following:

1. For each temperature:
2. Compute a convex hull, set X=0
3. Try to find a two phase region (miscibility gaps are detected by a site fraction tolerance between composition sets)
    I. If no two phase regions are found; T+=dT (go to 1)
    II. If a two phase region is found: set X=mean(composition sets)
4. Perform an equilibrium calculation
    I. If one phase is found: X+=dX; go to 3.
    II. If two phases are found: Save the composition sets; X+=dX; go to 4
 

Two assumptions are made in several places in this approach:
1. The diagrams are T-X
2. Binary systems only are assumed

The T-X assumption is fairly core to the code here. If we wanted to do isothermal calculations, e.g. isothermal ternaries, that would require a new mapping calculation function and significant changes to the ZPFBoundarySets object (maybe a new object would be more appropriate?).

We could get rid of the binary assumption and enable plotting of isopleths (including pseudo-binaries!). This would require
1. Supporting the ability to take in composition conditions as expressions with 1 composition degree of freedom.
2. The composition used everywhere in this code no longer matches up to a specific component, but more of a "reaction coordinate" of 1D compositions.
3. Math to project the equilibrium hyperplanes (and compositions) to the reaction coordinate.

Most of the plotting code (ZPFBoundarySets, TwoPhaseRegions) are relatively decoupled from the binary assumption, and it's just a matter of creating the CompSet2D objects in the mapping function to be in terms of the phases and compositions projected into the reaction coordinate.
If we eventually migrate to NP-conditions approach by following the ZPF lines, the code should look very similar and I think the changes would be fairly trivial to make.


richardotis:
Removing `_merge_property_slices` from equilibrium.py should help test coverage a bit.

bocklund:
What about just removing dask altogether? I don't think anyone has been able to use it successfully in at least 2 years, since that's when https://github.com/pycalphad/pycalphad/issues/101 was opened.

richardotis:
Okay. If it's a small change it can happen here; otherwise it should be its own PR.

bocklund:
Dask/distributed/dill/cloudpickle are removed

richardotis:
Can we improve test coverage for `ZPFBoundarySets` at all? It's okay for most of the actual plotting functions to remain uncovered, since we don't need to test that matplotlib works, but there are several important-looking methods in there that could easily regress without tests.

bocklund:
@richardotis Tests for TwoPhaseRegion and ZPFBoundarySets added and appear to be on their way to passing

Issue #207, opened 2019-03-13
Under-specifying conditions in `build_callables` leads to silent incorrect results


bocklund:
```python
>>> # build callables not done explicitly (correct behavior)
>>> calculate(dbf, comps, 'BCC_A2', model=model, output='HMR', T=1000, P=101325, points=np.array([[1-delta,delta],[delta,1-delta]])).HMR.values
array([[[2.12294678e-11, 3.17186277e-11]]])

>>> # explicit build_callables w/o specifing conditions, raises correctly
>>> cbs = build_callables(dbf, comps, [phase_name], conds=None, model=model, output='HMR', build_gradients=False)
ValueError: The following state variables must be specified: {T}

>>> # explicit build_callables with T specified, gives wrong answer silently 
>>> cbs = build_callables(dbf, comps, [phase_name], conds={v.T:1000000000}, model=model, output='HMR', build_gradients=False)
>>> calculate(dbf, comps, 'BCC_A2', model=model, output='HMR', T=1000, P=101325, points=np.array([[1-delta,delta],[delta,1-delta]]), callables=cbs).HMR.values
array([[[-2.52487526e+06, -1.24716281e+01]]])

>>> # explicit build_callables with T and P specified, works correctly
>>> cbs = build_callables(dbf, comps, [phase_name], conds={v.T:1000000000, v.P: 100000000}, model=model, output='HMR', build_gradients=False)
>>> calculate(dbf, comps, 'BCC_A2', model=model, output='HMR', T=1000, P=101325, points=np.array([[1-delta,delta],[delta,1-delta]]), callables=cbs).HMR.values
array([[[2.12294678e-11, 3.17186277e-11]]])

Issue #212, opened 2019-05-01
v.N not specified in conds for `build_callables`leads to incorrect energies


bocklund:

bocklund:
Example:


```python
from pycalphad import Database, equilibrium, binplot, variables as v
import pycalphad
print(pycalphad.__version__)
dbf = Database("/Users/brandon/Projects/pycalphad/examples/alzn_mey.tdb")
comps = ['AL', 'ZN', 'VA']
phases = ['LIQUID']
conds = {v.P: 101325, v.T: 800, v.X('AL'): 0.5}
# correct answer
print('correct answer')
print(equilibrium(dbf, comps, phases, conds))
from pycalphad.codegen.callables import build_callables
print('cbs')
cbs = build_callables(dbf, comps, phases, conds={v.T: 1000, v.P: 101325})
print(equilibrium(dbf, comps, phases, conds, callables=cbs))
print('cbs2')
cbs2 = build_callables(dbf, comps, phases, conds={v.T: 1000, v.P: 101325, v.X('AL'): 0.01})
print(equilibrium(dbf, comps, phases, conds, callables=cbs2))
print('cbs3')
cbs3 = build_callables(dbf, comps, phases, conds={v.T: 1000, v.P: 101325, v.N: 1})
print(equilibrium(dbf, comps, phases, conds, callables=cbs3))
```

## Output

#### Correct answer, no callables
correct answer
```
<xarray.Dataset>
Dimensions:    (N: 1, P: 1, T: 1, X_AL: 1, component: 2, internal_dof: 2, vertex: 3)
Coordinates:
  * N          (N) float64 1.0
  * P          (P) float64 1.013e+05
  * T          (T) float64 800.0
  * X_AL       (X_AL) float64 0.5
  * vertex     (vertex) int64 0 1 2
  * component  (component) <U2 'AL' 'ZN'
Dimensions without coordinates: internal_dof
Data variables:
    NP         (N, P, T, X_AL, vertex) float64 1.0 nan nan
    GM         (N, P, T, X_AL) float64 -3.807e+04
    MU         (N, P, T, X_AL, component) float64 -3.131e+04 -4.482e+04
    X          (N, P, T, X_AL, vertex, component) float64 0.5 0.5 ... nan nan
    Y          (N, P, T, X_AL, vertex, internal_dof) float64 0.5 0.5 ... nan nan
    Phase      (N, P, T, X_AL, vertex) <U6 'LIQUID' '' ''
Attributes:
[...]

Issue #208, opened 2019-04-08
Passing a raw Model instance to `build_callables` leads to silent incorrect results


richardotis:
Compare the following on `develop`:

```python
from pycalphad import Database, Model, calculate, variables as v

# Calculation Settings
dbf = Database('alfe_sei.TDB')
comps = ['AL', 'FE', 'VA']
phase_name = 'B2_BCC'
output = 'CPM'
endmember = [[0.5, 0.5, 0.5, 0.5, 1.0]]
temp_range = [300, 1000]

mod = Model(dbf, comps, phase_name)
contributions = mod.models.copy()
results = {}

for contribution, value in contributions.items():
    mod.models.clear()
    mod.models[contribution] = value
    result = calculate(dbf, comps, phase_name,
                       model=mod, output=output, points=endmember, T=(temp_range[0], temp_range[1], 5), P=1e5)
    if (getattr(result, output) == 0).all():
        continue
    results[contribution] = getattr(result, output)
```
The `build_callables` step will incorrectly generate fresh `Model` instances, leading to incorrect results.

For the intended behavior, one must explicitly pass a dict:
```python
from pycalphad import Database, Model, calculate, variables as v

# Calculation Settings
dbf = Database('alfe_sei.TDB')
comps = ['AL', 'FE', 'VA']
phase_name = 'B2_BCC'
output = 'CPM'
endmember = [[0.5, 0.5, 0.5, 0.5, 1.0]]
temp_range = [300, 1000]

mod = Model(dbf, comps, phase_name)
contributions = mod.models.copy()
results = {}

for contribution, value in contributions.items():
    mod.models.clear()
    mod.models[contribution] = value
    result = calculate(dbf, comps, phase_name,
                       model={phase_name: mod}, output=output, points=endmember, T=(temp_range[0], temp_range[1], 5), P=1e5)
    if (getattr(result, output) == 0).all():
        continue
    results[contribution] = getattr(result, output)
```

bocklund:
Here's a proposed solution for `build_callables`, which could help address the complex logic and edge cases that are in this, #212 and #207 

[...]

PR #214, opened 2019-05-15
MAINT: Refactor build_callables

bocklund:
The following changes are made:

1. `build_callables` now becomes `build_callables` and `build_phase_records`. 
    - Both require that correctly instantated model dicts, i.e. {'PHASE_NAME': Model()}
    - `build_callables` should be called by users and just builds the callable functions and JIT compiles them. It returns a dict of `{'output': {'callables': {'PHASE_NAME': AutowrapFunc()}, 'massfuncs': {'PHASE_NAME': AutowrapFunc()}, ...}}`. These callables are passed to equilibrium and calculate.
    - Calculate and equilibrium should be calling `build_phase_records`, which optionally take `callables`. If callables are provided for the output, they are not recomputed (if users pass callables, they are responsible for determine if they are compatible). If no callables are passed, compatible ones will be built automatically.
    - `equilibrium` should not be building callables or phase records for other outputs. It just passes along the user callables (if they are provided) and `calculate` handles building the phase records for the particular outputs in `_eqcalculate`.
   - It is the responsibility of the caller of `build_callables` to know whether gradients and Hessians are required. This means two cases can exist. Either the user calls `build_callables` (the user is responsible) or `build_callables` is called by `build_phase_records` (`calculate`/`equilibrium` are responsible for propagating the correct settings). For `calculate`, no gradients or Hessians are built. For `equilibrium`, gradients are always built, and Hessians are built if any of the conditions are instances of the classes in `pycalphad.variables.CONDITIONS_REQUIRING_HESSIANS`.
2. This also refactors all the code where state variables and models were retrieved from the result of the old `build_callables`. Getting these additional outputs violates single the responsibility principle. A convenience function for getting state variables from models and or conditions now exists for `calculate`, `equilibrium`, and `build_callables`/`build_phase_records` to use.
3. The following code in the test `test_eq_build_callables_with_parameters` was removed, because this is not valid - parameters are not added automatically through build callables because `PhaseRecords` are no longer built in `build_callables`.
```python
    # Check that passing callables should skip the build phase, but use the values from 'VV0000' saved in callables
    eq_res = equilibrium(dbf, comps, phases, conds, callables=callables)
    np.testing.assert_allclose(eq_res.GM.values.squeeze(), 20000.0)
```
4. v.N now gets specified automatically in calculate. Previously, it was automatically specified in equilibrium, but not calculate. When `calculate` was called from `equilibrium` that there was an extra dimension for `v.N` that is not present when `calculate` alone was called. A user-friendly alternative to the ConditionError error message is written if `v.N` is not in the equilibrium or calculate conditions. This would cause the `callables` to give incorrect answers in either `calculate` or `equilibrium` if called separately. All callables should be built with the `v.N` statevariable (while v.N=1 is a required condition internally) so that the behavior is constistent with both. Users building callables by hand should be aware that they may need to add `v.N` to the callables by the `additional_statevars` argument of `build_callables`. `test_issue116` was updated to reflect that N is added automatically. As much as possible, we should try to progressively expose pycalphad variables objects to the end user - IMO we state variables should be part of a state variables dict to calculate.

### DIFF
--- a/pycalphad/core/utils.py
+++ b/pycalphad/core/utils.py
@@ -398,10 +398,16 @@ def instantiate_models(dbf, comps, phases, model=None, parameters=None, symbols_
     -------
 
     """
-    from pycalphad import Model
+    from pycalphad import Model  # avoid cyclic imports
     parameters = parameters if parameters is not None else {}
     if symbols_only:
         parameters, _ = extract_parameters(parameters)
+    if isinstance(model, Model):  # Check that this instance is compatible with phases
+        if len(phases) > 1:
+            raise ValueError("Cannot instantiate models for multiple phases ({}) using a Model instance ({}, phase: {})".format(phases, model, model.phase_name))
+        else:
+            if phases[0] != model.phase_name:
+                raise ValueError("Cannot instantiate models because the desired {} phase does not match the Model instance () {} phase.".format(phases[0], model.phase_name, model))
     models_dict = unpack_kwarg(model, Model)
     for name in phases:
         mod = models_dict[name]
--- a/pycalphad/core/utils.py
+++ b/pycalphad/core/utils.py
@@ -3,10 +3,12 @@ The utils module handles helper routines for equilibrium calculation.
 """
 from __future__ import division
 import pycalphad.variables as v
+from pycalphad import Model
 from pycalphad.core.halton import halton
 from pycalphad.core.constants import MIN_SITE_FRACTION
 from sympy.utilities.lambdify import lambdify
 from sympy.printing.lambdarepr import LambdaPrinter
+from sympy import Symbol
 import numpy as np
 import operator
 import functools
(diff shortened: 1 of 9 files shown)


bocklund:
I think the main ideas here are in place following the discussion from #208. I am going to take another look at #208, #212, and #207 to see if there's a test that needs to be added.

richardotis:
Broad, forward-thinking question: In a world where we want to read phase records from binary files on disk ("compiled databases"), does this new design make it easy to add support for that?

bocklund:
That's the main idea, yeah. 

Currently, PhaseRecord objects require conditions that are specific to the equilibrium problem being solved, so they should be built and compiled JIT. On the other hand, the callables compiled by `build_callables` can be used in any problem that retains the same state variables, so it makes sense to split them up.

In the future, we could also break out the ability to build constraints into it's own function, but that's relatively easy to extract with this design.

bocklund:
Regarding the code in #208,

This 
```python
from pycalphad import Database, Model, calculate, variables as v
import numpy as np

# Calculation Settings
dbf = Database('tdbs/alfe_sei.TDB')
comps = ['AL', 'FE', 'VA']
phase_name = 'B2_BCC'
output = 'CPM'
endmember = [[0.5, 0.5, 0.5, 0.5, 1.0]]
temp_range = [300, 1000]

mod = Model(dbf, comps, phase_name)
contributions = mod.models.copy()
results = {}

for contribution, value in contributions.items():
    mod.models.clear()
    mod.models[contribution] = value
    result = calculate(dbf, comps, phase_name,
                       model=mod, output=output, points=endmember, T=(temp_range[0], temp_range[1], 5), P=1e5)
    if (getattr(result, output) == 0).all():
        continue
    results[contribution] = getattr(result, output)
# plot
cumulative = np.zeros_like(results['ref'].squeeze())
for contrib, arr in results.items():
    a = arr.squeeze()
    plt.fill_between(np.arange(temp_range[0], temp_range[1], 5), cumulative, a+cumulative, label=contrib)
    cumulative += a
plt.legend()
```

Produces this:

cpm-contribs

because the Model instance gets sent to `instantiate_models`, which calls `unpack_kwarg(Model_instance, Model)`, creating a dict `{'B2_BCC': Model_instance}`.

However, this can cause an issue if there are multiple phases and a Model instance, since the default_dict created will return the instances. So `instantiate_models` now raises a friendly error if `len(phases)` > 1 and `isinstance(model, Model)`.

bocklund:
#207  and #212 both have to do with the fact that callables did not have the right set of state variables.

There was a subtle bug, related to but not causing those issues, where calculate did not automatically add the `N` state variable, but equilibrium did implicitly add the `N` state variable, which was passed to calculate. This means that a set of callables that works with `equilibrium` did not work with `calculate` unless the `N` state variable was added to calculate calls (which `equilibrium ` did under the hood, so `equilibrium` was internally consistent). This PR fixes that bug by always making sure `N` is in the state variables of `calculate` implicitly, like in `equilibrium`.

There's also the direct issue of users creating incompatible callables. Since `N` is implicitly added to  every calculation, `T` is in about every Model model and pressure is the other condition specified, we give a warning if a users `build_callables` does not build callables with the `N, P, T` state variables and we instruct users to add the `N, P, T` state variables through the `additional_statevars` argument of `build_callables`. We don't raise in case there's a legitimate region to not use `N, P, T`.

Tests were added for #208, but I don't think these issues warrant any tests, so this should close #207, #208 and #212

pycalphad/core/utils.py
```diff
@@ -347,3 +348,112 @@ def all_sublattices_active(comps, phase):
...
+    True
+    """
+    state_vars = set()
+    if models is not None:
+        for mod in models.values():
+            state_vars.update(mod.state_variables)
+    if conds is not None:
+        for c in conds.keys():
+            # StateVariable instances are ok (e.g. P, T, N, V, S),
+            # however, subclasses (X, Y, MU, NP) are not ok.
```
richardotis: This turns out to be a pretty good heuristic, though I'll note that the actual criteria for `state_vars` is whether or not the variable is an independent degree of freedom in the model. Technically, they are all "state variables" (hence they subclass `StateVariable`), but in pycalphad internal lingo a "state variable" is an independent degree of freedom which needs to be pre-specified for global minimization purposes.

bocklund: I agree. For now, this is kind of a hack that takes advantage of the current type hierarchy. If this heuristic gets stale, then I think we should consider making changes to the type hierarchy so we provide a more explicit and consistent interface for defining and using StateVariables in Models and in the minimizer.

Issue #201, opened 2019-02-01
CI improvements


richardotis:
A few non-blocking CI nice-to-haves:
- Remove obsolete Docker-related commands in `.travis.yml`
- Switch to a `conda-build`-based approach for installing pycalphad on CI. This will put the conda recipe through CI and create a single source of truth for dependencies.
- Refactor the tests to limit the total JIT time. Prefer recycled callables objects as much as possible.

bocklund:
Also switching to pytest would be nice. Nose is not maintained and we don't use any advanced Nose features that should prevent us from switching

bocklund:
Closed through #225 and #216 

Note that conda-build was determined not useful at this time because it significantly hurts the performance of the test suite (10 mins -> 20 mins according to #225).

With symengine support, CI run time is ~10 mins instead of ~40-50 mins on Travis, which precludes the need to refactor tests around minimizing JIT time for now.

PR #216, opened 2019-05-17
TST/MAINT: Remove Docker code

bocklund:
Docker is no longer part of the testing toolchain. It is removed here as part of #201.


richardotis:
You need to add `nose` and `coverage` as packages in the build environment. Some of the errors you're getting are due to picking up the system `nose` instead of the one that should be in the conda environment.

For the Windows errors, I'm not sure, but my guess is a `conda` / `conda-build` version incompatibility.

bocklund:
For now, I think I'll just clean up the docker code paths. The WIP updated recipe for conda-build and conda-build on Travis is now at this branch: https://github.com/pycalphad/pycalphad/tree/ci-conda-build

PR #217, opened 2019-05-22
FIX: Improve Model instantiation

bocklund:
Two changes here:

1. `instantiate_models` now returns a dict instead of a default dict to be more explicit. Only phases passed to `instantiate_models` are included.
2. Fixed a subtle bug in calculate where over-specified models could have led to inconsistent maximum internal degree of freedom between `calculate` and `starting_point` because `calculate` based the internal dof size on all the model instances, while starting_point used the phase records (active phases only).

### DIFF
--- a/pycalphad/tests/test_utils.py
+++ b/pycalphad/tests/test_utils.py
@@ -35,7 +35,7 @@ def test_instantiate_models_only_returns_desired_phases():
     assert len(inst_mods) == len(phases)
 
     # models are underspecified w.r.t. phases
-    too_few_phases = ['FCC_A1', 'LIQUID', 'AL3NI1']
-    too_few_models = {phase: Model(ALNIPT_DBF, comps, phase) for phase in too_many_phases}
+    too_few_phases = ['FCC_A1']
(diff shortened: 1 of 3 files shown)


pycalphad/tests/test_utils.py
```diff
@@ -15,9 +15,27 @@ def test_filter_phases_removes_disordered_phases_from_order_disorder():
...
+    phases = ['FCC_A1', 'LIQUID']
+
+    # models are overspecified w.r.t. phases
+    too_many_phases = ['FCC_A1', 'LIQUID', 'AL3NI1']
+    too_many_models = {phase: Model(ALNIPT_DBF, comps, phase) for phase in too_many_phases}
+    inst_mods = instantiate_models(ALNIPT_DBF, comps, phases, model=too_many_models)
+    assert len(inst_mods) == len(phases)
+
+    # models are underspecified w.r.t. phases
+    too_few_phases = ['FCC_A1', 'LIQUID', 'AL3NI1']
```
richardotis: Is this test correct? It looks identical to the one above it.

bocklund: Good catch, should be fixed

Issue #219, opened 2019-05-31
Lower convex hull is missing phases


bocklund:
Lower convex hull is not picking certain phases on the current develop (https://github.com/pycalphad/pycalphad/commit/5959be6439f4a0d2e0522c41f01986f761340da5)

I have narrowed this down to an example for Al-Zn near the Al-rich melting point. Roughly the center of the equilibria to calculate is at X(ZN)=0.0609, but we can verify that the starting point can never find the fcc phase, while equilibrium finds it.

The following example can reproduce this (I have confirmed that this behavior is consistent on 0.7.1.post2), even with a high point density, the fcc phase is not picked up.

```python
from pycalphad import Database, equilibrium, binplot, variables as v, calculate
TDB_PATH = '/Users/brandon/Projects/pycalphad/examples/alzn_mey.tdb'
dbf = Database(TDB_PATH)
comps = ['AL', 'ZN', 'VA']
phases = list(dbf.phases.keys())
# True composition
# X_ZN = 0.0609
X_ZN = (0, 0.0609, 0.005)
conds = {v.P: 101325, v.T: 900.0, v.N: 1.0, v.X('ZN'): X_ZN}
eq = equilibrium(dbf, comps, phases, conds, calc_opts={'pdens': 200000})

print('\n\n\n equilibrium result', eq)
print('eq phases', eq.Phase.values)
```

I added the following lines to the end of `starting_point()` before the return to help debug this:

```python
    print('starting point result', result)
    print('starting point result phases', result.Phase.values)
```

With these additions, I get the following output, where the top dataset is from the starting point and the bottom from equilibrium.

```
starting point result <xarray.Dataset>
Dimensions:    (N: 1, P: 1, T: 1, X_ZN: 13, component: 2, internal_dof: 2, vertex: 3)
Coordinates:
  * N          (N) float64 1.0
  * P          (P) float64 1.013e+05
  * T          (T) float64 900.0
  * X_ZN       (X_ZN) float64 1e-12 0.005 0.01 0.015 ... 0.045 0.05 0.055 0.06
  * vertex     (vertex) int64 0 1 2
  * component  (component) <U2 'AL' 'ZN'
Dimensions without coordinates: internal_dof
Data variables:
    NP         (N, P, T, X_ZN, vertex) float64 3.914 -2.914 0.0 ... -2.52 0.0
    GM         (N, P, T, X_ZN) float64 -3.398e+04 -3.411e+04 ... -3.552e+04
[...]

PR #220, opened 2019-05-31
ENH: Use symengine

richardotis:
- Switch to SymEngine/LLVM-based codegen
- Remove SymPy codegen
- Drop Windows py27 support
- hyperplane: Fixes gh-219. A heuristic for forcing faster convergence would sometimes fail by causing feasible starting points close to the edge of composition space to appear infeasible. The heuristic is now removed.
- lower_convex_hull: Cleanup calls to `global_grid` which more correctly go to `grid`, to reduce array duplication


richardotis:
Done.

PR #222, opened 2019-06-08
ENH: Charge constraints

bocklund:
Opening this PR for comments.

These changes
* Add a constraint that a phase with the model hint `charged_phase` (type code `I`) is charge balanced (with a test)
* Add a check in the `Model` initializer that warns the user if a phase has charged species active, but is not using a Model that corresponds to charged phases (charged model, ionic liquid, gas, or aqueous)


richardotis:
I would try multiplying the return value of the calculated charge constraint by 100, and putting the constraint violation tolerance back to the previous value.

bocklund:
Due to the flakiness of the conditions, I'm closing this awaiting action on Hessian support (either merged or not supported).

PR #224, opened 2019-06-09
MAINT/TST: Put build_reference_model behind a property

bocklund:
Profiling showed that a significant amount of time is spent building reference models. This code puts the code to build_reference_models behind a property, which speeds up the test suite by about 30% based on my tests.

### DIFF
--- a/pycalphad/model.py
+++ b/pycalphad/model.py
@@ -302,7 +302,40 @@ class Model(object):
 
     @property
     def reference_model(self):
-        """Reference model built as needed improve performance."""
+        """
+        Return a Model containing only energy contributions from endmembers.
+
+        Returns
+        -------
+        Model
+
+        Notes
+        -----
+        The reference_model is defined such that subtracting it from the model
+        will set the energy of the endmembers for the _MIX properties of this
+        class to zero. The _MIX properties generated here allow users to see
+        mixing energies on the internal degrees of freedom of this phase.
+
+        The reference_model AST can be modified in the same way as the current Model.
+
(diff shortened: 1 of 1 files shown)


pycalphad/model.py
```diff
@@ -301,6 +300,13 @@ def degree_of_ordering(self):
...
     mixing_heat_capacity = CPM_MIX = property(lambda self: -v.T*self.GM_MIX.diff(v.T, v.T))
 
+    @property
+    def reference_model(self):
+        """Reference model built as needed improve performance."""
+        if self._reference_model is None:
+            self.build_reference_model(self._dbe)
+        return self._reference_model
+
     def build_reference_model(self, dbe, preserve_ideal=True):
```
richardotis: `build_reference_model` is mixing user-provided state (`dbe`) with instance state. This is no longer necessary since `Model` keeps a private `_dbe` attribute now. Can the API be changed? Should `build_reference_model` be a private function?

pycalphad/model.py
```diff
@@ -301,6 +300,13 @@ def degree_of_ordering(self):
     mixing_entropy = SM_MIX = property(lambda self: -self.GM_MIX.diff(v.T))
     mixing_heat_capacity = CPM_MIX = property(lambda self: -v.T*self.GM_MIX.diff(v.T, v.T))
 
+    @property
+    def reference_model(self):
+        """Reference model built as needed improve performance."""
```
richardotis: Awkward docstring wording

PR #225, opened 2019-06-10
TST: Switch to pytest, simplify CI

bocklund:
Tries to address #201. 
1. Switched to pytest on Travis and Appveyor
2. Simplified some of the Travis macOS conda-metachannel because it wasn't providing that much of an improvement (had to remove the `gxx_linux-64` dependency, moved all Linux builds over to Xenial over Trusty for the build toolchain).

My first pass at conda-build seemed to not work out. I think we could make it work, but it seems like it's taking longer to build - see the comparison for just before  and after adding conda build.

* Before: https://github.com/pycalphad/pycalphad/commit/907d9fe0453b2faa945e44810c69393c6ffbe88b, [Travis tests](https://travis-ci.org/pycalphad/pycalphad/builds/543640027?utm_source=github_status&utm_medium=notification) - 10 minutes to test successfully with coverage
* After: https://github.com/pycalphad/pycalphad/commit/dc12b2e08bf47ea13958f9dbf1f02d32ed549d10, [Travis tests](https://travis-ci.org/pycalphad/pycalphad/builds/543649641?utm_source=github_status&utm_medium=notification) - 20 minutes to fail after finishing `conda-build` without running tests

If we're just trying to get the dependencies so they aren't specified in three places, it might be better to try a different route where we can pull them in from a requirements or something. What do you think, @richardotis?


richardotis:
There are a few options, neither of which really solve the CI issue but could reduce duplication between meta.yaml and setup.py:
1. `conda skeleton pypi pycalphad` - generate recipes after uploading to PyPI; probably still requires some manual tweaking
2. `python setup.py bdist_conda` - keep the recipe metadata in setup.py

richardotis:
After discussion offline, I agree to abandon the present push for conda-build. Please merge the working CI changes when ready.

PR #236, opened 2019-09-01
ENH: User-specified starting points

richardotis:
- Give `equilibrium` a `user_starting_point` and `global_min` keyword argument
- `user_starting_point` works by specifying a list of tuples with internal degrees of freedom, like `[('BCC_A2', [1, 1e5, 600, 0.4, 0.6, 1.0]), ('LIQUID', [1, 1e5, 600, 0.3, 0.7])]`
- `global_min` is a boolean, default `True`
- `binplot` has been modified to take advantage of this feature
- Right now the two options can be used independently, though if `global_min=True` in principle the user-specified starting point shouldn't affect the final result
- This is mainly to enable things like T0 and parallel tangent calculation, where certain desired hyperplanes are metastable (e.g., inside a miscibility gap)
- The performance boost is significant for cases where the provided starting point is close to the solution

There is very minimal checking of the user input for this.


pycalphad/core/starting_point.py
```diff
@@ -77,23 +51,39 @@ def starting_point(conditions, state_variables, phase_records, grid):
...
+            phase_dof_without_statevars = phase_dof[len(state_variables):]
+            result['NP'][..., phase_idx] = 1./len(given_starting_point)
+            phase_records[phase_name].obj(out_energy[phase_idx], np.atleast_2d(phase_dof))
+            for comp_idx in range(len(nonvacant_elements)):
+                phase_records[phase_name].mass_obj(out_moles[phase_idx, :, comp_idx], np.atleast_2d(phase_dof), comp_idx)
+            result['Phase'][..., phase_idx] = phase_name
+            result['X'][..., phase_idx, :] = out_moles[phase_idx, 0]
+            result['Y'][..., phase_idx, :len(phase_dof_without_statevars)] = phase_dof_without_statevars
+            result['Y'][..., phase_idx, len(phase_dof_without_statevars):] = np.nan
+            out_energy[:] = 0
```
richardotis: If `len(given_starting_point) < max_phases` we should set the unused parts of the empty array to `np.nan`

richardotis:
This needs another test that explicitly exercises the lack of global min, e.g., a calculation inside a miscibility gap where we verify that the energy and chemical potentials are correct for inside that region.

pycalphad/core/eqsolver.pyx
```diff
@@ -181,7 +181,7 @@ cdef _solve_and_update_if_converged(composition_sets, comps, cur_conds, problem,
             phase_idx += 1
     return result
 
-def _solve_eq_at_conditions(comps, properties, phase_records, grid, conds_keys, state_variables, verbose,
+def _solve_eq_at_conditions(comps, properties, phase_records, grid, conds_keys, state_variables, global_min, verbose,
```
bocklund: I'm thinking that this function is starting to get complicated enough that breaking it up makes sense. There's several things this is responsible for:

1. Iterating through one condition in a broadcasted set at a time
2. Either
  a. Ignoring convergence (not solving)
  b. Solving without performing any global min (no checking for adding/removing phases)
  c. Solving with global min
3. Handling failed convergence (in cases 2b and 2c)
4. Propagating results from a single set of conditions back to the properties (closing the loop opened in 1)

I propose that we slightly refactor `_solve_eq_at_conditions` by extracting out the items in 2a-c to their own functions, which only know about one set of conditions at a time. Item 3 may also make sense to refactor, but it's fine if it's left in.

This can also be used to fix a possible spurious solve at here: https://github.com/pycalphad/pycalphad/blob/651af0735d4e3a3571564a05adacb7b4ea307257/pycalphad/core/eqsolver.pyx#L317

pycalphad/core/equilibrium.py
```diff
@@ -239,22 +247,25 @@ def equilibrium(dbf, comps, phases, conditions, output=None, model=None,
...
+        grid = calculate(dbf, comps, active_phases, model=models, fake_points=True,
+                         callables=callables, output='GM', parameters=parameters,
+                         to_xarray=False, **grid_opts)
+    else:
+        grid = None
     coord_dict = str_conds.copy()
     coord_dict['vertex'] = np.arange(len(pure_elements) + 1)  # +1 is to accommodate the degenerate degree of freedom at the invariant reactions
     coord_dict['component'] = pure_elements
-    properties = starting_point(conds, state_variables, phase_records, grid)
+    properties = starting_point(conds, state_variables, phase_records, grid, given_starting_point=user_starting_point)
```
bocklund: IMO `starting_point` should be two different functions that are called separately here to emphasize to the reader that a `grid` is not needed for a `user_starting_point`. E.g. signatures:

`starting_point(conditions, state_variables, phase_records, grid)`
`custom_starting_point(conditions, state_variables, phase_records, compsets)`, where `compsets` is the `(phase, dof)` tuple.

pycalphad/core/equilibrium.py
```diff
@@ -239,22 +247,25 @@ def equilibrium(dbf, comps, phases, conditions, output=None, model=None,
...
-    # 'calculate' accepts conditions through its keyword arguments
-    grid_opts = calc_opts.copy()
-    statevar_strings = [str(x) for x in state_variables]
-    grid_opts.update({key: value for key, value in str_conds.items() if key in statevar_strings})
-    if 'pdens' not in grid_opts:
-        grid_opts['pdens'] = 500
-    grid = calculate(dbf, comps, active_phases, model=models, fake_points=True,
-                     callables=callables, output='GM', parameters=parameters, 
-                     to_xarray=False, **grid_opts)
+    if global_min:
```
bocklund: I think this will fail if `(not global_min) and (not user_starting_point)` is True, since `starting_point()` will have no grid.

pycalphad/core/equilibrium.py
```diff
@@ -174,6 +175,11 @@ def equilibrium(dbf, comps, phases, conditions, output=None, model=None,
         Defaults to a pycalphad.core.solver.InteriorPointSolver.
     callables : dict, optional
         Pre-computed callable functions for equilibrium calculation.
+    global_min : bool, optional
+        Automatic miscibility gap detection ('global minimization').
+    user_starting_point : list of (phase name, dof) tuple
```
bocklund: Does having `user_start_point` as an argument to the high level `equilibrium` function make sense? It's probably meaningless to provide a starting point in this way for broadcasted conditions that are not close to each other. 

I've been chewing on this for awhile. My wish is to have an API written around calculating point equilibria that can efficiently be broadcasted to any set of conditions. Not sure how to accomplish this without a major rewrite, so I won't ask for that here.

bocklund: It might be possible, after refactoring `_solve_eq_at_conditions` per my previous comment, to only loop through conditions and call a "_solve_eq_at_condition" (no "s") function. 
Then you're just refactoring equilibrium to either take a broadcasted (`_solve_eq_at_conditions`) or unbroadcasted (`_solve_eq_at_condition`) code path.

bocklund:
@richardotis can you update the status of this PR? Do you still want to address adding user start points this way, or is it something you want to rework before/after the new solver is finished?

richardotis:
I'm still working through starting point concepts, and whether or not to move forward with this approach will probably become clearer once I have fixed-phase conditions working.

richardotis:
Now that the 0.9 series minimizer is nearing readiness, we have more flexibility with passing infeasible starting points to the solver. This makes it easier to consider metastable configurations without needing to specify exact site fractions for all phases. I've wrapped the new `pointsolve` function in some naive starting-point selection logic to create a minimum-viable `local_equilibrium` function:

```python
from pycalphad.core.eqsolver import pointsolve
from pycalphad.core.solver import SundmanSolver
from pycalphad.core.composition_set import CompositionSet
from pycalphad.core.utils import unpack_components, instantiate_models
from pycalphad.codegen.callables import build_phase_records
from pycalphad import calculate
import numpy as np

def local_equilibrium(dbf, comps, phases, conds, composition_sets=None):
    # Broadcasting conditions not supported
    cur_conds = {str(k): float(v) for k, v in conds.items()}
    species = sorted(unpack_components(dbf, comps), key=str)
    state_variables = np.array([cur_conds['N'], cur_conds['P'], cur_conds['T']], dtype=np.float64)
    # Note: filter_phases() not called, no all specified phases must be valid
    models = instantiate_models(dbf, species, phases)
    phase_records = build_phase_records(dbf, species, phases, conds, models,
                                        build_gradients=True, build_hessians=True)
    composition_sets = []

    # Choose a naive starting point for each phase
    # only one composition set per phase is chosen
    # here, we just choose the point with the minimum Gibbs energy
    # mass balance is not necessarily preserved
    for phase in phases:
        # arbitrary guess
        phase_amt = 1./len(phases)
        # callables are being unnecessarily rebuilt here
        calc_p = calculate(dbf, comps, phase, T=cur_conds['T'], P=cur_conds['P'],
                           pdens=10, model=models, callables=None)
        idx_p = np.argmin(calc_p.GM.values.squeeze())
        compset = CompositionSet(phase_records[phase])
        site_fractions = np.array(calc_p.Y.isel(points=idx_p).values.squeeze())
        compset.update(site_fractions, phase_amt, state_variables)
        composition_sets.append(compset)

    # Calculate a local equilibrium for the specified phases
    solver = SundmanSolver()
    result = pointsolve(composition_sets, species, cur_conds, solver)
    return result, composition_sets
```

We can see that miscibility gaps will not be computed by this function. Usage example:
```python

dbf = Database('alzn_mey.tdb')
result, composition_sets = local_equilibrium(dbf, ['AL', 'ZN', 'VA'],
                                             ['LIQUID', 'FCC_A1', 'HCP_A3'],
                                             {v.N: 1, v.P: 1e5, v.T: 600, v.X('ZN'): 0.3})
print('Converged: ', result.converged)
print('Chemical Potentials: ', result.chemical_potentials)
print(composition_sets)
```
```
Converged:  True
Chemical Potentials:  [-20603.11611491 -28529.45374882]
[CompositionSet(FCC_A1, [0.7 0.3], NP=1.0, GM=-22981.017405084138)]
```
Note that the fcc miscibility gap is the stable configuration, but we are successfully able to compute within the miscibility gap. Also note that, despite specification of the liquid and hcp phases, they are correctly optimized and removed during minimization. Local equilibrium does not necessarily mean fixing the amount of any phase.

Based on these preliminary results, my thinking is that it may make more sense to define a separate `local_equilibrium` function with the semantics we are after, rather than pushing for starting point hooks. I do want to be sensitive to avoiding too much duplication of code between equilibrium functions, but the MVP function I've created here is simple enough that it makes me think that just creating it, and then letting users copy and customize it to deal with edge cases like allowing some phases to have miscibility gaps. Of course it can still be improved and more keyword argument support can be added to bring it closer to parity with `equilibrium`. Thoughts?

bocklund:
I think this script and function are very reasonable for a user to use today. 

I'm not convinced yet that `local_equilibrium` is something that pycalphad needs to provide as part of the API, rather than being a part of the cookbook as one or more examples. 

However, I also agree that it doesn't make sense to add hooks for custom starting points in to `equilibrium` without a more significant refactoring. To me, the point of `equilibrium` is to be bulletproof and prevent mistakes. Custom starting points adds a lot of edge and corner cases to consider that would make it a lot less bulletproof. 

When someone needs to compute a local equilibrium, I can imagine as many cases where you _want_ control over the starting point for one reason or another. Maybe to control a certain set of phases or to choose a starting point in a miscibility gap that's closer to satisfying the overall composition condition (for convergence).

The starting point selection and composition set construction is the least-ergonomic part of this process (the rest of the `local_equilibrium` function is ~10 lines of code). Making it easier to create weird starting points and composition set objects would go a long way to making it easier for a user to derive something like this themselves that is more appropriate for their need than what we could provide in something like this `local_equilibrium` function.

richardotis:
Responding to your point about the ease of composition set creation, what do you think about keeping `local_equilibrium` as a cookbook function, but perhaps merging in something like the following `local_composition_sets` function?

```python
from pycalphad.core.eqsolver import pointsolve
from pycalphad.core.solver import SundmanSolver
from pycalphad.core.composition_set import CompositionSet
from pycalphad.core.utils import unpack_components, instantiate_models
from pycalphad.codegen.callables import build_phase_records
from pycalphad import calculate
import numpy as np

def local_composition_sets(dbf, comps, phases, conds):
    # Broadcasting conditions not supported
    cur_conds = {str(k): float(v) for k, v in conds.items()}
    species = sorted(unpack_components(dbf, comps), key=str)
    state_variables = np.array([cur_conds['N'], cur_conds['P'], cur_conds['T']], dtype=np.float64)
    # Note: filter_phases() not called, no all specified phases must be valid
    models = instantiate_models(dbf, species, phases)
    phase_records = build_phase_records(dbf, species, phases, conds, models, build_gradients=True, build_hessians=True)
    composition_sets = []

    # Choose a naive starting point for each phase
    # only one composition set per phase is chosen
    # here, we just choose the point with the minimum Gibbs energy
    # mass balance is not necessarily preserved
    for phase in phases:
        # arbitrary guess
        phase_amt = 1.0
        # callables are being unnecessarily rebuilt here
        calc_p = calculate(dbf, comps, phase, T=cur_conds['T'], P=cur_conds['P'], pdens=10, model=models, callables=None)
        idx_p = np.argmin(calc_p.GM.values.squeeze())
        compset = CompositionSet(phase_records[phase])
        site_fractions = np.array(calc_p.Y.isel(points=idx_p).values.squeeze())
        compset.update(site_fractions, phase_amt, state_variables)
        composition_sets.append(compset)
    return composition_sets

def local_equilibrium(dbf, comps, phases, conds, composition_sets=None):
    # Broadcasting conditions not supported
    cur_conds = {str(k): float(v) for k, v in conds.items()}
    species = sorted(unpack_components(dbf, comps), key=str)
    composition_sets = composition_sets or local_composition_sets(dbf, comps, phases, conds)
    # Calculate a local equilibrium for the specified phases
    solver = SundmanSolver()
    result = pointsolve(composition_sets, species, cur_conds, solver)
    return result, composition_sets
```

richardotis:
Superseded by #415

Issue #191, opened 2018-10-25
Automatic detection of pure-vacancy phases


richardotis:
**If you are interested in contributing to pycalphad, this bug would be a great place to start. Feel free to ask questions.**

There is an algorithm in `calculate` and `equilibrium` that automatically suspends phases (with a warning) if including the phase would leave any of its sublattices without any components. This makes it convenient to pass in all of the phases in a given database and have pycalphad include only the phases relevant to the selected components.

A bug currently occurs for phases with a TDB file like:
```python
from pycalphad import Database, calculate
TDB = """
 ELEMENT /-   ELECTRON_GAS              0.0000E+00  0.0000E+00  0.0000E+00!
 ELEMENT VA   VACUUM                    0.0000E+00  0.0000E+00  0.0000E+00!
 ELEMENT AL   FCC_A1                    2.6982E+01  4.5773E+03  2.8321E+01!
 ELEMENT CU   FCC_A1                    6.3546E+01  5.0041E+03  3.3150E+01!
 ELEMENT O    1/2_MOLE_O2(G)            1.5999E+01  4.3410E+03  1.0252E+02!
 ELEMENT ZR   HCP_A3                    9.1224E+01  5.5663E+03  3.9181E+01!

 PHASE BCC_A2  %  2 1   3 !
    CONSTITUENT BCC_A2  :AL,CU,ZR : O,VA% :  !

 PHASE ZRO2_CUBIC  %  2 1   2 !
    CONSTITUENT ZRO2_CUBIC  :VA,ZR% : O%,VA :  !
"""
dbf = Database(TDB)
res = calculate(dbf, ['AL', 'CU', 'VA'], ['BCC_A2', 'ZRO2_CUBIC'])
```

The error:
```
CodeWrapError: Error while executing command: /home/rotis/anaconda/envs/calphadpy3/bin/python /tmp/pycalphad-45k2zhah/setup.py build_ext --build-lib /tmp/pycalphad-45k2zhah. Command output is:
running build_ext
cythoning /tmp/pycalphad-45k2zhah/wrapper_module_096c0312_227a_478f_8eb7_0448fff6d58b.pyx to /tmp/pycalphad-45k2zhah/wrapper_module_096c0312_227a_478f_8eb7_0448fff6d58b.c
building 'wrapper_module_096c0312_227a_478f_8eb7_0448fff6d58b' extension
[...]

Issue #238, opened 2019-09-04
BLD: AppVeyor is returning passing error codes even when tests fail


richardotis:
Example: https://ci.appveyor.com/project/richardotis/pycalphad/builds/27135731/job/m03mn7enohl2o9y0 ( from https://github.com/pycalphad/pycalphad/pull/236 )

PR #239, opened 2019-09-15
FIX: pure vacancy phase detection

igorjrd:
Bugfix to detect phases containing pure vacancy sublattices, according to: https://github.com/pycalphad/pycalphad/issues/191

When any phase satisfies the requirement (containing at least one pure vacancy sublattice) an error message is raised.

### DIFF
--- a/pycalphad/model.py
+++ b/pycalphad/model.py
@@ -137,7 +137,8 @@ class Model(object):
 
         # Verify that this phase is still possible to build
         for sublattice in phase.constituents:
-            if len(set(sublattice).intersection(self.components)) == 0:
+            sublattice_comps = set(sublattice).intersection(self.components)
+            if len(sublattice_comps) == 0:
                 # None of the components in a sublattice are active
                 # We cannot build a model of this phase
                 raise DofError(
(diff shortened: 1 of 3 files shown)


pycalphad/model.py
```diff
@@ -137,15 +137,22 @@ def __init__(self, dbe, comps, phase_name, parameters=None):
...
+            if len(sublattice_comps) == 0:
                 # None of the components in a sublattice are active
                 # We cannot build a model of this phase
                 raise DofError(
                     '{0}: Sublattice {1} of {2} has no components in {3}' \
                     .format(self.phase_name, sublattice,
                             phase.constituents,
                             self.components))
-            self.constituents.append(set(sublattice).intersection(self.components))
+            if sum(set(map(lambda s : getattr(s, 'number_of_atoms'),sublattice_comps))) == 0:
```
bocklund: Thanks for submitting this PR! 

I think this line may be causing the tests to fail because *any* sublattice that has only VA will raise an error, but we only want to raise an error here if *all* sublattices have only `VA`. 

For example, given a phase with sublattices `(('FE', 'VA'), ('VA'))`, the phase should built if both `FE` and `VA` are specified (i.e. it's ok if the last sublattice is `VA` only, as long as the first sublattice has `FE` occupied).

It might be helpful to write a unit test that exercises some different conditions.

igorjrd: You're welcome! Its a pleasure to try contributing on pycalphad.

I see, as you said, my previous approach works detecting if the phase has any pure vancacy sublattice. I fixed this storing the `number_of_atoms` of each sublattice in a set called `is_pure_VA` and checking if the sum of `is_pure_VA` is equal to zero after get the `number_of_atoms` in each sublattice. Now i think that it will work, I tested building some `Models` used in the tests of TravisCI and it worked fine

bocklund:
The builds all seem to be passing (appveyor checked by hand because of #238), except for some non-test related bug where pip wants to install `certifi` over the conda installed version on macOS+Python 3.6. This is also happening over at #240. This looks ready to merge to me.

PR #240, opened 2019-09-18
ENH: Add reference data reading/writing to TDB parser

bocklund:
Adds a `refstate` attribute to the `Database` object, which holds information about the reference phase, mass, H298 and S298. 

Supports reading and writing these values.


richardotis:
How does this interact with pycalphad's reference state support?

bocklund:
Right now it doesn't interact at all within pycalphad. What this can enable is knowing what to set the reference state to if you want formation energies:

`ref_state = ReferenceState(el, dbf.refstates[el]['phase'])`

bocklund:
There may be corner cases with that simplification that don't make sense, for example an order-disorder model where the pure element endmember has an ordered contribution or the case where the database is something like:

```
ELEMENT Ni FCC_A1 0.0 0.0 0.0 !

PHASE % A1 1 1 !
```

I haven't seen such a database in the wild, but it's technically possible.

Issue #97, opened 2017-06-05
Use better colors in phase_legend()


bocklund:
The colors generated here are ok for the initial ones in the list in `phase legend`, but later ones are not appealing and hard to discern. Roughly half of the colors in the image are very hard to discern. 

The code to generate colors originally came from http://junweihuang.info/uncategorized/automatically-generate-n-distinct-colors-a-python-script/

But there might be better ways in the [original SO discussion](https://stackoverflow.com/questions/470690/how-to-automatically-generate-n-distinct-colors) where we can try to better generate ~50 default colors (currently we are using 54). 

![Sample of colors](https://cloud.githubusercontent.com/assets/7681751/26807735/1386dc30-4a0d-11e7-8d05-ff7d87f8c9d7.png)

bocklund:
The steps to accomplish this would simply be to 

1. Use one of the solutions in the [SO answers] (https://stackoverflow.com/questions/470690/how-to-automatically-generate-n-distinct-colors) and generate a better list of colors (or just take an existing list of colors)
2. replace the list of color values of the `phase_legend` function in `pycalphad.plot.utils`

Cactusmachete:
Hey!
I'm new here, but noticed that this issue is labelled beginner. Is it alright if I try and resolve this issue?

richardotis:
@Cactusmachete Welcome. Yes, you can work on resolving this issue.

bocklund:
Hi @Cactusmachete, have you made any progress addressing this? Feel free to let us know if you have any questions!

HUI1SUN:
There is a paper named A Colour Alphabet and the Limits of Colour Coding in web(https//www.researchgate.net/publication/237005166_A_Colour_Alphabet_and_the_Limits_of_Colour_Coding), which mentioned Kelly's 22 colors of maximum contrast. I'm trying to generate a new color list based on it, hoping to get a better list of colors.

richardotis:
@HUI1SUN In what order should the colors be used?

HUI1SUN:
@richardotis I'm thinking that put the order in this way. First, putting all the Kelly's colors in the front of the order, since these colors are most distinctive. Then, ranking the rest color according to their hue. And, there is the picture of ordered colors.
![figure_1](https://user-images.githubusercontent.com/37977573/43776018-3129e816-9a81-11e8-8d79-39e79b03313c.png)

richardotis:
[...]

PR #242, opened 2019-10-05
ENH: Use better colors in phase_legend()

igorjrd:
Improvement of phase_legend() colors according to issue #97 
The generated color palette is obtained ordering the colors in a way that the `i+1` color is the most distinctive compared to previous colors, according to CIEDE 2000 formula.


igorjrd:
I noticed that the failures occured at last commit were due the `#` at the strings of hex color names. The color palette that I sorted was the palette proposed by HUISUN at #97, I noticed that him posted the palette with the color `cea262` repeated, as observed. I can try to run the sorting algorithm with other colors to fill the gap in the palette.

bocklund:
@igorjrd I think what you have is great. I reformatted the list, just to even it out. 

Subjectively, I think these look much better than before. I'll go ahead and merge these changes.


```python
import numpy as np
from pycalphad.plot.utils import phase_legend
allph = list(map(str, range(100)))
leg, col = phase_legend(allph)
nc = len(set(col.values()))
x = np.random.rand(nc)
y = np.random.rand(nc)
plt.scatter(x, y, c=list(set(col.values())), s=100)
```

### Before
image

### After
image

bocklund:
Thanks again for submitting these changes and seeing this through @igorjrd!

Issue #226, opened 2019-06-11
Refactor function arguments for components/species


bocklund:
Sometimes it can be confusing when a functions like `equilibrium` and `_solve_eq_at_conditions` both take a `comps` argument, but `equilibrium` expects a list of strings, while `_solve_eq_at_conditions` expects a list of `v.Species`. Some of the docstrings for methods throughout pycalphad are out of date since Species were introduce. I propose that we do some maintenance and update the signatures of the functions to rename `comps` to `species` when the function actually takes Species objects and to update the docstrings accordingly as well.

PR #244, opened 2019-10-10
REFAC: function arguments for components/species

igorjrd:
Updating docstrings, variables names and function arguments from `comps` to `species` as requested at issue #226


richardotis:
Some of these changes look good, but I am not sure about all of them. I think renaming certain class attributes will be very confusing (opposite of what we want to do), and will also break downstream user code.

Perhaps you could start with renaming function arguments for the user-facing functions like `calculate` and `equilibrium`, and we can discuss that smaller set of changes first?

richardotis:
It would probably be easier to start a fresh PR than resolve all the conflicts at this point, so I'm going to close this. Feel free to revisit in the future, as time and interest allow.

PR #245, opened 2019-10-11
TST: CI maintenance

bocklund:


bocklund:
Closing this since it comes from bocklund:cifixes, and it doesn't look like Appveyor is using that branch's `appveyor.yml` file.

See #246 for the same PR with a branch originating from the pycalphad org's repo.

Issue #247, opened 2019-10-15
Cannot calculate the disordered phase in an order disorder model, even if it's the only active phase in the calculation


bocklund:
The disordered part of a phase in an order-disorder model gets filtered out as a possible phase by `pycalphad.core.utils.filter_phases`.

This is usually the behavior we want if the ordered phase will be active, for example in the ordered FCC_L12 phase with the disordered part FCC_A1.
```python
calculate(dbf, comps, ['FCC_A1', 'FCC_L12', 'LIQUID', 'BCC_A2'], T=300.0, N=1.0, P=101325.0)
# active phases in calculation: ['FCC_L12', 'LIQUID', 'BCC_A2'] (correct)
```
the FCC_A1 phase is made inactive in favor of the FCC_L12. However, if the ordered part is not passed in the list of phases to consider, we should allow the disordered part to be active, e.g.
```python
calculate(dbf, comps, ['FCC_A1', 'LIQUID', 'BCC_A2'], T=300.0, N=1.0, P=101325.0)
# active phases in calculate: ['LIQUID', 'BCC_A2'] (current)
# active phases in calculate: ['FCC_A1', 'LIQUID', 'BCC_A2'] (correct)
```
would use the disordered FCC_A1 phase and not the ordered FCC_L12 phase. And further, 
```python
calculate(dbf, comps, 'FCC_A1', T=300.0, N=1.0, P=101325.0)
# active phases in calculate: [] (current)
# active phases in calculate: ['FCC_A1'] (correct)

```
should work, but it currently does not (raises with the error that no phases are active).

igorjrd:
Seems like is a problem due `pycalphad.core.utils.filter_phases()` structure. `filter_phases()` is currently taking the dbf and the comps to check if the sublattices are active and, if the dbf contains simultaneously both ordered and ordered phases definition, the disordered phase model is not returned by this function. Maybe we can refactor filter_phases() to take `phases` as arg and check if the ordered phases of each phase is present in the `phases` argument.

PR #248, opened 2019-10-18
FIX: Calculating disordered phase only if respective ordered phase inactive

igorjrd:
Fixing the bug presented at https://github.com/pycalphad/pycalphad/issues/247
A new function `check_order_disorder` is now responsible to exclude the disordered phases when its respective ordered phase are active.

Some tests in `test_utils` had to be slightly changed, to run the new function `check_order_disorder` to remove the disordered phases when active phases are active.

### DIFF
--- a/pycalphad/core/calculate.py
+++ b/pycalphad/core/calculate.py
@@ -9,7 +9,7 @@ from pycalphad import ConditionError
 from pycalphad.core.utils import point_sample, generate_dof
 from pycalphad.core.utils import endmember_matrix, unpack_kwarg
 from pycalphad.core.utils import broadcast_to, filter_phases, unpack_condition,\
-    unpack_components, get_state_variables, instantiate_models
+    unpack_components, get_state_variables, instantiate_models, check_order_disorder
 from pycalphad.core.light_dataset import LightDataset
 from pycalphad.core.cache import cacheit
 from pycalphad.core.phase_rec import PhaseRecord
@@ -318,6 +318,7 @@ def calculate(dbf, comps, phases, mode=None, output='GM', fake_points=False, bro
     # Consider only the active phases
     list_of_possible_phases = filter_phases(dbf, comps)
     active_phases = sorted(set(list_of_possible_phases).intersection(set(phases)))
+    active_phases = check_order_disorder(dbf, active_phases)
     active_phases = {name: dbf.phases[name] for name in active_phases}
     if len(list_of_possible_phases) == 0:
         raise ConditionError('There are no phases in the Database that can be active with components {0}'.format(comps))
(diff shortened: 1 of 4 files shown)


igorjrd:
@richardotis, I will deprecate the proposed `check_order_disorder` combining this on `filter_phases`. To turn possible calculating disordered phases when its respective ordered phases are inactive, `filter_phases` will need to take the `phases` entried by the user as argument. As consequence, the tests will need to be reedited to add `phases` argument to `filter_phases`. Does this propose seems good?

richardotis:
@igorjrd Please move forward with your plan. I recommend following Brandon's suggestion on using a new keyword argument in `filter_phases`.

pycalphad/core/calculate.py
```diff
@@ -316,7 +316,7 @@ def calculate(dbf, comps, phases, mode=None, output='GM', fake_points=False, bro
     largest_energy = 1e10
 
     # Consider only the active phases
-    list_of_possible_phases = filter_phases(dbf, comps)
+    list_of_possible_phases = filter_phases(dbf, comps, phases)
     active_phases = sorted(set(list_of_possible_phases).intersection(set(phases)))
```
igorjrd: The new changes permit check evaluate the intersection between `candidate_phases` and database phases within `filter_phases`. So, what do you think about move `active_phases = sorted(set(list_of_possible_phases).intersection(set(phases)))` to `filter_phases`?

richardotis: Yes, I think that would now be correct behavior.

pycalphad/core/equilibrium.py
```diff
@@ -186,14 +186,7 @@ def equilibrium(dbf, comps, phases, conditions, output=None, model=None,
...
         raise NotImplementedError('Broadcasting cannot yet be disabled')
     comps = sorted(unpack_components(dbf, comps))
-    phases = unpack_phases(phases) or sorted(dbf.phases.keys())
-    # remove phases that cannot be active
-    list_of_possible_phases = filter_phases(dbf, comps)
-    active_phases = sorted(set(list_of_possible_phases).intersection(set(phases)))
-    if len(list_of_possible_phases) == 0:
-        raise ConditionError('There are no phases in the Database that can be active with components {0}'.format(comps))
-    if len(active_phases) == 0:
-        raise ConditionError('None of the passed phases ({0}) are active. List of possible phases: {1}.'.format(phases, list_of_possible_phases))
```
igorjrd: Coveralls report leads me to notice that this piece of code has become redundant after the last modifications in `filter_phases()`

richardotis: I suggest reverting this particular change, and changing it to call `filter_phases` twice: once with the `candidate_phases` keyword argument set to `None`, and the second time with `candidate_phases` set to `phases`. After the first call, check for `list_of_possible_phases` having length zero, as has been done, and then after the second call, you check `active_phases`.

This way, we can keep the `ConditionError` in `equilibrium` and `calculate`, where they have the most meaning, and `filter_phases` will not raise that exception when it might not make sense.

richardotis:
Thanks again for your work @igorjrd

PR #250, opened 2019-12-03
Rewrite pointsolver to require passage of the problem instance, not the class name

jwsiegel2510:


bocklund:
Sorry, I didn't realize these changes were to make a pull into the pointsolver branch. Merging now.

PR #251, opened 2019-12-03
MAINT: SymPy 1.5 compatibility

bocklund:
SymPy 1.5 release candidate 1 is out now. There is one breakage: EmptySet is now a singleton instead of a class ([see release notes](https://github.com/sympy/sympy/wiki/Release-Notes-for-1.5#backwards-compatibility-breaks-and-deprecations) ).

Here is the simplest change. We also need to bump the requirements to pin sympy>=1.5.

There are a couple other options for changes:

1. Current change, use the singleton comparison instead of the class
2. Use the `.is_EmptySet` property, which is now deprecated in favor of the `.is_empty` property. Allows us to stay backwards compatible until the deprecated code is removed (https://github.com/sympy/sympy/issues/17523)
3. Use the `.is_empty` property, which would still require bumping the version to sympy>=1.5

Staying with comparing to the singleton makes sense to me. I'm opening this PR for discussion. We cannot merge this until SymPy 1.5 is released.

### DIFF
--- a/conda_recipe/meta.yaml
+++ b/conda_recipe/meta.yaml
@@ -31,7 +31,7 @@ requirements:
     - pandas
     - pytest
     - xarray >=0.11.2
-    - sympy ==1.4
+    - sympy ==1.5
     - pyparsing
     - tinydb >=3.8
     - scipy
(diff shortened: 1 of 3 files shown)


bocklund:
SymPy 1.5 is released and tests are passing here (except for Python 3.8, see #257)

This is ready for comments @richardotis

bocklund:
Rebased on develop after merging #257 - will merge after tests pass

PR #252, opened 2019-12-04
BLD: Fix Travis CI Python 3.7 on macOS, build on Python 3.8 on macOS and Linux

bocklund:


richardotis:
I think this issue is related: https://github.com/cython/cython/issues/2694

One way to side-step this would be to switch all the Cython modules back to C, instead of C++. (We swapped to C++ during the transition to SymEngine, but upstream API improvements have made the original reasons for that moot.)

bocklund:
conda-forge CI seems to use 
```yaml
macos_machine:
- x86_64-apple-darwin13.4.0
```

(see https://github.com/conda-forge/pycalphad-feedstock/blob/master/.ci_support/osx_python3.7.yaml)

I am going to try to pin the `osx_image` on Travis to something on High-Sierra with an older xcode version to see that that fixes.

bocklund:
Python 3.7 tests are building and pass on macOS now. Python 3.8 builds show the regression fixed by #257. @richardotis okay to merge this?

richardotis:
Okay.

PR #211, opened 2019-04-22
Package updates

bocklund:
* Bump and pin to SymPy==1.4: Fixed a bug where Unions and Intersections do not automatically de-nest arguments (sympy/sympy#16344) anymore. Pinning to 1.4 because pycalphad is tightly coupled to SymPy and we don't want to break user environments when SymPy 1.5 is released.
* Update CI infrastructure to test on Py2.7, 3.6, 3.7
* Updates to dask
* Updates to Xarray (closes #204), minimum version is last supported release of Python 2.7

PR #257, opened 2019-12-18
FIX: Fix Python 3.8 deprecation of ast.Num in favor of ast.Constant

bocklund:
Accoring to the [Python 3.8 `ast` docs](https://docs.python.org/3.8/library/ast.html?highlight=ast#module-ast):

> Deprecated since version 3.8: Class ast.Constant is now used for all constants. Old classes ast.Num, ast.Str, ast.Bytes, ast.NameConstant and ast.Ellipsis are still available, but they will be removed in future Python releases.

The tdb.py module whitelists certain `ast` objects, including `ast.Num`, which are now `ast.Constant`, causing breakages in sympifying parsed strings. This adds `ast.Constant` to our whitelist.

Thanks @jwsiegel2510 for pointing this out at https://github.com/PhasesResearchLab/ESPEI/pull/110

As an aside: this is an interesting breakage because `ast.Num` still exists in Python 3.8, but instantiating an `ast.Num` actually gives you an `ast.Constant` instance, which breaks the whitelist even though people creating objects with `ast.Num` in Python 3.8 are not technically affected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,8 @@ install:
       echo "!!! Installing pycalphad dependencies via conda"
       conda install --yes python=$TRAVIS_PYTHON_VERSION numpy scipy matplotlib pytest pytest-cov pandas sympy pyparsing dask dill python-symengine xarray cython cyipopt
       echo "!!! conda installing test packages"
-      conda install --yes sphinx sphinx_rtd_theme coveralls ipython
+      # coverage<5, see https://github.com/coveralls-clients/coveralls-python/issues/213
+      conda install --yes sphinx sphinx_rtd_theme coveralls "coverage<5" ipython
       echo "!!! pip pycalphad as editable"
       pip install -e .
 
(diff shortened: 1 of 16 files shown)


bocklund:
Note that `ast.Constant` has existed since Python 3.6, so merging this commit would effectively drop Python 2.7 support. Since Python EOL is January 1 and many of our dependencies are no longer supporting Python 2.7, I'm okay with pycalphad 0.8.1.post1 being the last Python 2.7 release.

bocklund:
Rebased onto develop after merging #252 to verify that the changes fix Python 3.8 and all the tests pass on CI. Note the failing Python 2.7 tests per my comment above.

bocklund:
We haven't been testing 3.5 since we dropped support along with conda-forge (#211). I think conda-forge is only supporting the latest two versions of Python, so any compatibility we had with Python 3.5 was unofficial.

The latest changes drop Python 2.7 from the tests and removes most of the obvious Python 2 vs. Python 3 specific code. I'll merge after the tests are passing

richardotis:
The coverage change is noise in comparison to the coverage difference between the Python versions we test. You can merge anyway.

PR #258, opened 2019-12-22
FIX: parameters with components not in phase constituents producing incorrect models

### LINKED ISSUE #256: Parameters with active components that are not in the Phase constituent array produce incorrect models
The `Phase` object's constituent array (corresponding to the `CONSTITUENT` TDB keyword) defines the only species allowed to be active in each sublattice.

If a parameter is defined with an active species in a sublattice and that species is not in the Phases's constituent array it will produce a `Model` that is inconsistent with the Phase constituent array, which produces either convergence failures at best and silent garbage data or segmentation faults at worst. This is because the parameter searches by Model do not filter out parameters that have active species in sublattices they do not belong in.

Below are two failing tests that should pass when this is fixed. In this case, `B` is an active component in the ALPHA phase, but according to ALPHA's constituent array, B can only be present in the second sublattice, not the first. In the database below, there's a parameter that defines B in the first sublattice, which it cannot exist in.

A fix for this issue would be for Models to be created that do not include parameters that are not consistent with constituent array. A warning should also be emitted, since the either the parameter or constituent array may be unintentionally inconsistent and the user should be alerted to fix the problem.

```python
from pycalphad import Database, Model, equilibrium, variables as v
import numpy as np

TDB_BAD_CONSTITUENTS = """
ELEMENT A BLANK 0 0 0 !
ELEMENT B BLANK 0 0 0 !

TYPE_DEFINITION % SEQ * !
DEFINE_SYSTEM_DEFAULT ELEMENT 2 !

PHASE ALPHA %  2 0.5 0.5 !
CONSTITUENT ALPHA :A: A,B :  !

$  ALPHA  $
PARAMETER G(ALPHA,A:A;0) 1 -10; 10000 N !
PARAMETER G(ALPHA,A:B;0) 1 -10; 10000 N !

$ Erroneous parameter! B is not in the first sublattice
PARAMETER G(ALPHA,B:A;0) 1 1e5; 10000 N !
"""

def test_bad_constituents_not_in_model():
    dbf = Database(TDB_BAD_CONSTITUENTS)
    mod = Model(dbf, ['A', 'B'], 'ALPHA')
    assert v.SiteFraction('ALPHA', 0, 'B') not in mod.ast.free_symbols


def test_bad_constituents_do_not_affect_equilibrium():
    dbf = Database(TDB_BAD_CONSTITUENTS)
    assert np.isclose(equilibrium(dbf, ['A', 'B'], ['ALPHA'], {v.P: 101325, v.T: 300, v.N: 1, v.X('B'): 0.5}).GM.values.squeeze(), -10.0)
```

igorjrd:
Some improvements in filtering parameters methods, mainly `Model._purity_test`. New tests related to tests `_array_validity` and `_interaction_test`. Fixes #256.


pycalphad/model.py
```diff
@@ -452,19 +451,11 @@ def _interaction_test(self, constituent_array):
...
-            # check if all elements involved are also active
-            valid = set(sublattice).issubset(self.components) \
-                or sublattice[0] == v.Species('*')
-            if len(sublattice) > 1 and valid:
-                result = True
-            if not valid:
-                result = False
-                break
-        return result
+        if True in [True for sublattice in constituent_array if len(sublattice) > 1]:
```
bocklund: Instead of these lines, what about

```python
        return any(len(sublattice) > 1 for sublattice in constituent_array)
```

pycalphad/model.py
```diff
@@ -422,15 +422,14 @@ def build_phase(self, dbe):
...
         Check if constituent array only has one species in its array
-        This species must also be an active species
+        This species must also be an active species and contained in model sublattice
         """
-        if len(constituent_array) != len(self.constituents):
+        if not self._array_validity(constituent_array):
             return False
-        for sublattice in constituent_array:
-            if len(sublattice) != 1:
+        for param_sublattice, model_sublattice in zip(constituent_array, self.constituents):
```
bocklund: Correct me if I’m wrong, but shouldn’t the `_array_validity` check assure that `param_sublattice[0]` is always in the model sublattice or is a `*` (making the check redundant)? This seems like it could break in the case of a `*` anyway. 

If that’s true, could we just change the lines after the validity check to 

```python
return any(len(sublattice) != 1 for sublattice in constituent_array)
```

pycalphad/tests/test_model.py
```diff
@@ -68,4 +69,31 @@ def test_detect_pure_vacancy_phases():
...
+    assert v.SiteFraction('ALPHA', 0, 'B') not in mod.ast.free_symbols
+
+def test_bad_constituents_do_not_affect_equilibrium():
+    dbf = Database(TDB_PARAMETER_FILTERS_TEST)
+    assert np.isclose(equilibrium(dbf, ['A', 'B'], ['ALPHA'], {v.P: 101325, v.T: 300, v.N: 1, v.X('B'): 0.5}).GM.values.squeeze(), -10.0)
+
+def test_interation_test_method():
+    dbf = Database(TDB_PARAMETER_FILTERS_TEST)
+    interacting_consts = dbf.search(
+        where('parameter'))[4]['constituent_array']
```
bocklund: Is the order of the parameters returned by the search method guaranteed to be in the same order by TinyDB? If not, this test could be a source of flaky, random failures. (This goes for all tests here that assume an order of the parameter list)

bocklund:
Thanks @igorjrd, I think this is close and the changes and improvements are welcome. When you make your next changes, will you please rebase or merge in the develop branch so the tests will pass on Travis and Appveyor across Python 3.6-3.8, which we are now supporting?

pycalphad/model.py
igorjrd: `_array_validity` really checks if `param_sublattice[0]` is always in the model sublattice or is a `*`, but it does not ensure that `param_sublattice[0]` is contained in the correspondent Model sublattice. Actually, `if param_sublattice[0] not in model_sublattice` is the code line that fixes the bug related in the issue 256.
`_array_validity` returns True if each sublattice species is contained at any Model sublattice, while these changes ensure that each parameter sublattice is contained in respective Model sublattice (respective sublattice, instead of any sublattice).
Now seeing this, I realized: does not make more sense to perform the checking if parameter sublattices are contained in respective Model sublattices in `_array_validity` and just check if there are one specie by sublattice in `_purity_test`?

bocklund: Yes, you’re right. I thought that’s what `_array_validity` was doing, perhaps I shouldn’t review code so late at night 😬

> Now seeing this, I realized: does not make more sense to perform the checking if parameter sublattices are contained in respective Model sublattices in `_array_validity` and just check if there are one specie by sublattice in `_purity_test`?

Yeah, this makes more sense to me, since that requirement is common to both. I think the only thing that needs to change in `_array_validity` is to check if the sublattice constituents are a subset of the Model sublattice, rather than the components. 

If we do that, then we can make the simplifications I suggest in these comments.

pycalphad/tests/test_model.py
richardotis: The default tinydb storage implementation uses dicts, so I would expect order to be preserved as long as the database isn't modified, but this detail should not be relied upon.

richardotis: I think a better way to write this test would be to just check the result (that you get an error when you expect). That way we are free to adjust internal Model APIs without needing to modify tests.

igorjrd: @bocklund I have tried many tests and the order of parameters seems to not change. Anyway, I was rewriting the search conditions with the addition of tests to get only the parameters with specified sublattices. With this change, the TDB could be modified with minor chance to influence in tests.
@richardotis until now, the way that I see to make errors be raised due `_array_validity`, `_purity_test` or `_interaction_test` is testing their with phases or comps conditions that would build a phase with no parameter (I am not sure if actually this raises an error, but I really expect that it raises). Do you think that this would be a nice design to it?

richardotis: @igorjrd My preference would be just to calculate the energy at one temperature and composition. If the database returns the wrong parameters, you get the wrong energy. You just have to first make sure that the test fails without your fix.

igorjrd: I submitted a new commit that tests the functions using parameters selected by sublattices comps, but I can do some effort to write tests according your suggestion, @richardotis. (I just sent this, because the work was almost done, but it can be reverted).

pycalphad/model.py
igorjrd: Not problem about reviewing code late at night! I think that I applied all suggestions in new sent commits!

bocklund:
Hey @igorjrd, sorry it's been awhile since I've looked at this. I think this is almost ready. 

The tests are improved now that they don't depend on the order of parameters in the database, but I agree with Richard that the tests could be improved:

> @richardotis: My preference would be just to calculate the energy at one temperature and composition. If the database returns the wrong parameters, you get the wrong energy. You just have to first make sure that the test fails without your fix.

I think adding some tests for the energy to your tests would be enough to merge what you have. See the `check_energy` function and examples from the [test_energy.py](https://github.com/pycalphad/pycalphad/blob/develop/pycalphad/tests/test_energy.py) module for how to do this.

bocklund:
@igorjrd are you still interested in working on this PR?

igorjrd:
> 
> 
> @igorjrd are you still interested in working on this PR?

Sorry about this long time with no updates. I'm still interested in working on this PR! These last times I could not work properly on this PR, due my masters research activities.
However, in Brazil we are starting to adopt quarentine due covid-19 crysis. I think that until next week I would resume the work on this PR.

bocklund:
Great, no rush! Your research should definitely be your priority. 

Stay safe and let us know if you have any questions when you are able to pick this up again.

pycalphad/model.py
```diff
@@ -463,13 +451,18 @@ def _array_validity(self, constituent_array):
             # species can be specified in the anion sublattice without any
             # species in the cation sublattice.
             if self._dbe.phases[self.phase_name].model_hints.get('ionic_liquid_2SL', False):
-                pass
+                for sublattice in constituent_array:
+                    if (set(sublattice).issubset(self.constituents[1]) \
```
igorjrd, line 455: Is this `self.constituents[1]` appropriate here, or would be better to check if parameter sublattice is contained in `self.components`?

I think this PR is ready or almost ready. Do you think that should I remove the excessive tests and keep just the test now named as `test_model_energy`?

bocklund:
I'm sorry @igorjrd. I totally lost track of the fact that you made changes here in May and it was on me to respond to them.

I took what was here and cleaned it up a little. The code in `Model.py` is effectively unchanged from @igorjrd implemented. I cleaned up the test cases as @richardotis suggested. After looking at them, they were mostly redundant tests on the same database and checking whether the site fraction symbols are in the Model instance AST is enough to exercise these changes.

The tests are [here](https://github.com/igorjrd/pycalphad/actions) and are passing on all platforms (Coveralls is failing because @igorjrd's fork doesn't have our secrets) - we're paying that penalty for that PRs from forks, heh 😁.

I've requested reviews from both @igorjrd and @richardotis. Please let me know if you agree to merging this with your review!

bocklund (approved):

richardotis (approved):

igorjrd:
Don't worry, Brandon. I didn't warned you about the changes. Also, in these last months it was difficult to me to provide new contributions on pycalphad, what turned hard to me to contact you about active PR's progress.

I checked the changes and agreed with them. I also have to say sorry about the redundant tests and thanks for removing the redundant. I think this PR is ready to be merged!

PR #260, opened 2020-01-17
ENH: Quasichemical model

richardotis:
This PR adds support for the Modified Quasichemical Model (MQM), as described in many publications, most notably the series of papers starting with:
Pelton, A. D., Degterov, S. A., Eriksson, G., Robelin, C., & Dessureault, Y. (2000). The modified quasichemical model I—binary solutions. _Metallurgical and Materials Transactions B_, 31(4), 651-659. http://www.crct.polymtl.ca/courses/17-MQM-1.pdf

The MQM is important for accurate thermochemical descriptions of phases exhibiting strong short-range atomic ordering, such as in steel slag liquids.

In contrast to the Compound Energy Formalism (CEF), in the MQM, the unit of the model is not the atom, but rather the A-B quasimolecule (or "bond"). The energy of the phase is described as a function of the population of these "bonds," with mass conversation brought in by specifying effective coordination numbers, Z_i, for each species A and B. (Species are allowed to have more than one coordination number, e.g., Fe 3+ and Fe 2+.) Note that, despite the similarity, coordination number is *not* a charge. In this implementation of the MQM, all species and quasimolecules are neutral (but see below).

In this PR, we only support pairwise quasimolecules (A-B type), but we note that there is an extension to the MQM, using two sublattices and called the quadruplet approximation, which admits a description of A-B-X-Y quasimolecules. We defer this feature to future work.

The MQM is most famously implemented in the FactSage software, though Thermo-Calc also offers support. There are several challenges to implementing the MQM in software designed for the CEF: 
* The TDB file format offers no obvious encoding for the coordination number. In Thermo-Calc's implementation (TC Examples 49 and 50), the `VK` parameter type, ostensibly for high-pressure models, is reused for encoding the Z_i.
* The quasimolecules are implemented by taking advantage of an equivalence* between the MQM and the associate model: a quasimolecule behaves just like an associate (molecule) with atomic ratios specified by m/z, where m is the multiplicity and z is the coordination number. This means an A-A quasimolecule can be encoded as a A_(2/z) SPECIES, and an A-B quasimolecule can be encoded as a A_(1/zA)B_(1/zB) SPECIES. This equivalence seems to break down with some of the other binary mixing models (see below), and had to be (?) special-cased.
* These models tend to use more advanced mixing models, which necessitated implementing the polynomial (POLYNOM) binary excess model, as well as support for the TDB commands to switch them on. The required order for component specification was not obvious and had to be determined by trial-and-error.
* Normalization of parameters has to be done carefully, though this is more of a concern for assessment than calculation.

\* Note that the configurational entropy is calculated differently in the MQM versus the associate model. In the TDB file format, a `QUASICHEM_FACT00` type definition is used to tell the software to calculate the entropy according to the MQM. (It also appears to change how the binary mixing parameters are normalized, but this is undocumented.)

Things not yet done:
- [ ] Ternary extrapolation models, such as Kohler-Toop (or any kind of ternary testing)
- [ ] Legendre polynomial binary mixing models
- [ ] Writing `EXCESS_MODEL` TDB commands back out (reading works)


richardotis:
Superseded by #389

Issue #154, opened 2017-12-06
Develop a wrapper API for plotting properties from calculate and equilibrium


bocklund:
An API to a general effect of `propplot(res_dataset, x='T', y='CPM')` (name TBD). Should work equally well for equilibrium properties and calculated properties and be generally aware of phases (e.g. different color line segments in equilibrium properties, and different lines in calculate properties)

This is somewhat parallel with #133 in that it would be nice to do one calculate call that you could call `propplot` on many times. 

For now, since it would be reasonable to plot many properties from one dataset, I think it makes sense that this has more of an `eqplot` signature (it takes a dataset) rather than a `binplot` or `ternplot` style signature.

richardotis:
We could probably also use this to fix #55

bocklund:
Some out of band discussion on this today. To summarize:

A new version of `eqplot` can handle plotting both phase diagrams and property diagrams. It should detect them automatically and dispatch to `pdplot` for phase diagrams and `propplot` for property diagrams. A version of a general version of `pdplot` for binary and ternary systems is written as the `eqplot` function [here](https://github.com/PhasesResearchLab/PSU-MatSE410/blob/master/ternary/eqplot_helper.py) (to be worked in to pycalphad). `pdplot` and `propplot` can exist as public "advanced usage" to allow for special cases when `eqplot` may make the wrong assumption.

The key distinction between `pdplot` and `propplot` is that it is assumed that `v.MoleFraction` (`v.X`) axes in `pdplot` should correspond to phase compositions (the way they `eq.X` is for an equilibrium `Dataset`) while `v.X` axes for `propplot` means overall composition (i.e. `np.dot(eq[X].sel(component='B'), eq[NP])` or `eq.X_B`).

It also needs to be decided how to give the axes for `Dataset` data variables that are not state variables (i.e. `eq.NP`). Are they just input as strings? For strings, maybe you try to look them up in xarray first, then fall back to getattr(pycalphad.variables, axis_var), then try to parse special cases v.X from the string.

In addition, the following features might be useful:
1. `propplot` can also take `Dataset` objects produced by `calculate` as well as `equilibrium`
[...]

PR #262, opened 2020-02-06
DOC: Update examples

bocklund:
* Binary examples updated
  - Fix bug in `ax=ax` and `tielines=False` for the newer `binplot` API.
  - More details at the top about where to get TDBs and how to make sure it's in the right directory for a user's calculation
  - Make all the binary examples (phase diagrams and energies) self contained in each cell for easier copy-pasting
  - Comment all the binary examples so copy-pasters have some explanation
  - Add Al-Mg example, since it's a more complicated diagram than Al-Zn, but is still relatively fast (~40s on my machine). Contrast this with the other non-trivial examples (Al-Ni, Al-Fe, Nb-Re) which are all relatively slow.
  - Add citations for papers that exist with assessments.
  - Add axes labels to Gibbs energy curves example
* Add example for calculating and plotting along a specific internal degree of freedom


bocklund:
This is ready for review.

See the two updated docs here:
https://github.com/pycalphad/pycalphad/blob/doc-update-examples/examples/BinaryExamples.ipynb
https://github.com/pycalphad/pycalphad/blob/doc-update-examples/examples/ReferenceStateExamples.ipynb (only change is the new last cell and text)

richardotis:
Minor things:

1. > The Thermodynamic DataBase DataBase (TDBDB) has index many available databases

indexed

2. We still have a hard time with the really flat liquidus in the Nb-Re diagram. Not really in scope for this PR, but an observation.
3. "Enthalpy of mixing" first example
```python
ax.scatter(result.X.sel(component='RE'), result.HMR, marker='.', s=5, label='CHI_RENB')
ax.set_xlim((0, 1))
ax.set_xlabel('X(RE)')
ax.set_ylabel('HM_MIX')
ax.set_title('Nb-Re CHI Mixing Enthalpy')
```
I think both of these should be labeled liquid, not chi.

4. The convex hull example is really cool because it shows off the power of the multi-dimensional representation of the calculation results.
5. Last reference state example: `result.Y.squeeze()[:, 2]` - for a new person this syntax is confusing. A portion of the text explanation at the top (about the site fraction ordering) should be copied into a comment in the code example. Then something like `Y_CHI_1_RE = result.Y.squeeze()[:,2]` before using the new array in the plot.

bocklund:
1. Fixed
2. Yeah. I think it could be improved for that particular diagram by increasing the point density when calculating the grid, so that the two phase regions would be picked up by the hyperplane. Not sure whether that would take away from the example (or if that example should be removed or moved somewhere else with those "advanced" API usage). A line following mapping approach might be a longer term fix.
3. Fixed
4. Thanks, I agree! I think whatever the solution for #154 looks like will help to make these even easier to create
5. Comments added and `Y_CHI_1_RE` factored out.

PR #264, opened 2020-02-20
FIX: Support PhaseRecord pickling, switch SymEngine backend to LLVM (configurable)

bocklund:
This PR includes the following changes:

1. PhaseRecords can now be pickled, provided that all callables are built with LLVM callables. This is from adding internal references (e.g. `ofunc_` to to callable functions that are pickled and reused to create the `FastFunction` objects when unpickling.
2. LLVM callables are now the default everywhere, but this can be configured at runtime or through the `build_functions` or `build_constraint_functions` function APIs.
3. Improved module documentation for `sympydiff_utils` and `build_functions`/`build_constraint_functions`.
4. Updates to pin to `symengine=0.6`. This drastically improves the compile time performance for some of the worst-case systems and makes the compile times go from impossible on LLVM (>20,000 sec) to possible (~30-60 sec).
5. Updates releasing docs and GH PR template to include a checklist item for updating dependencies everywhere (this has been missed in release and tests few times)

Some performance benchmarks for the test suite:

Lambda: 138 s
LLVM `opt_level = 3`: 174 s
LLVM `opt_level=0`: 135 s
LLVM `opt_level=3` everywhere, Lambda Hessians: 125 s

From this data, it does appear that there's some runtime/compile time performance trade off, and the most reasonable defaults is LLVM everywhere (so everything can be pickled).

I'm pretty happy with the design and ability to override the options, but one potential gotcha is that if `build_functions` is called and the result cached, a subsequent change to the `DEFAULT_LAMBDIFY_XYZ` options will not take effect for the cached function (non-cached functions will work as expected). I think this is a pretty small tradeoff for the flexibility of this approach, and the fix for the aforementioned "gotcha" is just to restart the Python kernel and don't compile until the global `DEFAULT_LAMBDIFY_XYZ` is changed.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,4 +9,7 @@ Below are a few things we ask you kindly to self-check before getting a review. 
 
 Checklist
 * [ ] The documentation examples have been regenerated if the Jupyter notebooks in the `examples/` have changed. To regenerate the documentation examples, run `jupyter nbconvert --to rst --output-dir=docs/examples examples/*.ipynb` from the top level directory)
-
+* [ ] If any dependencies have changed, they are changed are reflected in the
+  * [ ] `setup.py`
+  * [ ] `.travis.yml`
+  * [ ] `appveyor.yml`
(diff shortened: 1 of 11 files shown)


richardotis:
Does it need a test?

richardotis:
What about PhaseRecord pickling itself? Or is that too micro of a test?

bocklund:
Oh yeah, I forgot about that

bocklund:
@richardotis ready to merge this with the added tests?

richardotis:
Please merge

PR #266, opened 2020-03-04
ENH: Reading large databases

richardotis:
I noticed `Database('steel_database_fix.tdb')` can take a long time to complete, 40-90 seconds depending on the machine. Switching to tinydb's `insert_multiple` API is a cheap and easy way to cut that time in ~half.

Most of the remaining time is in parsing the grammar for all the PARAMETERs, as well as calls to sympify for the function expressions. There's room to improve there (may be possible to get the time down to ~6 seconds), but those changes may break some internal or external APIs and will be more complicated to implement. This is the low-hanging fruit.

### DIFF
--- a/pycalphad/io/database.py
+++ b/pycalphad/io/database.py
@@ -211,7 +211,6 @@ class Database(object): #pylint: disable=R0902
         try:
             dbf = Database()
             format_registry[fmt.lower()].read(dbf, fd)
-            dbf._parameter_queue_process()
         finally:
             # Close file descriptors created in this routine
             # Otherwise that's left up to the calling function
@@ -473,7 +472,7 @@ class Database(object): #pylint: disable=R0902
         """
         return self._parameters.search(query)
 
-    def _parameter_queue_process(self):
+    def process_parameter_queue(self):
         """
         Process the queue of parameters so they are added to the TinyDB in one transaction.
         This avoids repeated (expensive) calls to insert().
--- a/pycalphad/io/database.py
+++ b/pycalphad/io/database.py
@@ -94,6 +94,7 @@ class Database(object): #pylint: disable=R0902
             obj.refstates = {}
             obj._structure_dict = {} # System-local phase names to global IDs
             obj._parameters = TinyDB(storage=MemoryStorage)
+            obj._parameter_queue = []
             obj.symbols = {}
             obj.references = {}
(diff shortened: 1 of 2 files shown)


pycalphad/io/database.py
```diff
@@ -210,6 +211,7 @@ def from_file(fname, fmt=None):
         try:
             dbf = Database()
             format_registry[fmt.lower()].read(dbf, fd)
+            dbf._parameter_queue_process()
```
bocklund: Would it make more sense for the format parsers to call `dbf._parameter_queue_process()`, since they would be the ones override `force_insert`? Trying to think about ways that the database could possibly be put into an inconsistent state by adding parameters manually.

Alternatively, would it make sense to come up with a solution to update process the queue automatically, maybe when when accessing `_parameters`?

pycalphad/io/database.py
```diff
@@ -465,3 +472,13 @@ def search(self, query):
         >>>> db.search(where('eid') == eid)
         """
         return self._parameters.search(query)
+
+    def _parameter_queue_process(self):
```
bocklund: I think `_process_parameter_queue()` could be a better name, since it's clearer about what calling the method actually does. Thoughts?

pycalphad/io/database.py
richardotis: 1. Yes, I'll move the function call.
2. It should be difficult to get the database into an inconsistent state: `force_insert` is designed to use the old behavior by default, so code unaware of this change should work the same.
3. Some kind of lazy/dynamic parsing of parameters would be a UX improvement here, definitely, since in most cases the vast majority of parameters are unused in downstream calculations. Getting this right would probably require breaking the API used by Model, or some really clever thinking.

bocklund: A hack could just be to change the tinydb object (`self._parameters`) to `self._parameter_database = TinyDB(storage=MemoryStorage)` probably needs a better name)

```python
@property
def _parameters():
    if len(self._parameter_queue()) > 0:
        self.process_parameter_queue()
    return self._parameter_database
```

richardotis: Oh, I see. We're talking about two different things. That's a good idea.

bocklund:
@richardotis still want to make changes on this or is it ready?

PR #268, opened 2020-03-29
ENH: TDB improvements

bocklund:
* Added a method to `Database` which goes through the parameters and checks for duplicate parameters, `Database.check_duplicate_parameters`. Duplicate parameters, i.e. when `PARAMETER G(FCC_A1,CU:VA;0)` is defined more than once, seem to be ignored in Thermo-Calc. That is, only the first entered parameter seems to be accepted. In pycalphad, duplicate parameters are accepted and are additive when constructing `Model` objects. Sometimes, a user might intend for this behavior, but the vast majority of the time having duplicate parameter is an error in the Database. This method allows for checking through all the parameters in the database and searching for duplicates. The function exists as a separate method, rather than checking at Database construction, because it could slow down the TDB reader if parameters are checked at the time of construction (and parameters are no longer inserted one at at time since #266). If a user tries to write a Database to a TDB that has duplicate parameters, it will fail because comparison of two `ParamTuple` objects in the Database writer will try to compare the SymPy expressions (since all the previous fields compare equal). This method could help a user track down which parameter causes that issue.
* Fix writing Piecewise intervals for cases where TDBs have an interval size of zero. One example is `FUNCTION UN_ASS 300 0; 300 N !`, where the start and end temperature in the Piecewise interval is the same. SymPy will simplifiy this to an empty `Piecewise()` (where pycalphad appends a `(0, True)` expression) and the TDB writer will fail because it will try to `Union(*[intervals])`, where the interval for the bad interval is an empty set. This behavior may be troublesome because a function `FUNCTION XYZ 300 100; 300 N !` would be extrapolated by Thermo-Calc and take effect anywhere it is used, but in pycalphad the zero-width Piecewise interval will be reduced and eliminated.

Left to do:
* [ ] Personally, I'm not sure how I feel about adding the duplicate parameter check. The check is important, but having a public method to do it is setting off some feature creep alarm inside of me. Could use some input here.
* [ ] Add support for parsing/ignoring some additional TDB syntax. For example: type definition `IF-THEN` statements should be ignored, as well as composition set type definitions. The `PHASE` keyword supports an auxiliary string after the sublattice stoichiometric ratios, but pycalphad cannot parse it. We should at least allow parsing these phases and then throw away the auxiliary string, but perhaps one day it could be read/written to. 
* [x] Support more aggressively adding order/disorder model hints to both ordered and disordered Phase objects. This is causing a bug in some databases (e.g. COST-507) where the `BCC_A2` and `BCC_B2` phase pass filter phases because `BCC_A2` does not have the type definition applied to it). Completed by #269

### DIFF
--- a/pycalphad/io/database.py
+++ b/pycalphad/io/database.py
@@ -3,7 +3,6 @@ The database module provides support for reading and writing data types
 associated with structured thermodynamic/kinetic data.
 """
 from io import StringIO
-import warnings
 from tinydb import TinyDB, where
 from tinydb.storages import MemoryStorage
 from datetime import datetime
@@ -404,13 +403,6 @@ class Database(object): #pylint: disable=R0902
             'reference': ref
         }
         if force_insert:
-            match_search = (where('phase_name') == new_parameter['phase_name']) & \
-                           (where('constituent_array') == new_parameter['constituent_array']) & \
-                           (where('parameter_type') == new_parameter['parameter_type']) & \
-                           (where('parameter_order') == new_parameter['parameter_order'])
-            existing_matching_params = self._parameters.search(match_search)
-            if len(existing_matching_params) > 0:
-                warnings.warn(f'Parameters matching {new_parameter} already in the database')
             self._parameters.insert(new_parameter)
         else:
             self._parameter_queue.append(new_parameter)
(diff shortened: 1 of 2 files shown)


Ebert-Daniel:
Hello all,

I'm new using pycalphad. I'm trying to calculate the HM_MIX of my phase for different disordered fraction sets of a A2-B2 ternary phase. The CEF is given:
A2: (A,B,VA)1(A,B,VA)6
B2: (A,B,VA)0.5(A,B,VA)0.5(C,VA)6
I could calculate the HM of the phase but when I set: output='HM_MIX' I have this error:
_---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-53-ca5b8ff06e86> in <module>
----> 1 result_HM = calculate(dbf_feti, comps, 'BCC_B2',P=101325, T=1400, output='HM_MIX', pdens=1000)

~/opt/pycalphad/pycalphad/core/calculate.py in calculate(dbf, comps, phases, mode, output, fake_points, broadcast, parameters, to_xarray, **kwargs)
    375                                    models=models, parameters=parameters,
    376                                    output=output, callables=callables, build_gradients=False, build_hessians=False,
--> 377                                    verbose=kwargs.pop('verbose', False))
    378     str_statevar_dict = collections.OrderedDict((str(key), unpack_condition(value)) \
    379                                                 for (key, value) in statevar_dict.items())

~/opt/pycalphad/pycalphad/codegen/callables.py in build_phase_records(dbf, comps, phases, conds, models, output, callables, parameters, verbose, build_gradients, build_hessians)
    202                                     additional_statevars=state_variables,
    203                                     build_gradients=build_gradients,
--> 204                                     build_hessians=build_hessians)
    205 
    206     # If a vector of parameters is specified, only pass the first row to the PhaseRecord

~/opt/pycalphad/pycalphad/codegen/callables.py in build_callables(dbf, comps, phases, models, parameter_symbols, output, build_gradients, build_hessians, additional_statevars)
    113         out = out.xreplace(dict(zip(undefs, undef_vals)))
    114         build_output = build_functions(out, tuple(state_variables + site_fracs), parameters=parameter_symbols,
--> 115                                        include_grad=build_gradients, include_hess=build_hessians)
    116         cf, gf, hf = build_output.func, build_output.grad, build_output.hess
    117         _callables['callables'][name] = cf

~/opt/pycalphad/pycalphad/core/cache.py in wrapper(*args, **kwds)
    139                         stats[HITS] += 1
    140                         return result
--> 141                 result = user_function(*args, **kwds)
    142                 with lock:
    143                     root, = nonlocal_root

~/opt/pycalphad/pycalphad/codegen/sympydiff_utils.py in build_functions(sympy_graph, variables, parameters, wrt, include_obj, include_grad, include_hess, func_options, grad_options, hess_options)
    112     inp = sympify(variables + parameters)
    113     graph = sympify(sympy_graph).xreplace({zoo: oo})
--> 114     func = lambdify(inp, [graph], **_get_lambidfy_options(func_options))
    115     if include_grad or include_hess:
    116         grad_graphs = list(graph.diff(w).xreplace({zoo: oo}) for w in wrt)

~/opt/miniconda3/lib/python3.7/site-packages/symengine/__init__.py in lambdify(args, exprs, **kwargs)
     33 
     34     def lambdify(args, exprs, **kwargs):
---> 35         return Lambdify(args, *exprs, **kwargs)
     36 
     37 

symengine_wrapper.pyx in symengine.lib.symengine_wrapper.Lambdify()

symengine_wrapper.pyx in symengine.lib.symengine_wrapper._Lambdify.__init__()

symengine_wrapper.pyx in symengine.lib.symengine_wrapper.LLVMDouble._init()

RuntimeError: Not implemented._

Am I missing something?

Thank you
EA

bocklund:
I don't intend to finish this PR, so I'm closing it for now. Perhaps some of the ideas here can be worked into another set of changes later.

PR #269, opened 2020-03-30
FIX: Improved database reading for out of order type definitions 

bocklund:
`TYPE_DEFINITION` lines in pycalphad needed to be defined prior to the `PHASE` keyword that the type definition belonged to, otherwise the type definition would not be applied when `_process_phase` was called. This PR puts all the added type definitions in a queue so they can be processed after all phases are processed, which decouples processing of type definitions from the the order that they are defined in the database. This has the following consequences

* `_process_typedef` now adds the `model_hints` to each phase directly, rather than `_process_phase` building model hints from the existing type definitions
* Two warnings are added. One warning if a type definition is defined in the `PHASE` keyword, but has no corresponding type definition. This can occur from a typo, or if the `PHASE` line was copied and the corresponding type definition left behind. Another warning is raised when there are type definitions defined, but not used by any phase, which could happen for similar reasons.
* `model_hints` are now added to both the ordered and disordered phases in the order-disorder partitioned model. This fixes a bug where `filter_phases` would not correctly filter out phases that did not have `model_hints` defined. This means that when writing TDBs, the type definition for the partitioned model will be applied to both phases*.
* A test is added that verifies that no matter the order of the ordered phase, disordered phase, and type definition, the order-disorder `model_hints` are applied to both phases. In this test, the type definition is applied only to the ordered phase, since that seems to be the predominant way that TDBs are found in the literature. Other variants for this test are possible, such as the type definition being applied to only the disordered phase or to both phases. All should produce the same result.
* Another test is added for a database in the test suite that `filter_phases` would fail to remove the disordered phase from.
* The `test_eq_ternary_inside_mass` test was failing. At first this seemed spurious, but after a closer look, I found that the type definitions for magnetism were after the definition of the BCC_A2 phase the magnetic `model_hints` was not getting added to the BCC_A2 phase. I added a test to check that these model hints are correctly added (it fails on develop, but passes here). This helps correctness. We weren't seeing this issue with the order-disorder type definition because the implementation of `_process_typedef` always updated the ordered phase `model_hints` explicitly if the ordered phase was in the database. I updated the Gibbs energies and chemical potentials of the `test_eq_ternary_inside_mass` to what are presumably the correct values now that magnetism is considered.
* A call to `filter_phases` is added to `binplot` to remove phases that cannot exist. Required because we don't directly call `equilibrium`, which would normally do this for us.

*The way the TDBs are read by Thermo-Calc seems to be that one phase is specified, then the type definition, then the other phase in the partitioned model, and it doesn't matter which one is first. Thermo-Calc will give a warning if the type definition line is run before both `PHASE` lines are run, but it seems to produce the correct phases and internal understanding.


pycalphad/io/tdb.py
```diff
@@ -271,37 +271,55 @@ def _tdb_grammar(): #pylint: disable=R0914
...
     if len(keyword) == 0:
-        raise ValueError('Unknown keyword: {}'.format(tokens[3]))
+        raise ValueError('Unknown type definition keyword: {}'.format(tokens[3]))
+    if len(matching_phases) == 0:
+        warnings.warn(f"The type definition character `{typechar}` in `TYPE_DEFINITION {typechar} {line}` is not used by any phase.")
     if keyword == 'MAGNETIC':
-        # magnetic model (IHJ model assumed by default)
-        targetdb.tdbtypedefs[typechar] = {
-            'ihj_magnetic':[float(tokens[4]), float(tokens[5])]
+        # Magnetic model (IHJ model assumed by default)
```
richardotis: IHJ is not assumed here (the comment predates the implementation of the Xiong model). The `ihj_` model hints are a bit of a misnomer, since we use `ihj_magnetic_afm_factor == 0` as a check for turning the Xiong model on in `Model`. I think removing the comment is ok; possible renaming of the magnetic model hints is out of scope for this PR.

pycalphad/tests/test_database.py
```diff
@@ -558,3 +558,105 @@ def test_database_parameter_with_species_that_is_not_a_stoichiometric_formula():
...
+    PHASE_B2 = """
+     PHASE BCC_B2  %C  3 .5 .5    3 !
+     CONST BCC_B2  :AL,NI : AL,NI : VA: !
+    """
+
+    TYPE_DEF_ORD = """
+     TYPE_DEFINITION C GES A_P_D BCC_B2 DIS_PART BCC_A2 !
+    """
+    import itertools
+    for (k1, v1), (k2, v2), (k3, v3) in itertools.permutations([('PHASE_A2 ', PHASE_A2), ('PHASE_B2 ', PHASE_B2), ('TYPE_DEF ', TYPE_DEF_ORD)]):
```
richardotis: This is a really cool use of test case generation.

pycalphad/tests/test_equilibrium.py
```diff
@@ -209,8 +209,8 @@ def test_eq_ternary_inside_mass():
     eq = equilibrium(ALCOCRNI_DBF, ['AL', 'CO', 'CR', 'VA'], ['L12_FCC', 'BCC_B2', 'LIQUID'],
                      {v.T: 1523, v.X('AL'): 0.44455555555555554,
                       v.X('CO'): 0.22277777777777777, v.P: 101325}, verbose=True)
-    assert_allclose(eq.GM.values, -105871.20, atol=0.1)
-    assert_allclose(eq.MU.values.flatten(), [-104655.532294, -142591.644379,  -82905.085459], atol=0.1)
+    assert_allclose(eq.GM.values, -105871.52, atol=0.1)
+    assert_allclose(eq.MU.values.flatten(), [-104653.814416, -142595.497201,  -82905.765934], atol=0.1)
```
richardotis: Can you check these in TC?

pycalphad/tests/test_database.py
```diff
@@ -461,8 +461,8 @@ def test_comma_templims():
      TYPE-DEF % SEQ * !
      TYPE-DEF A GES AMEND_PHASE_DESCRIPTION @ MAGNETIC -3 0.28 !
      TYPE-DEF B GES AMEND_PHASE_DESCRIPTION @ MAGNETIC -1 0.4 !
-     TYPE-DEF O GES AMEND_PHASE_DESCRIPTION B2_BCC DIS_PART A2_BCC !
-     TYPE-DEF Y GES AMEND_PHASE_DESCRIPTION FCC_4SL DIS_PART A1_FCC !
+    $ TYPE-DEF O GES AMEND_PHASE_DESCRIPTION B2_BCC DIS_PART A2_BCC !
+    $ TYPE-DEF Y GES AMEND_PHASE_DESCRIPTION FCC_4SL DIS_PART A1_FCC !
```
richardotis: Why are we commenting these out again?

pycalphad/io/tdb.py
bocklund: I changed the existing comment to clarify that the `ihj_magnetic_` prefix is legacy and it's used for both Xiong and IHJ models.

pycalphad/tests/test_database.py
bocklund: Thanks!

pycalphad/tests/test_equilibrium.py
bocklund: I followed the comment that they could not be checked before, but I double checked following your comment and the energy and chemical potentials are in agreement with the new values that we found (to less than 0.1 J). Updated the test and removed the comment.

pycalphad/tests/test_database.py
bocklund: The test is defined in and explicitly used for the `test_comma_templims` test. If the FCC_4SL type definition were defined, it would raise errors because this test database is truncated and does not define the FCC_4SL phase. I added an additional comment to clarify why the fcc type definition is commented.

PR #270, opened 2020-04-06
FIX: eqcalculate: Calculate degrees of freedom based on Model

richardotis:
When the Phase is defined in a Database with many more components than are defined
to be active in the present calculation, `_eqcalculate()` would incorrectly calculate degrees of freedom based on all possible components, instead of the active ones.

This would then lead to errors inside `calculate()`, when the length of the passed `points` array would exceed the `maximum_internal_dof`.

This fix does two things:
1. Calculates degrees of freedom using the already-passed `models` object, which is based on the active components.
2. Fixes `unpack_phases` to return only the active phases, instead of returning nothing and causing a fallback to all phases in the Database. (This wasn't a correctness issue, but it did cause unnecessary computation.)

### DIFF
--- a/pycalphad/core/equilibrium.py
+++ b/pycalphad/core/equilibrium.py
@@ -31,8 +31,8 @@ def _adjust_conditions(conds):
     return new_conds
 
 
-def _eqcalculate(dbf, comps, phases, conditions, output, data=None, per_phase=False, callables=None, parameters=None,
-                 **kwargs):
+def _eqcalculate(dbf, comps, phases, conditions, output, data=None, per_phase=False, callables=None, models=None,
+                 parameters=None, **kwargs):
     """
     WARNING: API/calling convention not finalized.
     Compute the *equilibrium value* of a property.
@@ -66,10 +66,12 @@ def _eqcalculate(dbf, comps, phases, conditions, output, data=None, per_phase=Fa
     per_phase : bool, optional
         If True, compute and return the property for each phase present.
         If False, return the total system value, weighted by the phase fractions.
-    parameters : dict, optional
-        Maps SymPy Symbol to numbers, for overriding the values of parameters in the Database.
     callables : dict
         Callable functions to compute 'output' for each phase.
+    models : a dict of phase names to Model
+        Model class to use for each phase.
+    parameters : dict, optional
+        Maps SymPy Symbol to numbers, for overriding the values of parameters in the Database.
     kwargs
         Passed to `calculate`.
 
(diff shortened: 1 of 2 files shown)


pycalphad/core/equilibrium.py
```diff
@@ -79,7 +81,7 @@ def _eqcalculate(dbf, comps, phases, conditions, output, data=None, per_phase=Fa
     """
     if data is None:
         data = equilibrium(dbf, comps, phases, conditions, to_xarray=False)
-    active_phases = unpack_phases(phases) or sorted(dbf.phases.keys())
+    active_phases = unpack_phases(phases)
```
bocklund, line 84: Should this also be passed through `filter_phases`?

pycalphad/core/equilibrium.py
```diff
@@ -101,7 +103,7 @@ def _eqcalculate(dbf, comps, phases, conditions, output, data=None, per_phase=Fa
     # For each phase select all conditions where that phase exists
     # Perform the appropriate calculation and then write the result back
     for phase in active_phases:
-        dof = sum([len(x) for x in dbf.phases[phase].constituents])
+        dof = len(models[phase].site_fractions)
```
bocklund, line 106: This makes `models` a required argument, right?

pycalphad/core/equilibrium.py
```diff
@@ -79,7 +81,7 @@ def _eqcalculate(dbf, comps, phases, conditions, output, data=None, per_phase=Fa
     """
     if data is None:
         data = equilibrium(dbf, comps, phases, conditions, to_xarray=False)
```
bocklund, line 83: We don't call `_eqcalculate` without data internally. Should we support this? If we want to keep this, we need to pass through the arguments like `models` and `parameters`.

pycalphad/core/equilibrium.py
richardotis: This is already taken care of by the caller. Including extra phases won't affect the result here; in fact, the current implementation (before this fix) is falling back to setting this equal to _all_ the phases in the Database.

pycalphad/core/equilibrium.py
richardotis: Yes.

pycalphad/core/equilibrium.py
bocklund: If we assume that filtering is done by the caller, should this also assume that the phases are in canonical form and don't require unpacking?

richardotis: Right now a dict is what's passed by the caller:
`active_phases = {name: dbf.phases[name] for name in filter_phases(dbf, comps, phases)}`
I think it's more robust to leave it as-is.

pycalphad/core/equilibrium.py
```diff
@@ -113,7 +115,7 @@ def _eqcalculate(dbf, comps, phases, conditions, output, data=None, per_phase=Fa
         if statevars.get('mode', None) is None:
             statevars['mode'] = 'numpy'
```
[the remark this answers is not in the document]
richardotis: I think changing that would force altering some old tests too, I am going to leave it for now.

PR #273, opened 2020-04-17
FIX: Two sublattice ionic liquid fixes and tests

bocklund:
This PR fixes two bugs in the implementation of ionic liquid two sublattice model

1. Neutral species in the anion sublattice can be specified in the TDB using one sublattice parameters. These are special cased now in `Model` to be valid parameters and to correct their sublattice index in the Redlich-Kister sum to produce the correct `v.SiteFraction` objects.
2. A sorting bug was fixed in the anion lattice, where the model from Sundman [1] describes the anion sublattice being sorted by first charged species, then vacancies, then anion species. Within each species group, the species are sorted alphabetically as usual. This is changed within Model itself, rather than the TDB which sorts the parameters at parse time. Because it's specific to this implementation of the Model, I think it makes sense to put it in the `Model` class corresponding to any phase with the `ionic_liquid_2SL` model hint applied.

This PR includes 
* tests that compare the energy to the energies computed by the same database in Thermo-Calc
* some documentation improvements and cleanup for the ionic liquid

[1] Sundman, "Modification of the two-sublattice model for liquids", Calphad 15(2) (1991) 109-119 https://doi.org/d3jppb


pycalphad/io/tdb.py
```diff
@@ -238,7 +238,8 @@ def _tdb_grammar(): #pylint: disable=R0914
     # PHASE
     cmd_phase = TCCommand('PHASE') + symbol_name + \
         Suppress(White()) + CharsNotIn(' !', min=1) + Suppress(White()) + \
-        Suppress(int_number) + Group(OneOrMore(float_number)) + LineEnd()
+        Suppress(int_number) + Group(OneOrMore(float_number)) + \
```
richardotis, line 241: I'm okay to leave this in but it should be documented separately that this commit is for MatCalc compatibility.

bocklund: I hadn’t intended to sneak this in here, that commit got swept in by accident. 

I can leave it in or take it out. However, I do want to note that this is actually a documented feature for the PHASE keyword in the Thermo-Calc documentation. Anything after the sublattice numbers is an “auxiliary text string”

richardotis: Leaving it is fine.

PR #292, opened 2020-12-11
Permit color and legend customization

### LINKED ISSUE #291: Enable color and legend customization in eqplot
Make it easier to customize the colors of tielines and tie triangles. Also enable custom legends for phases.

dschwen:
Closes #291

### DIFF
--- a/pycalphad/plot/binary/plot.py
+++ b/pycalphad/plot/binary/plot.py
@@ -11,7 +11,7 @@ from pycalphad.plot.utils import phase_legend
 from .map import map_binary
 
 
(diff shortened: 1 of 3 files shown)


dschwen:
Ok, I see, I need to import phase_legend. Is that the best design I can do, repeatedly specifying the same default value at all levels?

bocklund:
Thanks @dschwen! These changes go a long way in allowing users to customize the look and feel of their phase diagrams. I made a few smalls changes to the docstrings to better match how matplotlib describes their colors API (e.g. you could use this to provide any valid MPL color as a tieline_color, not just RGBA values) and I made the `tieline_color` arguments immutable tuples instead of mutable lists. 

I'm not worried about the codecov/patch check failing. Our test suite just has some small integration tests over the plotting API, since it's more difficult to test the correctness of the graphics produced.

bocklund (approved):

PR #295, opened 2020-12-29
FIX: Improve Triangular axes projection

bocklund:
This PR houses a few changes, all in the `TriangluarAxes` projection class.

1. simplification of the transformations code
2. stop calling classmethods of `Axes` and passing self
3. fix several methods, such as `get_xaxis_text1_transform`, to call the superclass and get the transform *after* padding. This means setting the padding of the ticks and axis labels now works correctly for both axes.
4. add an override to `set_ylabel` that sets a default rotation of 60 degrees

### DIFF
--- a/pycalphad/plot/eqplot.py
+++ b/pycalphad/plot/eqplot.py
@@ -112,8 +112,6 @@ def eqplot(eq, ax=None, x=None, y=None, z=None, tielines=True, tieline_color=(0,
     elif projection == 'triangular':
         x = indep_comps[0] if x is None else x
         y = indep_comps[1] if y is None else y
-        # plot settings
-        ax.yaxis.label.set_rotation(60)
         # Here we adjust the x coordinate of the ylabel.
         # We make it reasonably comparable to the position of the xlabel from the xaxis
         # As the figure size gets very large, the label approaches ~0.55 on the yaxis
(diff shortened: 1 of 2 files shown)


bocklund:
There's still some outstanding "like to have" items, but I don't think I'll address them in this PR, mainly:

* Tune the padding/position of the ylabel better automatically. Currently the position is at (0, 0.5) of the untransformed axes coordinates, but it would probably make sense for the position to be transformed. I think it should be transformed using the transforms already defined (like the ticks), but for some reason it is not.
* Tune the number of ticks/labels. Currently the ticks are hardcoded to be in [0, 1] in 0.1 increments. Without this hardcoded, there are two issues. First, the x- and y-axes may have different divisions, which looks awkward. Second, the divisions tend to only go down to increments of 0.2 at reasonable label sizes. This is caused by how the automatic tick numbering in matplotlib works. Two other classes would need to be overriden: first the `XAxis`/`YAxis` classes to modify how `get_tick_space` computes how many tick spaces are available (some tuning would need to be done to balance the number of ticks and the spacing of the labels), and second the `MaxNLocator` class would need to be overridden to increase the number of ticks that can form in automatic mode from 9 to >=10 (to allow increments of 0.1).

richardotis (approved):

PR #308, opened 2021-03-01
FIX: Improve strictness when parsing TDB FUNCTION and PARAMETER lines

### LINKED ISSUE #307: unexpected behavior of G^ref when temperature range is set?
Hi all,

When I try to calculate the Gibbs energy of the bcc phase, the Gref seems to not include the Fe end-member if I use parameters with different range of temperature (please see in the tdb file and the commented parameter). 

That is:
if I include G(BCC,Fe:H,0) with different temperature ranges (uncommented parameter), the term representing G(BCC,Fe:VA,0), in this case GHSERFE, does not appear in the final  GM(BCC) function. Is there something I'm not seeing? 

Attached you may find the tdb for running this question using the script below:

_from pycalphad import Model
dbf = Database('bcc.tdb')
comps = ['FE','H','VA']
dbf.symbols = {}
bcc_mod = Model(dbf, ['FE','H','VA'], 'BCC')
bcc_mod.GM_

tdb:
[bcc.txt](https://github.com/pycalphad/pycalphad/files/6061713/bcc.txt)

bocklund:
Checklist
  * [x] `setup.py`
  * [x] `environment-dev.yml`

This PR makes the TDB parser more strict by requiring that PARAMETER and FUNCTION lines are terminated by an optional-suppressed reference key and a line end. This fixes behavior where the parser would silently consume all TDB content until the next line terminator (`!`) for these two commands once the `func_expr` grammar was successfully parsed.

- `FUNCTION` and `PARAMETER` pyparsing grammars are updated to include through the line ending, if the parsed function expression and reference key is not following by a line ending, a `ParseException` is now raised.
- A failing test based on #307 was created and merging this will close #307
- Before this change, I had 522 databases that were known to be parsed successfully pycalphad. After making the change, all 522 still parsed successfully. The only TDB that did not parse correctly was the ALNIPT_TDB database in our test suite, which had references of `REF: 0`, which I changed to `REF0` (I also verified that Thermo-Calc 2020b says that `REF: 0` is an invalid reference format in GES6)


bocklund:
The failing Ubuntu Python 3.9 test is caused by an error uploading coverage to codecov. The test suite itself passed.

richardotis (approved):

PR #311, opened 2021-03-09
FIX: Fix assumptions for substitutional/interstitial sublattices in partitioned order/disorder phases

### LINKED ISSUE #310: Bug in order-disorder models when species other than VA are present in the "vacancy sublattice"
Below I have a complete, runnable example that demonstrates an error in handling order-disorder phases with non-vacancy constituents in the vacancy sublattice.

Given a binary A-B subsystem of a hypothetical A-B-C system, we can define a couple different sublattice models for the ordered phase (named `ORDERED`):

1. `(A,B,VA)(A,B,VA)(VA)` (suffixed in the test with `_VA_VA`, since there are vacancies in the disordered sublattice and "vacancy" sublattice)
2. `(A,B,VA)(A,B,VA)(C,VA)`(suffixed in the test with `_VA_VA_C `, since there are vacancies in the disordered sublattice and vacancies and C in the "vacancy" sublattice)
3. `(A,B)(A,B)(VA)` (suffixed in the test with `_VA_C `, since there are vacancies and C in the "vacancy" sublattice)

Now we can define a binary A-B model (`Model(db_{suffix}, ["A", "B", "VA"], "ORDERED")`) for each system and check the energy of pure A. 

The test shows that `_VA_VA` and `_VA_C` produce the correct energy for pure A (-10 kJ in this case), but the `_VA_VA_C` produces an incorrect energy (0 kJ in this case). 

There are two issues at play here:

1. Even though component `C` is not active in the model, the answer changes. This is because [this line](https://github.com/pycalphad/pycalphad/blob/be404461ad44a4677897a0cc0d843dd7bec1007e/pycalphad/model.py#L982) is identifying the vacancy sublattice by the Phase object's (active and inactive) constituents, rather than the Model object's active constituents. It might be necessary to look at the phase model so that we don't mis-identify a sublattice as a vacancy sublattice, but I think we could use a better heuristic to determine the vacancy sublattice and eliminate this as a potential bug.
2. [This `if` statement](https://github.com/pycalphad/pycalphad/blob/be404461ad44a4677897a0cc0d843dd7bec1007e/pycalphad/model.py#L985) incorrectly classifies the last sublattice as "not a vacancy sublattice"  in `_VA_C` and `_VA_VA_C` cases, since the vacancy sublattice of the Phase constituents contains more than one species. That leads to the `Model.mole_fractions` of species in this sublattice to be substituted for site fractions, which is incorrect. The incorrect behavior only manifests if `VA` is present in other sublattices than the vacancy sublattice (i.e. the `_VA_VA_C` case). The `_VA_C` gives the correct answer because for the wrong reason because `Model.mole_fractions` for `VA` in the vacancy sublattice reduces to `1`.  

The behavior is really subtle and only occurs if the vacancy sublattice has multiple constituents (active and inactive constituents, since we use `Phase.constituents`) AND one or more of those constituents shows up in the other sublattices.

```python
from pycalphad import Database, Model, variables as v
from pycalphad.tests.test_energy import check_energy

def test_order_disorder_interstitial_sublattice():
    TDB_OrderDisorder_VA_VA = """
    ELEMENT VA   VACUUM   0.0000E+00  0.0000E+00  0.0000E+00 !
    ELEMENT A    DISORD     0.0000E+00  0.0000E+00  0.0000E+00 !
    ELEMENT B    DISORD     0.0000E+00  0.0000E+00  0.0000E+00 !
    ELEMENT C    DISORD     0.0000E+00  0.0000E+00  0.0000E+00 !

    DEFINE_SYSTEM_DEFAULT ELEMENT 2 !
    DEFAULT_COMMAND DEF_SYS_ELEMENT VA !

    TYPE_DEFINITION % SEQ *!
    TYPE_DEFINITION & GES A_P_D DISORD MAGNETIC  -1.0    4.00000E-01 !
    TYPE_DEFINITION ( GES A_P_D ORDERED MAGNETIC  -1.0    4.00000E-01 !
    TYPE_DEFINITION ' GES A_P_D ORDERED DIS_PART DISORD ,,,!

    PHASE DISORD  %&  2 1   3 !
    PHASE ORDERED %('  3 0.5  0.5  3  !

    CONSTITUENT DISORD  : A,B,VA : VA :  !
    CONSTITUENT ORDERED  : A,B,VA : A,B,VA : VA :  !

    PARAMETER G(DISORD,A:VA;0)  298.15  -10000; 6000 N !
    PARAMETER G(DISORD,B:VA;0)  298.15  -10000; 6000 N !

    """

    TDB_OrderDisorder_VA_VA_C = """
    $ Compared to TDB_OrderDisorder_VA_VA, this database only
    $ differs in the fact that the VA sublattice contains C

    ELEMENT VA   VACUUM   0.0000E+00  0.0000E+00  0.0000E+00 !
    ELEMENT A    DISORD     0.0000E+00  0.0000E+00  0.0000E+00 !
    ELEMENT B    DISORD     0.0000E+00  0.0000E+00  0.0000E+00 !
    ELEMENT C    DISORD     0.0000E+00  0.0000E+00  0.0000E+00 !

    DEFINE_SYSTEM_DEFAULT ELEMENT 2 !
    DEFAULT_COMMAND DEF_SYS_ELEMENT VA !

    TYPE_DEFINITION % SEQ *!
    TYPE_DEFINITION & GES A_P_D DISORD MAGNETIC  -1.0    4.00000E-01 !
    TYPE_DEFINITION ( GES A_P_D ORDERED MAGNETIC  -1.0    4.00000E-01 !
    TYPE_DEFINITION ' GES A_P_D ORDERED DIS_PART DISORD ,,,!

    PHASE DISORD  %&  2 1   3 !
    PHASE ORDERED %('  3 0.5  0.5  3  !

    CONSTITUENT DISORD  : A,B,VA : C,VA :  !
    CONSTITUENT ORDERED  : A,B,VA : A,B,VA : C,VA :  !

    PARAMETER G(DISORD,A:VA;0)  298.15  -10000; 6000 N !
    PARAMETER G(DISORD,B:VA;0)  298.15  -10000; 6000 N !

    """

    TDB_OrderDisorder_VA_C = """
    $ Compared to TDB_OrderDisorder_VA_VA_C, this database only
    $ differs in the fact that the disorderd sublattices do not contain VA

    ELEMENT VA   VACUUM   0.0000E+00  0.0000E+00  0.0000E+00 !
    ELEMENT A    DISORD     0.0000E+00  0.0000E+00  0.0000E+00 !
    ELEMENT B    DISORD     0.0000E+00  0.0000E+00  0.0000E+00 !
    ELEMENT C    DISORD     0.0000E+00  0.0000E+00  0.0000E+00 !

    DEFINE_SYSTEM_DEFAULT ELEMENT 2 !
    DEFAULT_COMMAND DEF_SYS_ELEMENT VA !

    TYPE_DEFINITION % SEQ *!
    TYPE_DEFINITION & GES A_P_D DISORD MAGNETIC  -1.0    4.00000E-01 !
    TYPE_DEFINITION ( GES A_P_D ORDERED MAGNETIC  -1.0    4.00000E-01 !
    TYPE_DEFINITION ' GES A_P_D ORDERED DIS_PART DISORD ,,,!

    PHASE DISORD  %&  2 1   3 !
    PHASE ORDERED %('  3 0.5  0.5  3  !

    CONSTITUENT DISORD  : A,B : C,VA :  !
    CONSTITUENT ORDERED  : A,B : A,B : C,VA :  !

    PARAMETER G(DISORD,A:VA;0)  298.15  -10000; 6000 N !
    PARAMETER G(DISORD,B:VA;0)  298.15  -10000; 6000 N !

    """

    db_VA_VA = Database(TDB_OrderDisorder_VA_VA)
    db_VA_C = Database(TDB_OrderDisorder_VA_C)
    db_VA_VA_C = Database(TDB_OrderDisorder_VA_VA_C)

    mod_VA_VA = Model(db_VA_VA, ["A", "B", "VA"], "ORDERED")
    mod_VA_C = Model(db_VA_C, ["A", "B", "VA"], "ORDERED")
    mod_VA_VA_C = Model(db_VA_VA_C, ["A", "B", "VA"], "ORDERED")

    # Site fractions for pure A
    subs_dict = {
        v.Y('ORDERED', 0, v.Species('A')): 1.0,
        v.Y('ORDERED', 0, v.Species('B')): 0.0,
        v.Y('ORDERED', 0, v.Species('VA')): 0.0,
        v.Y('ORDERED', 1, v.Species('A')): 1.0,
        v.Y('ORDERED', 1, v.Species('B')): 0.0,
        v.Y('ORDERED', 1, v.Species('VA')): 0.0,
        v.Y('ORDERED', 2, v.Species('VA')): 1.0,
        v.T: 300.0,
    }

    check_energy(mod_VA_VA, subs_dict, -10000, mode='sympy')
    check_energy(mod_VA_C, subs_dict, -10000, mode='sympy')
    check_energy(mod_VA_VA_C, subs_dict, -10000, mode='sympy')


test_order_disorder_interstitial_sublattice()
```

Thanks to @Ebert-Daniel for pointing this out to me via email.

bocklund:
Checklist
  * [x] `setup.py`
  * [x] `environment-dev.yml`

This PR aims to close #310 by changing the heuristic for how pycalphad considers a sublattice in partitioned order/disorder and lays out in the code documentation how we treat the energetic contributions. Previously, the implementation was limited because it:

1. only considered one interstitial sublattice
2. that sublattice had to be the last sublattice in the ordered and disordered phase constituents
3. the sublattice could only contain vacancies

To summarize the new changes that frees all these assumptions (copied from the new documentation):

> This method assumes that the first sublattice of the disordered phase is the substitutional sublattice and all other sublattices are interstitial. In the ordered phase, all sublattices with constituents that match the disordered substitutional sublattice will be treated as disordered (with site fractions replaced by mole fractions in the ordered sublattices) and the interstitial sublattices will not have any site fractions substituted.

The only assumptions here is that there is one set of sublattices in the ordered phase that can disorder. Any number of sublattices may disorder and any number of interstitial sublattices are possible.

The main changes made in this PR:

1. Refactor the `Model.atomic_ordering_energy` method to clearly split up the two responsibilities of the method: computing the ordering energy and replacing the ordered energy contributions with the disordered energy contributions
2. Fix test case for #310 and a more complicated case where one of the substitutional elements is also an interstitial element
2. Revamped documentation for `Model.atomic_ordering_energy`
3. Make `Model.mole_fraction` a private method (`Model._mole_fraction`) to better reflect its purpose and to discourage users from using it, since it is not applicable outside the order-disorder model. It was not applicable outside the `atomic_ordering_energy` before, since it counts vacancies as regular species. This behavior is made more explicit by making `Model._mole_fraction` private. Since its usage is limited to `Model.atomic_ordering_energy`, I added a parameter to take in the substitutional sublattice indices. Strictly speaking, this could break backwards compatibility to anyone depending on `mole_fraction`, but `mole_fraction` does not normalize to mole fraction of atoms since it counts vacancies as atoms and anyone using it for the purpose of computing the mole fraction of atoms should be using `Model.moles` instead.


pycalphad/model.py
```diff
@@ -986,23 +997,13 @@ def atomic_ordering_energy(self, dbe):
...
-        which are not substituted (i.e. they do not contribute to the ordering
-        energy). This formulation makes pycalphad conceptually consistent with
-        the implementation in OpenCalphad.
+        This method assumes that the first sublattice of the disordered phase is
+        the substitutional sublattice and all other sublattices are
+        interstitial. In the ordered phase, all sublattices with constituents
+        that match the disordered substitutional sublattice will be treated as
+        disordered (with site fractions replaced by mole fractions in the
+        ordered sublattices) and the interstitial sublattices will not have any
+        site fractions substituted.
```
Ebert-Daniel, line 1006: I don't know if I got it correctly, but for the ord-dis model, we should take the overall mole fraction and substitute in the ordered site fractions to get the ordering contribution, only if the phase can be totally disordered. But in the case of 
A2: (A,B,Va)m+n(C,Va)p
B2: (A,B,Va)m(A,B,Va)n(C,Va)p
The substitution shouldn't be done with mole fraction but with the sub-lattice fraction of the disordered state that represents the disordered state of the sub-lattice sets that undergo disordering of the ordered state. Is this what you mean in this phrase, am I missing something? e. g.:

Gm = Gm_dis + ΔGm_ord 
ΔGm_ord = Gm_ord (y1i, y2j, y3k) -  Gm_ord (y1i = y2j, y3k)

One reference that used this model: 
Acta Materialia 79 (2014) 1–15;
doi: http://dx.doi.org/10.1016/j.actamat.2014.07.006

bocklund: > The substitution shouldn't be done with mole fraction but with the sub-lattice fraction of the disordered state that represents the disordered state of the sub-lattice sets that undergo disordering of the ordered state.

Thank you for the reference. Most of the original publications for the partitioned order/disorder model do not include the interstitial sublattices and leave that out of their derivations. I actually had trouble tracking down a reference for how the interstitial sublattices are intended to work, so thank you for the recent reference.

### TL;DR

**I believe this formulation is the same as the one we are using, just with slightly different notation.** Please make any suggestions that would help clarify the documentation!

### Discussion

That Equation (11) from the work of Phan, Paek, and Kang, Acta Materialia 79 (2014) 1–15 that you cited:

Screen Shot 2021-03-10 at 11 57 25 AM

treats the disordered sublattices the same as the one in Equation (3) of Servant and Ansara, Calphad 25, (2001) 509–525 that is cited in the documentation

Screen Shot 2021-03-10 at 12 02 27 PM

except that the work of Phan, Paek and Kang also specify that the interstitial sublattice should not be substituted (which is left out of Servant and Ansara's derivation, since they did not use an interstitial sublattice).

What the code here intends to do is as follows (you'll have to excuse the fast and loose notation here, since in pycalphad we don't make any assumptions about the number of sublattices so the notation from Phan, Paek and Kang which specifies the sublattices explicitly doesn't translate well):

image

where the "subs" superscript means "on substitutional sublattices" and the "inter" superscript means "on interstitial sublattices".  The "mole fraction" from `Model._mole_fraction` that is used here only is really more of a "quasi mole fraction" rather than a real "mole fraction". It's named in the spirit of the original derivations. `Model._mole_fraction` replaces the site fractions on substitutional sublattices and does not normalize the mole fraction for vacancy site fractions as a true "mole fraction" would. This is the motivation for making the `_mole_fraction` method private: because it's not a real mole fraction.

bocklund: For another brief description in the published literature, see section 3.3 of Connectable _et al._ Calphad 32 (2008) 361-370 [doi: 10.1016/j.calphad.2008.01.002](https://doi.org/10.1016/j.calphad.2008.01.002), specifically equations 14-18.

Ebert-Daniel: Thank you @bocklund. This was really helpful. I think I could understand it better!

bocklund:
One of the TDB files I had uses the Al-C-Fe system from Connectable _et al._ as a subsystem. I wrote that out to a TDB file. 
[al-c-fe.tdb.txt](https://github.com/pycalphad/pycalphad/files/6124007/al-c-fe.tdb.txt)


This is my test script, with the energies from Thermo-Calc (site fractions were plugged in via `set-start-constitution`)

```python
# coding: utf-8
import pycalphad
print(pycalphad.__version__)
from pycalphad import Database, equilibrium, variables as v, calculate, Model
dbf = Database('al-c-fe.tdb')
phases = ['B2_BCC']

# Al-C
comps = ['AL', 'C', 'VA']
cr = calculate(dbf, comps, phases, N=1, P=101325, T=300, pdens=10)
mod = Model(dbf, comps, 'B2_BCC')
print(sorted(comps))
print([sorted([sp.name for sp in subl]) for subl in mod.constituents])
print(cr.isel(points=-1).Y.values.squeeze())
print(cr.isel(points=-1).GM.values.squeeze(), 'TC: 31911.163')

# Al-Fe
comps = ['AL', 'FE', 'VA']
cr = calculate(dbf, comps, phases, N=1, P=101325, T=300, pdens=10)
mod = Model(dbf, comps, 'B2_BCC')
print(sorted(comps))
print([sorted([sp.name for sp in subl]) for subl in mod.constituents])
print(cr.isel(points=-1).Y.values.squeeze())
print(cr.isel(points=-1).GM.values.squeeze(), 'TC: 86629.399')

# C-Fe
comps = ['FE', 'C', 'VA']
cr = calculate(dbf, comps, phases, N=1, P=101325, T=300, pdens=10)
mod = Model(dbf, comps, 'B2_BCC')
print(sorted(comps))
print([sorted([sp.name for sp in subl]) for subl in mod.constituents])
print(cr.isel(points=-1).Y.values.squeeze())
print(cr.isel(points=-1).GM.values.squeeze(), 'TC: 48974.223')

# Al-C-Fe
comps = ['AL', 'FE', 'C', 'VA']
cr = calculate(dbf, comps, phases, N=1, P=101325, T=300, pdens=10)
mod = Model(dbf, comps, 'B2_BCC')
print(sorted(comps))
print([sorted([sp.name for sp in subl]) for subl in mod.constituents])
print(cr.isel(points=-1).Y.values.squeeze())
print(cr.isel(points=-1).GM.values.squeeze(), 'TC: 34659.484')

```


with outputs:

```
0.8.4+26.gea5c234a
['AL', 'C', 'VA']
[['AL', 'VA'], ['AL', 'VA'], ['C', 'VA']]
[0.35136561 0.64863439 0.79724251 0.20275749 0.44382726 0.55617274]
31911.16739624921 TC: 31911.163
['AL', 'FE', 'VA']
[['AL', 'FE', 'VA'], ['AL', 'FE', 'VA'], ['VA']]
[0.46828033 0.00228176 0.52943791 0.43587845 0.01705016 0.54707139
 1.        ]
86629.40328004515 TC: 86629.399
['C', 'FE', 'VA']
[['FE', 'VA'], ['FE', 'VA'], ['C', 'VA']]
[0.35136561 0.64863439 0.79724251 0.20275749 0.44382726 0.55617274]
48974.22829354314 TC: 48974.223
['AL', 'C', 'FE', 'VA']
[['AL', 'FE', 'VA'], ['AL', 'FE', 'VA'], ['C', 'VA']]
[0.23632422 0.09387751 0.66979827 0.40269437 0.55906662 0.03823901
 0.12888967 0.87111033]
34707.15643943977 TC: 34659.484
```

all the binary cases agree, but the ternary does not. It's unclear why, but it's blocking this issue until it's figured out.

bocklund:
This seems to be related to magnetism. I commented out the magnetic parameters in the ordered phase in the Al-C-Fe database and now the energies agree. Interestingly[1] the Curie temperature for the "non-magnetic ordered phase" is 301.8 K according to Thermo-Calc (this is also what pycalphad sees). 

The `atomic_ordering_energy` method replaces the TC and BMAG in the ordered phase from the disordered phase). I need to look into it more, but my guess is that the TC and BMAG are parameters should be handled like the other energetic contributions, TC = TC<sup>dis</sup> + ΔTC<sup>ord</sup>.


[1] This is actually really lucky that my test calculations were at 300 K, because if it were much higher than 300 K, the magnetic energy contribution would diminish and this bug would have gone un-noticed! It's actually unrelated to the issue this PR fixes and may have affected any database that includes magnetic parameters in the ordered phase)

Ebert-Daniel:
I think you are right. 
In H.L. Lukas, S.G. Fries, and B. Sundman. Computational Thermodynamics: the Calphad Method, volume 131. Cambridge University Press, Cambridge, United Kingdom, 2007. In section 5.8.6, it is said that the physical properties may also have their parameters partitioned between ordered and the disordered part.

bocklund:
Thanks @Ebert-Daniel! I have read through 5.8.2.4 and 5.8.4 two or three times since investigating #310 and I missed 5.8.6! Definitely key information.

bocklund:
Fixing the partitioning of the physical properties is going to be slightly more complicated. The changes here will partition the `TC`, `BMAG` and `NT` expressions and update the `Model` attributes appropriately (e.g. `Model().TC` will be correct). However, the AST expressions are different objects than the expressions in the attributes, so properly partitioning `Model().TC` in the ordered phase doesn't fix the Gibbs energy expression.

Using magnetic energy as an example: both the ordered energy and disordered energy need to have the partitioned `TC`, but there's currently no way for the ordered model to make the disordered magnetic energy use the partitioned `TC` because the values of the `TC` are inserted directly into the disordered magnetic energy expression and are not symbolically accessible later. This applies to all physical models, like the Einstein and two-state models.

The partitioning of the Gibbs energy contributions from physical models like the magnetic model is technically orthogonal to the issue this actually solves, so I propose to get this PR merged and raise a new issue to track the partitioning of physical properties. This PR still includes the failing test, but it is skipped for now. 

That would make this ready for review.

bocklund:
The failures are due to coverage upload failures. The tests passed.

pycalphad/model.py
```diff
@@ -952,84 +962,183 @@ def mole_fraction(species_name, phase_name, constituent_array,
...
+
     def atomic_ordering_energy(self, dbe):
         """
         Return the atomic ordering contribution in symbolic form.
-        Description follows Servant and Ansara, Calphad, 2001.
+
+        If the current phase is anything other than the ordered phase in a
+        paritioned order/disorder Gibbs energy model, this method will return
+        zero. If the current phase is the ordered phase, ordering energy is
+        computed by equation (18) of Connectable *et al.* [1]_:
```
richardotis, line 1000: typo in author's name

pycalphad/model.py
```diff
@@ -952,84 +962,183 @@ def mole_fraction(species_name, phase_name, constituent_array,
...
+            # explicit way to specify which sublattices are ordering. We try to
+            # compensate for this assumption by validating (next).
+            if len(disordered_subl_constituents.symmetric_difference(subl_constituents)) == 0:
+                substitutional_sublattice_idxs.append(idx)
+        # validate
+        num_substitutional_sublattice_idxs = len(substitutional_sublattice_idxs)
+        num_ordered_interstitial_subls = len(ordered_phase.sublattices) - num_substitutional_sublattice_idxs
+        num_disordered_interstitial_subls = len(disordered_phase.sublattices) - 1
+        if num_ordered_interstitial_subls != num_disordered_interstitial_subls:
+            raise ValueError(
```
richardotis, line 1072: Is there an easy test which could be written for this?

bocklund: Added in c5d0852

bocklund:
This should have addressed the outstanding review items and also added a warning from an out-of-band discussion to warn the user if they instantiate a Model that a physical `Model.contribution` that must be partitioned.

richardotis (approved):

PR #316, opened 2021-03-15
MAINT: Bump SymEngine pin to 0.7.0

bocklund:
Checklist
  * [x] `setup.py`
  * [x] `environment-dev.yml`

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -16,7 +16,7 @@ dependencies:
   - numpy>=1.13
   - pyparsing
   - python>=3.6
-  - python-symengine=0.7
+  - python-symengine=0.7.0.post1
   - scipy
   # We don't import the C++ symengine library directly in pycalphad, but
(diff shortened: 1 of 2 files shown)


richardotis (approved):
Does this need to be bumped for the post-releases?

bocklund:
> Does this need to be bumped for the post-releases?

It doesn't look like the post builds are being released on conda-forge. https://github.com/conda-forge/python-symengine-feedstock/pull/33/files

At least with conda, the single equals sign spec should match all post builds if I'm understanding correctly the [example](https://docs.conda.io/projects/conda-build/en/stable/resources/package-spec.html#examples):

> The fuzzy constraint numpy=1.11 matches 1.11, 1.11.0, 1.11.1, 1.11.2, 1.11.18, and so on.

For PyPI, assuming PEP-440 is followed, we would need `==0.7.0.*` to match the post builds [source](https://www.python.org/dev/peps/pep-0440/#version-matching)

PR #330, opened 2021-05-21
CI: Use PyPA/build to set up source distribution in deploy

bocklund:
- Use [PyPA/build](https://pypa-build.readthedocs.io/en/latest/index.html) to build distributions as recommended by [setuptools](https://setuptools.readthedocs.io/en/latest/build_meta.html#how-to-use-it). The source distribution builds and passes `twine check` as verified [here](https://github.com/pycalphad/pycalphad/runs/2641952743?check_suite_focus=true).
- Update `pyproject.toml` to explicitly use the legacy setuptools backend, which is required since we vendor a local versioneer and cannot do full PEP 517 build isolation.
- Add `long_description_content_type` to our `setup.py` to pass `twine check`


richardotis:
Thanks for working this. Does this ensure the tests are run and pass before a tagged commit is built and uploaded?

bocklund:
This particular workflow does not do that. I don't really see it as blocking because:

1. The current releasing workflow where developers upload source distributions manually doesn't necessarily ensure that the tests pass in CI
2. You can mitigate and ensure the tests pass "manually" on CI by pushing the commit you're planning to tag first to see if the tests pass. If they pass, you can tag and `git push origin master --tags` and trigger this workflow.

richardotis:
I agree that it's not a regression to do it this way and that it's fine to move forward. I would eventually like to test the exact artifacts we're uploading to PyPI, particularly once we start building platform wheels.

richardotis (approved):

PR #341, opened 2021-05-26
BLD: use setuptools_scm instead of versioneer

bocklund:
setuptools_scm is the PyPA blessed single source versioning solution. We are moving away from versioneer so we don't have to vendor it ourselves and becaue it depends on distutils which is deprecated in Python 3.10.

- Adds setuptools_scm as a runtime dependency so editable local installs can get the version dynamically
- Remove versioneer vendored module and shipped files from MANIFEST.in. We can now properly use `setuptools.build_meta` as our build system instead of `setuptools.build_meta:__legacy__`.
- Changes the version scheme to be strict [PEP 440](https://www.python.org/dev/peps/pep-0440/#version-matching). For example, our current version as of a commit on this branch is `0.8.6.dev19+gf3765968`. More version scheme details at https://github.com/pypa/setuptools_scm#default-versioning-scheme

Checklist
  * [x] `setup.py`
  * [x] `environment-dev.yml`

I checked the generated sdist and verified that the version in `PKG-INFO` matches the the one seen by `python setup.py --version`. The wheels show the version correctly in the CI as well.


richardotis:
Thanks for getting this done. This is a good step forward for compatibility with the future of Python packaging. We eventually would like to eliminate the runtime dependency on `setuptools_scm` by writing a hardcoded version file which is distributed in the packaged version, but currently PEP 517 build isolation doesn't work with editable installs using `pip install -e. `. There are workarounds, but the real fix is PEP 660: https://www.python.org/dev/peps/pep-0660/ 
Support for PEP 660 in `pip` is in progress: https://github.com/pypa/pip/pull/8212

So, for now, we can take the runtime import hit in packaged versions (for dev installs, it'll be the same) and eventually make the switch to hardcoded version files when PEP 660 is implemented.

bocklund:
I think the `__init__.py` implementation needs a small change to give `root` and `relative_to` arguments so that the dynamic versioning works for editable installs if you are outside the package directory.

bocklund:
The step didn't work in the tests. Not sure yet how to configure it to prefer the git lookup, but fall back on the version from the package metadata.

The bare `get_version()` seems to be able to look at the git version of the current directory or the package metadata. It seems like setting the `root` and `relative_to` allowed us to get the git version even from another directory with it's own git,  but it is not able to find the package metadata.

Not sure what the correct API usage for `get_version` should be, and it's not well documented.

Worst case, we do a try-except?

richardotis:
I think `get_version()` accepts keyword arguments corresponding to the entries in our `pyproject.toml`, and when we call it without those arguments, they are ignored.

bocklund:
Switched to a slightly different try/except scheme suggested by @richardotis.

The idea is that we use [this trick from StackOverflow](https://stackoverflow.com/questions/43348746/how-to-detect-if-module-is-installed-in-editable-mode/66869528#66869528) to use a file that isn't packaged as a hack for detecting if a user has a local checkout from an SCM system. If this file is present, the user has a local checkout/editable install, then we can get the version from the SCM using setuptools_scm. If the file is not present (i.e. regular installs from an sdist or a wheel), then the version c

 This adds `importlib_metadata` as an explicit dependency until we support Python 3.8 or later when `importlib.metadata` becomes part of the stdlib.

richardotis:
References for astropy's approach: https://github.com/astropy/astropy/pull/10774 https://github.com/astropy/astropy/pull/10831
Explanation for why we use a directory, instead of a single file, to exclude in `MANIFEST.in`: https://github.com/astropy/astropy/pull/10774#discussion_r501237395 https://github.com/pypa/setuptools/issues/511#issuecomment-570177072

richardotis:
Can the dependency for `importlib_metadata` be updated so it's only included for Python <3.8?

bocklund:
> Can the dependency for `importlib_metadata` be updated so it's only included for Python <3.8?

In the setup.py?

richardotis:
I was thinking in the setup.py and in the environment definition for the dev environment.

richardotis (approved):

richardotis:
It's also okay if they're just marked with a comment so we don't forget in ~2 years when Py38 support ends.

richardotis:
References for numpy ABI compatibility issues related to the release of numpy 1.20:
- https://github.com/scikit-learn-contrib/hdbscan/issues/457#issuecomment-773671043
- https://stackoverflow.com/questions/66060487/valueerror-numpy-ndarray-size-changed-may-indicate-binary-incompatibility-exp
Recommended solution using `oldest-supported-numpy` metapackage: https://numpy.org/devdocs/user/depending_on_numpy.html#build-time-dependency

richardotis (approved):

richardotis:
Getting some py37 errors at build-time, but everything seems to still build and work:
```
Collecting numpy==1.14.5; python_version == "3.7" and platform_machine != "aarch64" and platform_system != "AIX" and platform_python_implementation != "PyPy"
  Using cached numpy-1.14.5-cp37-none-win_amd64.whl (13.4 MB)
ERROR: scipy 1.6.3 has requirement numpy<1.23.0,>=1.16.5, but you'll have numpy 1.14.5 which is incompatible.
```

richardotis:
Changing `oldest-supported-numpy` to `numpy<1.20` in the `pyproject.toml` build requirements seems to work locally for me on Windows py37.

bocklund:
Things seemed to pass in reasonable time with the new pin. Do you want to give one last pass, @richardotis?

richardotis:
Editable mode seems to be working locally without errors, and it's still compatible with `python setup.py develop` so I don't have to pay the build isolation cost if I choose not to. Editable mode seems to work in my dirty Py37 working environment as well as a fresh Py38 environment.

richardotis (approved):

bocklund:
Not to suggest changing this yet again, but I think the editable workaround we are using would enable using the hard coded `write_to` mechanism instead of the import lib mechanism.

richardotis:
Open to it, but let's make it a separate PR.

bocklund:
I also tested a local development install manually as well as `pip install git+https://github.com/pycalphad/pycalphad.git@8a73bcb04c2627fb5793c626183ba0b244aa7d86`

PR #342, opened 2021-05-28
DOC/BLD: Fix documentation builds

bocklund:
- Sphinx `conf.py`: Don't load the version from a hardcoded `_version.py` (changes from #341)
- CI: Build (but don't deploy) documentation on pull requests so we can catch regressions in the docs in the future
- CI: Use pip to install rather than conda in docs builds

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - develop
-      - BLD-fix-sphinx-builds
   pull_request:  # all pull requests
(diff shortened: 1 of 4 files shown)


richardotis:
I notice an error from the removed `pycalphad.refdata` module in the docbuild. Everything else looks good.

bocklund:
Thanks for catching that - I rebuilt the RST file. Eventually I'd like to see us error on all warnings.

richardotis (approved):

PR #343, opened 2021-05-29
FIX: Minimizer: CompositionSet objects set for removal are not removed

bocklund:
The `new_free_stable_compset_indices` are created after phase removal [here](https://github.com/pycalphad/pycalphad/blob/657f5ba38ef582573a31cf6646ea5528286db83b/pycalphad/core/minimizer.pyx#L816), but these new stable indices with the phases removed are essentially reset [here](https://github.com/pycalphad/pycalphad/blob/657f5ba38ef582573a31cf6646ea5528286db83b/pycalphad/core/minimizer.pyx#L841) (by the code removed in this change).

This is just one way to address this issue, maybe the most straightforward? An alternative could be set the phase amounts to the phases scheduled for removal to zero and to leave in the code that reconstructs the `new_stable_compset_indices` (which won't include the phases with new zero phase amounts).

The removed code could also fix a bug where a fixed phase is included in the free stable phase indices if it has non-zero phase amount.


richardotis (approved):

richardotis:
I checked this with the six-component steel calculation and the binary phase diagram examples, and it looks like a good change.

Issue #347, opened 2021-06-03
Potential pyparsing 3.0 breakages


bocklund:
Working on #346, I installed a pre-release of pyparsing (`3.0.0b2`) where our tests failed to parse some databases:

`ALCOCRNI_TDB` and `ALNIFCC4SL_TDB`. All three cases share a common prefix of `FUN` shortened from `FUNCTION` (other previously defined functions that use the full `FUNCTION` seem to work).

Full output below:

<details>
```
__________________________________________________________________________________________ ERROR collecting anaconda3/envs/testenv/lib/python3.9/site-packages/pycalphad/tests/test_eqplot.py ___________________________________________________________________________________________
anaconda3/envs/testenv/lib/python3.9/site-packages/pycalphad/tests/test_eqplot.py:13: in <module>
    ALCOCRNI_DBF = Database(ALCOCRNI_TDB)
anaconda3/envs/testenv/lib/python3.9/site-packages/pycalphad/io/database.py:122: in __new__
    return cls.from_string(fname, fmt=fmt)
anaconda3/envs/testenv/lib/python3.9/site-packages/pycalphad/io/database.py:239: in from_string
    return cls.from_file(StringIO(data), **kwargs)
anaconda3/envs/testenv/lib/python3.9/site-packages/pycalphad/io/database.py:213: in from_file
    format_registry[fmt.lower()].read(dbf, fd)
anaconda3/envs/testenv/lib/python3.9/site-packages/pycalphad/io/tdb.py:961: in read_tdb
    tokens = grammar.parseString(command)
anaconda3/envs/testenv/lib/python3.9/site-packages/pyparsing/core.py:869: in parseString
    raise exc.with_traceback(None)
[...]

PR #348, opened 2021-06-04
FIX: pyparsing 3.0 support

richardotis:
Fixes gh-347, which reports a regression with the pyparsing 3.0 pre-release. This change should be backwards compatible with pyparsing 2.4+. I do not know the precise root cause of the regression, as I was able to narrow the problem down empirically. My best guess is that the semantics of `pyparsing.White()` have changed in some way.

@bocklund This should be run against your private collection of TDBs to identify regressions.

Checklist
  * [x] `setup.py`
  * [x] `environment-dev.yml`

### DIFF
--- a/pycalphad/io/tdb.py
+++ b/pycalphad/io/tdb.py
@@ -206,7 +206,7 @@ def _tdb_grammar(): #pylint: disable=R0914
     # a convenience function will handle the piecewise details
     func_expr = (float_number | ZeroOrMore(',').setParseAction(lambda t: 0.01)) + OneOrMore(SkipTo(';') \
         + Suppress(';') + ZeroOrMore(Suppress(',')) + Optional(float_number) + \
-        Suppress(Word('YNyn', exact=1) | White()))
+        Suppress(Optional(Word('Yy', exact=1))), stopOn=Word('Nn', exact=1)) + Suppress(Optional(Word('Nn', exact=1)))
     # ELEMENT
     cmd_element = TCCommand('ELEMENT') + Word(alphas+'/-', min=1, max=2) + ref_phase_name + \
         float_number + float_number + float_number + LineEnd()
--- a/setup.py
+++ b/setup.py
@@ -58,7 +58,7 @@ setup(
         'importlib_metadata',  # drop when pycalphad drops support for Python<3.8
         'matplotlib>=3.3',
         'numpy>=1.13',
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -58,7 +58,7 @@ def read(fname):
         'importlib_metadata',  # drop when pycalphad drops support for Python<3.8
         'matplotlib>=3.3',
         'numpy>=1.13',
-        'pyparsing>=2.0,<3',
```
bocklund, line 61: This one is `2.0` while the other one is `2`. Do we need to set `2.4` as the minimum per your comment?

pycalphad/io/tdb.py
```diff
@@ -206,7 +206,7 @@ def _tdb_grammar(): #pylint: disable=R0914
     # a convenience function will handle the piecewise details
     func_expr = (float_number | ZeroOrMore(',').setParseAction(lambda t: 0.01)) + OneOrMore(SkipTo(';') \
         + Suppress(';') + ZeroOrMore(Suppress(',')) + Optional(float_number) + \
-        Suppress(Word('YNyn', exact=1) | White()))
+        Suppress(Optional(Word('Yy', exact=1))), stopOn=Word('Nn', exact=1)) + Suppress(Optional(Word('Nn', exact=1)))
```
bocklund, line 209: Does `PARAMETER` need similar treatment?

bocklund:
I’ll try to get the pre-release installed tomorrow and test against the databases I have.

setup.py
richardotis: At minimum we need 2.1 for the `stopOn` keyword argument, per https://github.com/pyparsing/pyparsing/blob/d108a29db062c9250f50c978e3a86381d1b0746b/CHANGES#L1552

Since we aren't backtesting pyparsing extensively and there do seem to be subtle differences between versions, pinning to `pyparsing>=2.4` is probably the best move.

bocklund: Sounds good to me

pycalphad/io/tdb.py
bocklund: That is what I meant, sorry!

bocklund:
I compared 3.0.0b2 to 2.4.7 locally and there were no regressions across 522 databases.

bocklund (approved):

PR #350, opened 2021-06-05
WIP/BLD: Unify requirements specification

richardotis:
This is a sketch of a concept, inspired by [PEP 621](https://www.python.org/dev/peps/pep-0621/#example), for putting both build- and run-time dependencies into a single static file, `pyproject.toml`. While setuptools does not yet have official PEP 621 support, nothing prevents us from putting as much metadata as we can into `pyproject.toml` and parsing it from inside `setup.py` at build time, which is the approach taken here.

Also, `environment-dev.yaml` is removed, tentatively, though this may need more discussion.

I have also added a `develop.py` file, which works around https://github.com/pypa/pip/issues/9314 by shelling out to `pip` to ensure build dependencies are satisfied before we run an editable installation. Thus developers only need to run `python develop.py` in their development iteration loop.

Checklist
* [ ] The documentation examples have been regenerated if the Jupyter notebooks in the `examples/` have changed. To regenerate the documentation examples, run `jupyter nbconvert --to rst --output-dir=docs/examples examples/*.ipynb` from the top level directory)
* [ ] If any dependencies have changed, the changes are reflected in the
  * [ ] `setup.py`
  * [ ] `environment-dev.yml`
  * [ ] `pyproject.toml`


bocklund (commented):
I am for moving to PEP 621 and the changes made to setup.py. I also support dropping environment-dev.yml.

It's a great idea to put all our dependencies in one spot and something I'd like to do as well. After some out-of-band discussions around the `develop.py`, I'm not yet convinced that it's enough of a value-add to do something non-standard and hopefully this is something that pip puts forward eventually. In practice, it doesn't really reduce the number of steps a user has to take, since it has a dependency on the `toml` package to run. It does alleviate having multiple files that developers have to update if the dependencies change.

richardotis:
I hear you on `develop.py`, and the lack of a stdlib toml parser makes this problem annoying. What would you think about a `install-build-deps.py` that a developer could ostensibly call one time?

```python
import subprocess
import sys
import pathlib
try:
    import toml
except ImportError:
    subprocess.run([sys.executable, "-m", "pip", "install", "toml"])
    import toml

path_to_develop = pathlib.Path(__file__).parent.absolute()

c = toml.load(path_to_develop / "pyproject.toml")
build_deps = c["build-system"]["requires"]
subprocess.run([sys.executable, "-m", "pip", "install"] + build_deps)
```

bocklund:
> I hear you on `develop.py`, and the lack of a stdlib toml parser makes this problem annoying. What would you think about a `install-build-deps.py` that a developer could ostensibly call one time?

Nice workaround. IMO, this is something to keep in our back pocket. I think that there's some value in standardizing on the `requirements.txt` pattern and keeping the user in-the-loop about what is being done, rather than hiding it in a custom script.

richardotis:
I think it may be too early to move metadata out of `setup.py`. A partial/non-standard `pyproject.toml` may cause problems in the future if setuptools gains native PEP 621 support, whereas the current approach is likely to be backwards compatible for a long time.

Considering the full discussion here, as well as the documentation improvements being made in #351, I think we can close this PR and return at a later date, preferably [if setuptools gains the native ability to be configured by `pyproject.toml`](https://github.com/psf/fundable-packaging-improvements/blob/master/FUNDABLES.md#add-support-for-pyprojecttoml-as-a-way-to-configure-setuptools).

PR #351, opened 2021-06-05
DOC: Overhaul installation documentation

### LINKED ISSUE #320: DOC: pycalphad installation documentation overhaul
The installation docs could be improved:

## General issues

1. Conda environments (also in above). Consider what to do for release versions. Environments are best practices and it's a lot easier to manage pycalphad starting from a clean slate environment with no packages installed. Many of our users are relatively new Python and/or Conda users and forgetting to activate is such a common gotcha that we've advocated to install into `base`. If we do choose to install into an environment, it's critical for users to install Jupyter into the environment too (and make sure that Jupyter from the `base` environment can pick up the kernel - it might be necessary to install `nb_conda`, but I'm not sure.)
2. Most pycalphad users probably also want to install Jupyter, though it is not an explicit dependency for the package. It's in `base` today (but maybe won't be forever?) and needs to be installed separately for other environments.

## Development mode specific issues

3. (Windows) Windows users need to bring their own compilers. Installing the `c-compiler` and `cxx-compiler` packages on Windows basically gives the [`vc` package](https://github.com/conda-forge/vc-feedstock), which provides an [activation script](https://github.com/conda-forge/vc-feedstock/blob/master/recipe/activate.bat) for a user-installed MSVC.
4. (Windows) Most Windows users will also need to install `git`, since it does not come with the OS.

Windows Subsystem for Linux is an option, but it's not a "true" Windows installation and users advanced enough to install and use WSL should be able to follow the instructions as a Linux user (and recognize that they should do so). [This tutorial](https://beenje.github.io/blog/posts/how-to-setup-a-windows-vm-to-build-conda-packages/) is linked to in the conda-forge documentation and really good and simple for walking users through installing `git` and a compatible MSVC compiler. Users can be instructed to skip to the "Developer tools installation" section.

5. Users should be instructed to create Conda environments for development versions. Activating has an extra barrier, but ideally most users interested in development installations should be familiar enough to use Conda environments and it's worth the extra protection from wrecking the `base` environment or losing a release version of pycalphad
6. The development mode steps should include steps to build a working pycalphad without installing and removing it.

bocklund:
This PR:

- Adds a JS-driven configuration tool to the installation docs to help configure the installation steps.
  - This is where users can configure some basic system- and configuration-specific options to get the installation that's right for them.
  - Mainly, this resolves items 1, 2, 5, 6, and 7 in #320 because we are giving users options w.r.t. pip vs. conda, environments and Jupyter. 
  - The idea of the tool is that the HTML file, `installation_configurator.html` is mostly static except and the configurations are automatically generated and hardcoded with the logic from `installation_configurator_generator.py`. The HTML block is copy-pasted into the HTML file.
- Rewrites each tool sections to give an overview of the main installation highlights. Because of the configuration tool, we don't have to spend a lot of time explaining the nuances of every possible configuration and platform.
- The text mentions that git and compilers are required and points to a tutorial for installing the ones needed for Windows, addressing items 3 and 4 of #320.
- Adds `requirements-dev.txt` to give the build and development requirements beyond the runtime requirements. Used in the source installation steps.

Closes #320. 

Credit to spaCy's excellent [installation documentation](https://spacy.io/usage) for the inspiration on the configuration tool and layout of the installation documentation content.


richardotis:
Great work, this looks awesome. For mobile it looks like the configurator isn't responsive in its layout yet:
image

bocklund:
> Great work, this looks awesome. For mobile it looks like the configurator isn't responsive in its layout yet:
> image

Thank you. 

For mobile, we can probably take the lead here from spaCy again, I just have to figure out the right CSS.

richardotis:
I'm also open to toggling the docs theme over to Furo, if you're interested.

bocklund:
image

Responsive on an iPhone 5SE size screen

I also updated to Furo theme and removed the environment-dev.yml

bocklund:
I fixed the docs builds in CI, so it's passing now. I used the same commands as the configuration tool says to, so we have some "rough" testing of that, at least. I had to add the `-U` upgrade flag to get setuptools to be upgrade correctly (consistent with what the spaCy docs do).

I also added the ability to upload documentation build artifacts to GitHub, so they can be downloaded the GitHub Actions interface, for example: [here](https://github.com/pycalphad/pycalphad/actions/runs/910658198). You can simply unpack that at your local machine, enter the directory, and run `python -m http.server` to run the docs as-built 😄 

Coverage uploads are still broken and that's why the tests are failing.

bocklund:
Dark mode with the Furo doesn't look that good with all our content yet and there's no way to hard-code only light mode or give users a toggle to switch between them yet. Not sure if that's a dealbreaker.

richardotis (approved):

PR #352, opened 2021-06-06
FIX: Change checkout action for website deploy

richardotis:


bocklund:
I like this solution better than mine, but at least the netlify works :)

PR #357, opened 2021-06-08
MAINT: Refactor solver internals

bocklund:
- Remove `Problem` and update the `Solver.solve` API to `Solver.solve(composition_sets, conditions)`.
- Combine `_solve_and_update_if_converged` and `pointsolve` into `solve_and_update`. I'm not sure `solve_and_update` being a `cpdef` function would impact performance at all compared to `_solve_and_update_if_converged` being a `cdef` function. `_solve_and_update_if_converged` was calling out to our pure Python `Solver()` class anyway.
- Remove the need to pass around the list of active `Species` objects in the solver internals. Species were primarily used to determine the active, system-wide, non-vacant pure elements. Now the code uses the non-vacant pure elements in the 
`PhaseRecord.nonvacant_elements` attribute. This implies that `PhaseRecord.components` (of type: `List[Species]`) must be system-wide active species, i.e. mixing and matching `CompositionSet`/`PhaseRecord` objects from different active species subsets of a system is not allowed. This was already enforced implicitly via `build_phase_records`. Mixing and matching `PhaseRecords` in the same calculation would probably error or give silently incorrect results today, so this isn't a regression, but it does limit our ability to change this later without changing the API again.

I'm open to thoughts/suggestions. 

- Is `solve_and_update` useful enough compared to a user (and the internals) just calling `Solver.solve()` and `update_composition_sets` manually? I'd imagine that most of the time a user who is using the `solve_and_update` would want the composition sets to be removed automatically, but it's only one extra line of code to update and remove via `update_composition_sets`.


richardotis:
I'll review soon, but just a few preliminary responses to your main points:

> * Combine `_solve_and_update_if_converged` and `pointsolve` into `solve_and_update`. I'm not sure `solve_and_update` being a `cpdef` function would impact performance at all compared to `_solve_and_update_if_converged` being a `cdef` function. `_solve_and_update_if_converged` was calling out to our pure Python `Solver()` class anyway.

The performance impact should be negligible given that we are calling back into Python anyway, as you note. Even if we fully Cythonize the path from `solve_and_update` to `find_solution`, calling a cpdef function from Cython should be about the same as calling a cdef function from Cython.

> * Remove the need to pass around the list of active `Species` objects in the solver internals. Species were primarily used to determine the active, system-wide, non-vacant pure elements. Now the code uses the non-vacant pure elements in the
>   `PhaseRecord.nonvacant_elements` attribute. This implies that `PhaseRecord.components` (of type: `List[Species]`) must be system-wide active species, i.e. mixing and matching `CompositionSet`/`PhaseRecord` objects from different active species subsets of a system is not allowed. This was already enforced implicitly via `build_phase_records`. Mixing and matching `PhaseRecords` in the same calculation would probably error or give silently incorrect results today, so this isn't a regression, but it does limit our ability to change this later without changing the API again.

This assumption is correct. In my "local equilibrium" test code, I added an assertion to verify that `PhaseRecord.components` was the same for all specified `PhaseRecord` objects:
```python
# PhaseRecord components should be the same for all composition sets
all_pr_components = [cs.phase_record.components for cs in composition_sets]
assert all_pr_components.count(all_pr_components[0]) == len(all_pr_components)
```

> * Is `solve_and_update` useful enough compared to a user (and the internals) just calling `Solver.solve()` and `update_composition_sets` manually? I'd imagine that most of the time a user who is using the `solve_and_update` would want the composition sets to be removed automatically, but it's only one extra line of code to update and remove via `update_composition_sets`.

Great question. For stepping/mapping, deleting the metastable composition sets is not necessarily desirable (since they may come back). For point computations, it could be cleaner UX for metastable composition sets to not be shown. It might be that this could be addressed with a better repr for `CompositionSet`, `SolverResult`, or by wrapping `composition_sets` in a list-like container which has a special repr. In any case, this distinction is not presently surfaced in the API at all and should be addressed.

pycalphad/core/eqsolver.pyx
```diff
@@ -5,8 +5,6 @@ cimport numpy as np
 cimport cython
 cdef extern from "_isnan.h":
     bint isnan (double) nogil
-import scipy.spatial
```
richardotis, line 8: With the removal of this import, pycalphad only depends on `scipy` for the Cython wrappers for LAPACK in `scipy.linalg.cython_lapack`. Specifically, we depend on `cython_lapack.dgelsd()` (least squares) and `cython_lapack.dgesv()` (solve Ax=b).

pycalphad/core/eqsolver.pyx
```diff
@@ -84,41 +82,43 @@ cdef bint add_new_phases(object composition_sets, object removed_compsets, objec
...
-# comps: List[v.Species]
-# cur_conds: OrderedDict[str, float]
-# iter_solver: SolverBase instance
-cpdef pointsolve(composition_sets, comps, cur_conds, iter_solver):
-    "Mutates composititon_sets with updated values if it converges. Returns SolverResult."
-    return _solve_and_update_if_converged(composition_sets, comps, cur_conds, Problem, iter_solver)
 
-cdef _solve_and_update_if_converged(composition_sets, comps, cur_conds, problem, iter_solver):
-    "Mutates composititon_sets with updated values if it converges. Returns SolverResult."
+cpdef update_composition_sets(composition_sets, solver_result, remove=True):
```
richardotis, line 86: Could the keyword argument be renamed to something more descriptive, like `remove_metastable`?

pycalphad/core/eqsolver.pyx
```diff
@@ -84,41 +82,43 @@ cdef bint add_new_phases(object composition_sets, object removed_compsets, objec
...
-    for idx in reversed(compsets_to_remove):
-        del composition_sets[idx]
+    if remove:
+        # Watch removal order here, as the indices of composition_sets are changing!
+        for idx in reversed(compsets_to_remove):
+            del composition_sets[idx]
+
+
+# composition_sets: List[CompositionSet]
+# cur_conds: OrderedDict[str, float]
```
richardotis, line 112: What do you think about allowing a `OrderedDict[v.StateVariable, float]` to be passed in here, and taking care of the conditions dict conversion for the user? The `solve` interface will remain available for users who don't want to pay the performance penalty.

richardotis (changes requested):

pycalphad/core/eqsolver.pyx
richardotis: I thought about this some more, and it's okay to defer a decision for now, since we can always add support for it later. I anticipate more refactoring of this part of the core (if not this exact function) in the near future, so we may want to see how our revised handling of, e.g., starting points interacts with this function.

pycalphad/core/eqsolver.pyx
richardotis: Agreed. In a world of fast network connections and binary packages, the weight of SciPy as a dependency is not a big deal. Plus, users are very likely to already have it installed. In addition, linking LAPACK is apparently a non-trivial and highly platform-dependent activity which I see extremely limited benefit for pycalphad to take on, given that SciPy is already meeting the need.

bocklund:
Screen Shot 2021-06-08 at 4 48 03 PM

Screen Shot 2021-06-08 at 4 49 49 PM

@richardotis thoughts on which to use for the signature?

richardotis:
I think the top one (with indent) is fine.

richardotis (approved):

PR #360, opened 2021-06-30
ENH: Refactor solver and improve solver performance

bocklund:
I have noticed in benchmarking that roughly 1/3 of the time in the solver is spent performing this copy operation on the `state`. We don't do anything meaningful with the old state except for using its data, so I propose to remove it and replace the couple instances with explicitly storing the relevant variables before taking a step.

If we want to do more advanced things with the previous state(s) later, that's definitely something we could do, but I think all the memory allocations in one of the innermost loops is probably not worth it compared to copying the values we need out.

This change could probably use a more real benchmark to justify it, but I wanted to get it posted for feedback before I invest much more time into it.

### DIFF
--- a/pycalphad/core/eqsolver.pyx
+++ b/pycalphad/core/eqsolver.pyx
@@ -250,7 +250,7 @@ def _solve_eq_at_conditions(properties, phase_records, grid, conds_keys, state_v
             result = solve_and_update(composition_sets, cur_conds, iter_solver)
 
             chemical_potentials[:] = result.chemical_potentials
-            changed_phases |= add_new_phases(composition_sets, removed_compsets, phase_records,
+            changed_phases = add_new_phases(composition_sets, removed_compsets, phase_records,
                                             grid, curr_idx, chemical_potentials, state_variable_values,
                                             1e-4, verbose)
             iterations += 1
--- a/pycalphad/core/eqsolver.pyx
+++ b/pycalphad/core/eqsolver.pyx
@@ -203,7 +203,7 @@ def _solve_eq_at_conditions(properties, phase_records, grid, conds_keys, state_v
                                     [np.asarray(properties.coords[b][a], dtype=np.float_)
                                      for a, b in zip(it.multi_index, conds_keys)]))
         # assume 'points' and other dimensions (internal dof, etc.) always follow
-        curr_idx = [it.multi_index[i] for i, key in enumerate(conds_keys) if key in str_state_variables]
+        curr_idx = tuple(it.multi_index[i] for i, key in enumerate(conds_keys) if key in str_state_variables)
         state_variable_values = [cur_conds[key] for key in str_state_variables]
         state_variable_values = np.array(state_variable_values)
         # sum of independently specified components
@@ -235,6 +235,22 @@ def _solve_eq_at_conditions(properties, phase_records, grid, conds_keys, state_v
             composition_sets.append(compset)
         chemical_potentials = prop_MU_values[it.multi_index]
         energy = prop_GM_values[it.multi_index]
+        # Add unrepresented phases as metastable composition sets
+        # This should help catch phases around the limit of stability
+        unrepresented_phases = sorted(set(phase_records.keys()) - set(prop_Phase_values[it.multi_index]) - {'', '_FAKE_'})
+        driving_forces = (chemical_potentials * grid.X[np.index_exp[curr_idx]]).sum(axis=-1) - grid.GM[np.index_exp[curr_idx]]
+        for phase_name in unrepresented_phases:
+            phase_record = phase_records[phase_name]
+            phase_amt = 0.0
+            compset = CompositionSet(phase_record)
+            idx_grid_for_phase = grid.Phase[np.index_exp[curr_idx]] == phase_name
+            minimum_df_idx = np.argmax(driving_forces[idx_grid_for_phase])
+            if driving_forces[idx_grid_for_phase][minimum_df_idx] < 1000:
+                continue
+            df_idx = np.arange(len(driving_forces))[idx_grid_for_phase][minimum_df_idx]
+            sfx = np.atleast_1d(np.squeeze(grid.Y[curr_idx + (df_idx,) + np.index_exp[:phase_record.phase_dof]]))
+            compset.update(sfx, phase_amt, state_variable_values)
+            composition_sets.append(compset)
         #print('Composition Sets', composition_sets)
         phase_amt_sum = 0.0
         for compset in composition_sets:
--- a/pycalphad/core/eqsolver.pyx
+++ b/pycalphad/core/eqsolver.pyx
@@ -203,7 +203,7 @@ def _solve_eq_at_conditions(properties, phase_records, grid, conds_keys, state_v
                                     [np.asarray(properties.coords[b][a], dtype=np.float_)
                                      for a, b in zip(it.multi_index, conds_keys)]))
         # assume 'points' and other dimensions (internal dof, etc.) always follow
-        curr_idx = tuple(it.multi_index[i] for i, key in enumerate(conds_keys) if key in str_state_variables)
+        curr_idx = [it.multi_index[i] for i, key in enumerate(conds_keys) if key in str_state_variables]
         state_variable_values = [cur_conds[key] for key in str_state_variables]
         state_variable_values = np.array(state_variable_values)
         # sum of independently specified components
@@ -235,22 +235,6 @@ def _solve_eq_at_conditions(properties, phase_records, grid, conds_keys, state_v
             composition_sets.append(compset)
         chemical_potentials = prop_MU_values[it.multi_index]
         energy = prop_GM_values[it.multi_index]
-        # Add unrepresented phases as metastable composition sets
-        # This should help catch phases around the limit of stability
-        unrepresented_phases = sorted(set(phase_records.keys()) - set(prop_Phase_values[it.multi_index]) - {'', '_FAKE_'})
-        driving_forces = (chemical_potentials * grid.X[np.index_exp[curr_idx]]).sum(axis=-1) - grid.GM[np.index_exp[curr_idx]]
-        for phase_name in unrepresented_phases:
-            phase_record = phase_records[phase_name]
-            phase_amt = 0.0
-            compset = CompositionSet(phase_record)
-            idx_grid_for_phase = grid.Phase[np.index_exp[curr_idx]] == phase_name
-            minimum_df_idx = np.argmax(driving_forces[idx_grid_for_phase])
-            if driving_forces[idx_grid_for_phase][minimum_df_idx] < 1000:
-                continue
-            df_idx = np.arange(len(driving_forces))[idx_grid_for_phase][minimum_df_idx]
-            sfx = np.atleast_1d(np.squeeze(grid.Y[curr_idx + (df_idx,) + np.index_exp[:phase_record.phase_dof]]))
-            compset.update(sfx, phase_amt, state_variable_values)
-            composition_sets.append(compset)
         #print('Composition Sets', composition_sets)
         phase_amt_sum = 0.0
         for compset in composition_sets:
--- a/pycalphad/core/minimizer.pyx
+++ b/pycalphad/core/minimizer.pyx
@@ -766,7 +766,7 @@ cdef void remove_and_consolidate_phases(SystemSpecification spec, SystemState st
             state.phase_amt[dof_idx] = 0
 
 
-cdef bint change_phases(SystemSpecification spec, SystemState state, metastable_phase_iterations, times_compset_removed, can_add_phases):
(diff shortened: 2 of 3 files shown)


bocklund:
An alternative to this could be to just have one `old_state` where these quantities are copied in using a similar pattern instead of copying every iteration.

richardotis:
This approach is fine.

pycalphad/core/minimizer.pyx
```diff
@@ -750,13 +751,16 @@ cpdef find_solution(list compsets, int num_statevars, int num_components,
...
 
-        old_state = copy(state)
+        previous_chemical_potentials[:] = state.chemical_potentials[:]
+        previous_phase_amt[:] = state.phase_amt[:]
+        previous_phase_compositions[:, :] = state.phase_compositions[:, :]
+
         take_step(spec, state, step_size)
-        delta_phase_amt = np.array(state.phase_amt) - np.array(old_state.phase_amt)
-        if ((state.mass_residual > 1e-2) and (not np.all(np.array(state.chempot_diff) < 1.0))) or (iteration == 0):
+        delta_phase_amt = np.asarray(state.phase_amt) - np.asarray(previous_phase_amt)
```
richardotis, line 762: Profiling would tell you if it makes a difference or not, but this could be rewritten to avoid allocating a new array if you use a `for` loop.

bocklund:
Below are some profiling runs from a 10 iteration ESPEI run of a binary system (with Cython `profile` directive enabled).

These changes give ~30% performance improvement. I did a couple runs with Cython profiling on and off and the improvement in the `find_solution` timings is repeatable. 

I also tried annotating the Cythonized output and playing with some Cython optimizing directives, which didn't seem to make much measurable difference.

#### Develop
 develop-profile

#### This branch
minimizer-no-copy-state-profile

The fourth box to the right are calls to `np.all` (14 seconds in total), so there's a little to squeeze out, but there are probably hotter spots to optimize still.

richardotis:
Wow, copying `SystemState` objects is really expensive. 30% performance improvement for such a small change definitely sounds good to me.

bocklund:
Copying the state is about half the overall speedup (it's the labeled box in the fourth row of the develop branch benchmarks, accounting for 82 seconds). The other half comes from refactoring the allocations and calls to `np.max` where the lazier `np.all` or `np.any` calls give the same result.

pycalphad/core/minimizer.pyx
```diff
@@ -600,6 +582,20 @@ cdef class SystemState:
                 for i in range(num_phase_dof+spec.num_statevars):
                     csst.moles_normalization_grad[i] += csst.mass_jac[comp_idx, i]
 
+    cdef double[::1] driving_forces(self, SystemSpecification spec):  # TODO: spec only used for # of components, is that something the State should know on its own?
```
richardotis, line 585: You can avoid passing in the spec by accessing `len(self.compsets[0].phase_record.nonvacant_elements)` if all you need is the number of components.

richardotis: You have to do `compset = self.compsets[0]` so that you get the fully-typed attribute access

richardotis: Or just `self.masses.shape[0]` which is initialized to be equal to `spec.num_components`

bocklund:
Fresh benchmarks on https://github.com/pycalphad/pycalphad/pull/360/commits/e20464f2c9a16d60075256a9a5e0283ddb41cb1d Screen Shot 2021-07-15 at 9 54 23 PM

bocklund:
The bigger step size and new Gibbs phase rule code calls `take_step` about half as many times as the previous benchmark run, so that's why the total `find_solution` time is about 70% faster in the most recent commit compared to my previous benchmark.

pycalphad/core/minimizer.pyx
bocklund: This should be addressed now. I used the shape of the chemical potentials, since we use those to compute the driving forces anyway.

bocklund:
I think the changes here are ready for review. All of these changes account for ~2.5x speedup overall.

The PR contains the following changes:

## Possible changes to correctness
- the step size in `find_solution` is now hardcoded to `1.0`. Previously, we had a default step size of `0.1` with possibly bigger steps up to size `1.0` if the mass residual and change in energy was small. These larger steps sometimes destabilize phases (stepping to negative phase amounts) more aggressively, but the larger steps with fewer total iterations seems to make up for any additional phase cycling.
- Add extra metastable composition sets in `_solve_eq_at_conditions` from the [richardotis/fix-gibbs-phase-rule branch](https://github.com/richardotis/pycalphad/tree/fix-gibbs-phase-rule)
- `compute_phase_matrix` has the terms removed for subtracting a contribution of the `chemical_potential*mass_hessian`, which I think is not correct according to the equations in Sundman's 2015 paper. @richardotis please double check me on this.
- `change_phases` (name changed from `check_convergence_and_change_phases`) had code added to obey Gibbs phase rule from the [richardotis/fix-gibbs-phase-rule branch](https://github.com/richardotis/pycalphad/tree/fix-gibbs-phase-rule) (correctness change).
- `remove_and_consolidate_phases` (see below) now ensures that any consolidated phases have some non-zero phase amount to fix a bug where the total number of moles could become zero if consolidated phases had equal and opposite phase amounts and all the other phases became unstable in the same iteration.

## Semantic/numeric changes
- Do not make a copy of the `SystemState` every iteration which is expensive due to the allocations. Instead, we only need the chemical potentials, phase amounts, and phase compositions from the last iteration.
- Factor out all explicit array allocations in the minimizer. All array allocations occur exclusively in the object initializers (`SystemState`, `CompsetState`, and `SystemSpecification`) or in `find_solution`.
  - There's one exception in that `take_step` does allocations for the equilibrium matrix and equilibrium solution/rhs which change size if the number of stable phases changes.
  - There are also still many lingering allocations from calls into NumPy (e.g. `np.array`) or creation of `set` objects, etc. These cases will probably need a more careful look because sometimes the copying is logically required, for example unilaterally replacing `np.array(x)` with `np.asarray(x)` (which does not copy if `x` is already an array) breaks some tests.
- Dead code was removed in many places, in particular, the largest change was to remove `phase_rhs` from the `CompsetState`, which was not used anywhere outside or inside the class.
- `find_solution` was simplified and refactored. The core iterating loop now has 44 lines, down from 121 lines.
  - the code to remove and consolidate phases in `find_solution` was refactored to a `remove_and_consolidate_phases` function
  - driving forces computed in `find_solution` are now a method in `SystemState`
  - code to compute `delta_m` is removed since it was not used after changes to the step size control
  - slight change in the semantics where `phase_counter` is replaced with `iterations_since_last_phase_change` which counts up instead of `phase_counter` counting down with a floor of zero
  - `check_convergence_and_change_phases` was renamed to `change_phases` with a slight refactoring of the arguments
- replace some calls to `np.max`, etc. with lazier `np.any`/`np.all` calls

## Future work

My estimation is that there's at least another 20-30% that could be shaved off by continuing to hunt down allocating operations. It might be helpful to continue to do some line profiling or refactor to some small temporary functions to help track these down. Algorithmically, I think there is room to reduce the total number of iterations by:

- having smarter phase changes (phase addition, but also by tracking the history of sets of feasible phases to prevent cycling). 
- improving starting point generation to not include phases that are very far from the starting point hyperplane
- taking smarter steps in phase amount (similar to Δy) where we bound the length of a step in phase amounts so that no phase can have negative amount. This strategy would prevent mass from disappearing from the system when `remove_and_consolidate_phases` removes or consolidates phases with very negative phase amounts.

bocklund:
The benchmark above was for a binary case. I also ran a benchmark equilibrium calculation in a 5 component system and can verify the performance holds up compared to `develop` where we still see a 2.6x overall improvement. 

In the 5 component benchmark, each call to `take_step` is about 25% faster than develop, but there are almost 2x as many steps on average to find the solution, leading to a 50% slowdown in the time per `find_solution`. However, `find_solution` is called only 400 times compared to 1800 (a 4x decrease), so we are taking half as many steps in total here compared to `develop`. I attribute this behavior largely to having more metastable candidates and phase cycling in the 5 component case, but fewer overall calls to `find_solution` because the true stable composition sets likely make it into the candidates earlier.

bocklund:
I noticed that `map_binary` still has pdens set to 2000. The `_solve_eq_at_conditions` has a performance regression here due to the new code added, which goes away almost completely by lowering the pdens down to 50. Since users can pass higher point densities, this might be something to look into before we merge.

pycalphad/core/eqsolver.pyx
```diff
@@ -235,6 +235,22 @@ def _solve_eq_at_conditions(properties, phase_records, grid, conds_keys, state_v
...
+            phase_amt = 0.0
+            compset = CompositionSet(phase_record)
+            idx_grid_for_phase = grid.Phase[np.index_exp[curr_idx]] == phase_name
+            minimum_df_idx = np.argmax(driving_forces[idx_grid_for_phase])
+            if driving_forces[idx_grid_for_phase][minimum_df_idx] < 1000:
+                continue
+            df_idx = np.arange(len(driving_forces))[idx_grid_for_phase][minimum_df_idx]
+            sfx = np.atleast_1d(np.squeeze(grid.Y[curr_idx + (df_idx,) + np.index_exp[:phase_record.phase_dof]]))
+            compset.update(sfx, phase_amt, state_variable_values)
+            composition_sets.append(compset)
```
richardotis, line 253: The changes in `eqsolver.pyx` can be reverted and worked in a future PR per offline discussion. It's more correct to write `if driving_forces[idx_grid_for_phase][minimum_df_idx] < -1000` (note the flipped sign) but the larger issue is `driving_forces` is sometimes containing positive values, which should be impossible given a well-formed starting point. That issue will be worked separately from this PR.

richardotis (changes requested):
Great work!

pycalphad/core/minimizer.pyx
bocklund: Resolving because this code is removed

pycalphad/core/minimizer.pyx
```diff
@@ -695,49 +648,145 @@ cpdef take_step(SystemSpecification spec, SystemState state, double step_size):
...
+        else:
+            state.free_stable_compset_indices = np.array(sorted(set(state.free_stable_compset_indices) - compset_indices_to_remove), dtype=np.int32)
+            phases_changed = True
+    return phases_changed
+
+cdef bint change_phases(SystemSpecification spec, SystemState state, int[::1] metastable_phase_iterations, int[::1] times_compset_removed):
+    cdef int idx
+    phase_amt = state.phase_amt
+    current_free_stable_compset_indices = state.free_stable_compset_indices
+    driving_forces = state.driving_forces()
```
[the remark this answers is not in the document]
bocklund: I'm not sure if this is something Cython can infer, but I added the type anyway.

pycalphad/core/minimizer.pyx
```diff
@@ -695,49 +648,145 @@ cpdef take_step(SystemSpecification spec, SystemState state, double step_size):
...
+    if len(compsets_to_add) > 0:
+        if max_allowed_to_add < 1:
+            # We are at the maximum number of allowed phases, yet there is still positive driving force
+            # Destabilize one phase and add only one phase
+            possible_phases_to_destabilize = set(current_free_stable_compset_indices) - compsets_to_add - compsets_to_remove
+            # Arbitrarily pick the lowest index to destabilize
+            idx_to_remove = sorted(possible_phases_to_destabilize)[0]
+            compsets_to_remove.add(idx_to_remove)
+            phase_amt[idx_to_remove] = 0
+            # TODO: This should be ordered by driving force
```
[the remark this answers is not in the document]
bocklund: I agree. With some of the refactoring here, we're in good shape to make another pass to clean up and improve the phase changing logic. I think there's a case for an object that tracks the history of the attempted composition sets, replacing and holding on to some of the behavior we get from `metastable_phase_iterations` and `times_compset_removed`.

Is there anything you think should be addressed in this PR for this set of code?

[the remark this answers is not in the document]
richardotis: No, not for this PR.

pycalphad/core/minimizer.pyx
```diff
@@ -617,9 +622,10 @@ cpdef take_step(SystemSpecification spec, SystemState state, double step_size):
         state.chemical_potentials[comp_idx] = spec.initial_chemical_potentials[comp_idx]
 
     # Update phase internal degrees of freedom
-    for idx, compset in enumerate(state.compsets):
+    for idx in range(len(state.compsets)):
         # TODO: Use better dof storage
         x = state.dof[idx]
+        compset = state.compsets[idx]
```
richardotis, line 628: I don't think it's necessary to assign `compset` here.

richardotis (changes requested):

pycalphad/core/minimizer.pyx
bocklund: Removed

richardotis (approved):

PR #361, opened 2021-07-19
ENH: Allow passing dictionaries of PhaseRecord objects to equilibrium and calculate

bocklund:
Changes:

- `build_phase_records`
  - Defaults to build derivatives of the output. This is the more common case and SymEngine makes the compilation fast enough that's it's not as much of a burden to build them by default.
  - The `comps` argument now accepts string components or `Species` objects via `unpack_components`. Before this change, only `Species` objects were allowed.
- `calculate` and `equilibrium`
  - Accept the `phase_records` keyword argument 
  - The `callables` keyword argument is used if the phase records are constructed. If the `phase_records` argument is supplied by the caller, the `callables` argument is ignored.
  - `equilibrium` passes `phase_records` down to `calculate` instead of rebuilding them in `calculate`
  - Both functions check that all the active phases have entries in `phase_records` and raise an error if not. We currently don't check that the `output` is correct at all
- This approach is relatively simple, but one gap is that there's no mechanism to pass `phase_records` in to `equilibrium` that get re-used by `_eqcalculate`. `calculate` will always build `phase_records` when called from `_eqcalculate` (the `callables` can still be passed down for now).

### DIFF
--- a/pycalphad/tests/test_calculate.py
+++ b/pycalphad/tests/test_calculate.py
@@ -118,7 +118,7 @@ def test_missing_phase_records_passed_to_calculate_raises():
 
     models = instantiate_models(ALFE_DBF, comps, subset_phases)
     # Dummy conditions are just needed to get all the state variables into the PhaseRecord objects
-    phase_records = build_phase_records(ALFE_DBF, comps, subset_phases, {v.N: None, v.P: None, v.T: None}, models)
+    phase_records = build_phase_records(ALFE_DBF, comps, subset_phases, [v.T, v.P, v.N], models)
 
     with pytest.raises(ValueError):
         calculate(ALFE_DBF, comps, my_phases, T=1200, P=101325, N=1, phase_records=phase_records)
--- a/pycalphad/tests/test_calculate.py
+++ b/pycalphad/tests/test_calculate.py
@@ -4,9 +4,11 @@ Model quantities correctly.
 """
 
 import pytest
-from pycalphad import Database, calculate, Model
+from pycalphad import Database, calculate, Model, variables as v
 import numpy as np
 from numpy.testing import assert_allclose
+from pycalphad.codegen.callables import build_phase_records
+from pycalphad.core.utils import instantiate_models
 from pycalphad import ConditionError
 from pycalphad.tests.datasets import ALCRNI_TDB as TDB_TEST_STRING, ALFE_TDB, CUMG_PARAMETERS_TDB
 
@@ -106,3 +108,17 @@ def test_single_model_instance_raises():
     mod = Model(DBF, comps, 'L12_FCC')  # Model instance does not match the phase
     with pytest.raises(ValueError):
         calculate(DBF, comps, ['LIQUID', 'L12_FCC'], T=1400.0, output='_fail_', model=mod)
+
+
+def test_missing_phase_records_passed_to_calculate_raises():
+    "calculate should raise an error if all the active phases are not included in the phase_records"
+    my_phases = ['LIQUID', 'FCC_A1']
+    subset_phases = ['FCC_A1']
+    comps = ['AL', 'FE', 'VA']
+
+    models = instantiate_models(ALFE_DBF, comps, subset_phases)
+    # Dummy conditions are just needed to get all the state variables into the PhaseRecord objects
+    phase_records = build_phase_records(ALFE_DBF, comps, subset_phases, {v.N: None, v.P: None, v.T: None}, models)
+
+    with pytest.raises(ValueError):
+        calculate(ALFE_DBF, comps, my_phases, T=1200, P=101325, N=1, phase_records=phase_records)
--- a/pycalphad/tests/test_equilibrium.py
+++ b/pycalphad/tests/test_equilibrium.py
@@ -10,9 +10,9 @@ from sympy import Symbol
 from numpy.testing import assert_allclose
 import numpy as np
 from pycalphad import Database, Model, calculate, equilibrium, EquilibriumError, ConditionError
-from pycalphad.codegen.callables import build_callables
+from pycalphad.codegen.callables import build_callables, build_phase_records
 from pycalphad.core.solver import SolverBase, Solver
-from pycalphad.core.utils import get_state_variables
+from pycalphad.core.utils import get_state_variables, instantiate_models
 import pycalphad.variables as v
 from pycalphad.tests.datasets import *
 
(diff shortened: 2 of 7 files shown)


bocklund:
Calling `equilibrium` in tight loops could be really expensive because the time to continually rebuild the phase records. After making the changes in this PR, I was able to call `build_phase_records` in the scheil package and pass them to equilibrium and see huge boost in performance, see these profiles of a single simulation. The benchmark images below show a reduction from 491 seconds -> 8 seconds by eliminating the redundant calls to `build_phase_records` every time `equilibrium` is called:


## Before: passing the callables from `build_callables` to `equilibrium` (491 seconds, 95% of time in `build_phase_records` called 1204 times)

image

## After: passing the phase records from `build_phase_records` to `equilibrium` (8.6 seconds, 0.9 seconds of time in `build_phase_records` called one time)

image


## Why didn't we do this before?

Before the new minimizer was implemented, the choice of external conditions was tightly coupled with `PhaseRecord` objects via `Model.get_multiphase_constraints`. That is, changing the conditions for an equilibrium calculation in an A-B binary system from `N`, `P`, `T`, `X(B)` to `N`, `P`, `T`, `X(A)` required a new set of phase records. This tight coupling was what led to passing around `callables`, which only depended on the state variables and not the external conditions. Since the conditions in the new minimizer are implemented inside the minimizer itself, each `PhaseRecord` now only needs to know about the state variables and the internal constraints for that phase.

pycalphad/tests/test_equilibrium.py
```diff
@@ -53,6 +53,39 @@ def test_eq_binary():
...
+    "Pre-built phase records can be passed to equilibrium."
+    my_phases = ['LIQUID', 'FCC_A1', 'HCP_A3', 'AL5FE2', 'AL2FE', 'AL13FE4', 'AL5FE4']
+    comps = ['AL', 'FE', 'VA']
+    conds = {v.T: 1400, v.P: 101325, v.N: 1.0, v.X('AL'): 0.55}
+
+    models = instantiate_models(ALFE_DBF, comps, my_phases)
+    phase_records = build_phase_records(ALFE_DBF, comps, my_phases, conds, models)
+
+    # Without models passed
+    eqx = equilibrium(ALFE_DBF, comps, my_phases, conds, verbose=True, phase_records=phase_records)
```
richardotis, line 67: I don't think the models being rebuilt when phase records are passed in is good behavior, from a least-surprise perspective. This could lead to surprising, non-intuitive results for custom models. I think either (a) `phase_records` being passed in should require `models` being passed in, or (b) somehow the `models` should be inferred from the `phase_records` without being rebuilt.

pycalphad/tests/test_calculate.py
```diff
@@ -106,3 +108,17 @@ def test_single_model_instance_raises():
...
+
+def test_missing_phase_records_passed_to_calculate_raises():
+    "calculate should raise an error if all the active phases are not included in the phase_records"
+    my_phases = ['LIQUID', 'FCC_A1']
+    subset_phases = ['FCC_A1']
+    comps = ['AL', 'FE', 'VA']
+
+    models = instantiate_models(ALFE_DBF, comps, subset_phases)
+    # Dummy conditions are just needed to get all the state variables into the PhaseRecord objects
+    phase_records = build_phase_records(ALFE_DBF, comps, subset_phases, {v.N: None, v.P: None, v.T: None}, models)
```
richardotis, line 121: `{v.StateVariable: None, ...}` makes me think the function signature of `build_phase_records` should change to accept the state variables as a list or set instead. The dict could be kept for backwards compatibility (temporarily) since it's just 1-2 lines to extract the keys. Or do we want the flexibility to backtrack on multi-phase conditions down the line?

richardotis (changes requested):

pycalphad/tests/test_calculate.py
bocklund: I'm in favor of accepting a sequence of state variables here. Our next release also changes some of the internal solver APIs, so we might as well feel free to change others as well.

pycalphad/tests/test_equilibrium.py
bocklund: I agree. I'm leaning towards (a) because I don't want to enforce coupling between phase records and models. 

One complication is that the `model` argument can be unpacked via `unpack_kwarg` and it can have many forms as input, so it may be tricky to validate.  Some options of how to handle it in the case that `phase_records` is passed:

1. Do not `instantiate_models`, but make a best effort check that `all(isinstance(mod, Model) for mod in model.values())`. This alone wouldn't catch the case where the `model` argument only overrides a subset of active phases (and `instantiate_models` would typically build the rest), but we could address that in a similar way.
2. Do not instantiate models. Don't do any validation and just pass whatever the `model` argument was to `build_phase_records`, letting `build_phase_records` and `build_callables` raise appropriately.

richardotis: I think the first option is probably reasonable and likely to give good error messages.

bocklund: Here's the commit that adds code to handle this behavior in the spirit of the first option: https://github.com/pycalphad/pycalphad/pull/361/commits/a13cbcc6a3453ba415a1fee3ea72b07c137fae70

It's safer, but doesn't really protect users from providing incompatible model instances. The user is still responsible for making sure their argument to `model` is actually correct. Model objects really only serve two purposes in the `equilibrium` code path:

1. building phase records
2. getting the size of the internal degrees of freedom and the indices of pure vacancy endmembers for point sampling

if we could somehow do the second one without a Model instance, we could make it so providing a phase record or model are mutually exclusive per phase.

richardotis: I think this approach is fine for the time being.

pycalphad/tests/test_calculate.py
bocklund: I changed the name of the `conds` argument of `build_phase_records` to `state_variables`, so that covers the semantic change.

For the implementation, I made a [slight modification](https://github.com/pycalphad/pycalphad/pull/361/commits/854cf4e4951e7596d1f6a20d2bdc55253ebb6ca8#diff-47b2bfb2462e33b519b7489e73ff5bac523728463855804c4d40ea7ced4bb26e) to `get_state_variables` to loop over `conds` directly instead of accessing `conds.keys()`. With this modification, we can take advantage of the fact that iterating over Python dictionaries means iterating over the keys. Thus, passing dictionaries of conditions to `state_variables` is backwards compatible and passing any iterable (list, set, etc.) of state variables also works.

richardotis (approved):

PR #362, opened 2021-07-31
ENH: Add metastable composition sets to starting point

richardotis:
This PR adds a new function, `add_nearly_stable`, which augments the starting point of the solver with additional `CompositionSet` objects, which are entered with zero amount. These objects are added by the following criteria:
- The phase does not already exist on the starting hyperplane (this will not add missing composition sets of a miscibility gap)
- For a given phase, the added configuration is the closest to the starting hyperplane, within 1000 J/mol-atom

The purpose of this change is to improve the accuracy of solutions when sampling densities are not high enough to find the true set of equilibrium composition sets. This can easily happen in the multi-component / multi-sublattice case.


pycalphad/core/eqsolver.pyx
```diff
@@ -235,6 +278,8 @@ def _solve_eq_at_conditions(properties, phase_records, grid, conds_keys, state_v
             composition_sets.append(compset)
         chemical_potentials = prop_MU_values[it.multi_index]
         energy = prop_GM_values[it.multi_index]
+        add_nearly_stable(composition_sets, phase_records, grid, curr_idx, chemical_potentials,
```
bocklund, line 281: Does it make sense that the nearly stable composition sets have one shot to make it into the solution? After the first call to `find_solution`, all the remaining metastable composition sets will be removed if `find_solution` converged and the nearly metastable composition sets won't be able to make it in.

Since we have more metastable composition sets in the solution, does it make sense to remove the metastable composition sets less aggressively after `find_solution`, waiting until we think we found the global min to remove the metastable candidates?

bocklund: Following up on the previous comment, can we replace `add_new_phases` by `add_nearly_stable` if we slightly modify the behavior of `add_nearly_stable` to add a small amount of a phase with positive driving force (like the 1e-6 we use in `add_new_phases`)?

pycalphad/core/eqsolver.pyx
```diff
@@ -82,6 +81,50 @@ cdef bint add_new_phases(object composition_sets, object removed_compsets, objec
...
+    cdef double[:,::1] current_grid_X = grid.X[*current_idx, ...]
+    cdef double[::1] current_grid_GM = grid.GM[*current_idx, ...]
+    cdef unicode phase_name
+    cdef CompositionSet compset = composition_sets[0]
+    cdef set stable_phases = {compset.phase_record.phase_name for compset in composition_sets}
+    cdef PhaseRecord phase_record
+    cdef int num_statevars = len(compset.phase_record.state_variables)
+    cdef int df_idx, minimum_df_idx
+    cdef bint phases_added = False
+    driving_forces = np.dot(current_grid_X, chemical_potentials) - current_grid_GM
```
bocklund, line 109: Does it make more sense to allocate driving forces outside `add_nearly_stable` and pass the array to the `out` keyword argument of `np.dot`? For broadcasted equilibrium calculations, this will allocate an array of the same size as `grid.GM` for every set of conditions, but the size of the grid doesn't change.

Profiling would reveal if it's worth doing.

pycalphad/core/eqsolver.pyx
```diff
@@ -82,6 +81,50 @@ cdef bint add_new_phases(object composition_sets, object removed_compsets, objec
...
+    cdef int i
+    cdef int result = 0
+    cdef double highest = -1e30
+    for i in range(a_shape):
+        if a[i] > highest:
+            highest = a[i]
+            result = i
+    return result
+
+def add_nearly_stable(object composition_sets, dict phase_records,
```
bocklund, line 95: Does it make any difference for this to be `cdef` or `cpdef`?

bocklund (commented):
Does this need a test and what does that look like? Do we need a unit test that checks that composition sets are added or is an integration test for a tricky case enough?

pycalphad/core/eqsolver.pyx
richardotis: Including calls to `add_nearly_stable` inside the main solve loop seems to cause convergence issues for `test_eq_stepsize_reduction`. Setting `remove_metastable=False` inside the loop and putting a separate call to `update_composition_sets` at the end also causes a convergence failure in that test. I haven't fully run the issue down. Because the logic for how phases are changed is only a heuristic, it's possible to fail to come up with the stable set of phases in a finite number of iterations.

pycalphad/core/eqsolver.pyx
richardotis: It's only worth it if we call `add_nearly_stable` inside the solve loop, which I'm not sure that we should at this time.

pycalphad/core/eqsolver.pyx
richardotis: I don't have an explanation for this, but I see a 15-20% increase (regression) in the run-time of the test suite when I change this function to `cpdef`. The difference is noisy but repeatable.

richardotis:
> Does this need a test and what does that look like? Do we need a unit test that checks that composition sets are added or is an integration test for a tricky case enough?

I think an integration test for a test case is the right idea, because the use of metastable composition sets in the starting point is an implementation detail. Do you have a good test case? I hesitate to use the multi-component system that inspired this because of the additional test time.

bocklund (approved):
Thanks for putting together this PR!

Issue #359, opened 2021-06-21
Database parsing fails if some tokens are not uppercase


bocklund:
Here's a failing test that reproduces the issue that could be added to `test_database.py`:

```python
from pycalphad import Database
from pycalphad.tests.datasets import ALFE_TDB

def test_load_lowercase():
    "Test loading a database coerced to lowercase loads correctly."
    dbf = Database.from_string(ALFE_TDB, fmt='tdb')
    dbf_lower = Database.from_string(ALFE_TDB.lower(), fmt='tdb')
    assert dbf == dbf_lower
```

ingenious452:
Hey I am new to opensource and would like to contribute, could you explain what the issue is and what I have to do.

richardotis:
@ingenious452 To start, you can install the development version of pycalphad by following the instructions here: https://pycalphad.org/docs/latest/INSTALLING.html#development-version
Then you should try to run the above test code by putting the following into a separate Python script:
```python
from pycalphad import Database
from pycalphad.tests.datasets import ALFE_TDB

def test_load_lowercase():
    "Test loading a database coerced to lowercase loads correctly."
    dbf = Database.from_string(ALFE_TDB, fmt='tdb')
    dbf_lower = Database.from_string(ALFE_TDB.lower(), fmt='tdb')
    assert dbf == dbf_lower
    
test_load_lowercase()
```
When you run this script (`python mytestscript.py`), you should get an `AssertionError` as the output of the script. That's how you know you're reproducing the bug! Once you get to this step, we can talk about ways to fix the bug.

richardotis:
@ingenious452 Someone else fixed this issue already, but feel free to take a look at pycalphad's list of other open beginner bugs: https://github.com/pycalphad/pycalphad/labels/beginner

PR #364, opened 2021-08-01
FIX: Database parsing fails if some tokens are not uppercase

nandiniraja348:
Updated the database string to always be parsed to uppercase. Also passed the test cases mentioned in the issue #359


richardotis (approved):

richardotis:
@bocklund Any thoughts on merging this?

bocklund:
LGTM

PR #365, opened 2021-08-04
FIX: Improved custom state variable support

richardotis:
This PR makes two small improvements to better support the use of custom state variable objects.

One change silences a callables warning that would occur every time a custom state variable was used; this has been improved to only warn when expected state variables are _missing_. The other change relaxes an implicit assumption / bug that state variables would never be equal to zero (when calculating a percent change in value); now we detect this case and switch to an absolute difference.

### DIFF
--- a/pycalphad/core/minimizer.pyx
+++ b/pycalphad/core/minimizer.pyx
@@ -315,8 +315,10 @@ cdef void extract_equilibrium_solution(double[::1] chemical_potentials, double[:
     for i in range(free_statevar_indices.shape[0]):
         statevar_idx = free_statevar_indices[i]
         delta_statevars[statevar_idx] = equilibrium_soln[soln_index_offset + i]
-    for i in range(delta_statevars.shape[0]):
-        psc = abs(delta_statevars[i] / dof[0][i])
+        if dof[0][statevar_idx] == 0:
+            psc = np.inf
+        else:
+            psc = abs(delta_statevars[statevar_idx] / dof[0][statevar_idx])
         largest_statevar_change[0] = max(largest_statevar_change[0], psc)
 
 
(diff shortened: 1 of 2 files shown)


pycalphad/core/minimizer.pyx
```diff
@@ -316,7 +316,10 @@ cdef void extract_equilibrium_solution(double[::1] chemical_potentials, double[:
         statevar_idx = free_statevar_indices[i]
         delta_statevars[statevar_idx] = equilibrium_soln[soln_index_offset + i]
     for i in range(delta_statevars.shape[0]):
-        psc = abs(delta_statevars[i] / dof[0][i])
+        if dof[0][i] != 0:
+            psc = abs(delta_statevars[i] / dof[0][i])
+        else:
+            psc = abs(delta_statevars[i])
```
bocklund, line 322: It feels odd that the case where `dof[0][i] == 0` is now effectively measuring the absolute `delta_statevar`, while the previous `dof[0][i] != 0` case is the relative change.

I can see the reasoning behind having relative changes, since we don't necessarily know the scale of the state variable _a priori_. If we applied the relative logic, should the case where `dof[0][i] == 0` give `psc = np.inf` or a very large number that captures the spirit of the relative change metric?

richardotis: I'm not sure on the right answer here. It's not clear to me that poorly-scaled free state variables will be optimized properly with this or the current logic anyway. This loop will run even when all state variables are fixed.

richardotis: I've reworked this loop so it's only computed for free state variables, with a similar check for zero-values. Real free-state variable calculations where zero is the true solution should still mostly work, though I expect in most cases it will merely converge to a small non-zero value.

bocklund (approved):

PR #370, opened 2021-08-27
Create new axes with specified projection without using gca(...)

### LINKED ISSUE #369: deprecation warning in ternary example
The [ternary example](https://pycalphad.org/docs/latest/examples/TernaryExamples.html) warns

> `pycalphad/plot/eqplot.py:102: MatplotlibDeprecationWarning`:
> Calling `gca()` with keyword arguments was deprecated in Matplotlib 3.4. Starting two minor releases later, `gca()` will take no keyword arguments. The `gca()` function should only be used to get the current axes, or if no axes exist, create new axes with default keyword arguments. To create a new axes with non-default arguments, use `plt.axes()` or 
> `plt.subplot().ax = fig.gca(projection=projection)`.

This should be straightforward to fix on lines 101--102 since a new figure is generated, and I believe line 103 is a no-op? As it stands:

```python
100 >>> if ax is None:
101 >>>     fig = plt.figure()
102 >>>     ax = fig.gca(projection=projection)
103 >>> ax = plt.gca(projection=projection) if ax is None else ax
```

tkphd:
Closes #369.


richardotis:
@tkphd General isopleth calculation is on the roadmap and probably the biggest feature blocking "pycalphad 1.0". @bocklund has been thinking hard about general phase diagram mapping, including isopleth calculation. A recent paper from Sundman et al describes the challenges with the general case: https://doi.org/10.1016/j.calphad.2021.102330

In short, there's a lot of state to track for multicomponent isopleths, and it's easy to accidentally compute a metastable diagram. I can't promise a timeline except to say that this year's new energy minimization module (shipped in 0.9) was definitely a prerequisite to further progress, and I think we're in good position to ship "something" related to mapping in 2022. We are open to collaborations in this direction,

PR #375, opened 2021-09-22
MAINT: Bump to SymPy 1.9

bocklund:
Bump support for SymPy 1.9. A release candidate is currently out.

1. Change `__getnewargs__` to `__getnewargs_ex__`. SymPy Symbol subclasses now use `__getnewargs_ex__`, which broke our subclasses. See https://github.com/sympy/sympy/pull/21260. Since Python 3.6, `__getnewargs_ex__` has been used by Pickle protocol 2, which was introduced in Python 2.3. `__getnewargs_ex__` takes precedence over `__getnewargs__` if it is defined. See https://docs.python.org/3/library/pickle.html#object.__getnewargs_ex__.

Remaining items:

- [x] SymPy 1.9 released
- [ ] Fix deprecation of `expr_free_symbols` used in our patched `Piecewise.eval` https://github.com/sympy/sympy/issues/21494. Switching to `free_symbols` causes one test failure by introducing a `zoo`, but I think it could be trivially fixed. There's a significant performance regression in `subs` and `xreplace` that causes the test suite time to grow from ~2 minutes to ~5 minutes locally.


richardotis:
Thanks for being proactive with this.
> There's a significant performance regression in `subs` and `xreplace` that causes the test suite time to grow from ~2 minutes to ~5 minutes locally.
Maybe we can push back on upstream to investigate the cause before the final release.

bocklund:
I added a comment in https://github.com/sympy/sympy/issues/21494 with some details about our use case and the performance regression.

bocklund:
Oh, I think the example is this one: https://github.com/sympy/sympy/issues/14933#issuecomment-927170303

bocklund:
We'll skip SymPy 1.9 because of some performance regressions. SymPy>1.10 should address the performance issues or SymPy will be superseded by SymEngine via #376.

PR #389, opened 2021-12-20
ENH: Add support for the modified quasichemical model in the quadruplet approximation

bocklund:
## What's in this PR:

- Adds a `Database()` reader for the ChemSage `.dat` format. It supports reading the most common models that I have seen in the wild and have acquired databases for. Currently supported models include the modified quasichemical model in the quadruplet approximation (MQMQA), CEF models with Redlich-Kister-Muggianu mixing, stoichiometric phases (cast to stoichiometric CEF representations). Not yet supported are CEF models with the generalized Kohler-Toop mixing, real gases, aqueous phases, and others not mentioned previously. A ChemSage `.dat` writer is not implemented here.
- Add `pycalphad.models` subpackage for holding future models
- Implement the MQMQA in `pycalphad.models.model_mqmqa`
- Add support QKTO mixing to `Model` ("QKTO" is a CEF model that uses generalized polynomial mixing with Kohler or Kohler-Toop extrapolation)
- Add pytest fixture machinery to load database files from `pycalphad.tests.databases` 
- Slight modification of the `check_energy` test helper, where we test on the `GM` property, rather than `ast`. IMO the AST is not guaranteed to be a part of the Model API (hopefully we formalize this at some point with a protocol). In-fact the MQMQA has all contributions in terms of moles of formula such that `G=ast; GM=G/normalization` instead of `GM=ast; G=GM*normalization`, as in the Model class. The former should make for smaller, easier to debug and faster to compile functions.

## What's left for this PR:
- [x] Parser
  - [x] Do another pass through to try to clear out the remaining TODO items
  - [x] SUBQ zeta support in the pairs
  - [x] Cleanup nomenclature of the MQMQA to better match up with `ModelMQMQA` implementation (i.e. "Endmembers" vs. "Pairs")
- [x]  `ModelMQMQA`
  - [x] Adding tests for anion mixing and reciprocal quadruplets
  - [x] Testing and supporting MQMQA models with different "zeta" and SUBQ type
  - [x] Consolidate/clean up the tinydb `constituent_array` tests

## Open questions, looking for feedback:
- [x] Is there any more work that _needs_ to go in to the DAT reader for this PR (that can't be deferred)?
- [x] changes in `pycalphad.io.grammar` 
- [x] generally, the approach of having mangled quadruplet species for site fractions -- In the MQMQA cation and anion species names are mangled (by adding the charge) for MQMQA to be able to disambiguate quadruplets that use the same pure element of different valence states (e.g. Fe+2 and Fe+3) when they are put into a fictitious quadruplet species. Without the name mangling a quadruplet `v.Species` object would not be able to differentiate Fe+2Fe+2XX, Fe+2Fe+3XX, and Fe+3Fe+3XX quadruplets.
- [x] In the DAT parser, I write some code to workaround assumptions in the `Database.add_parameter` method that assumes that the parameter form will be CEF-like, as we have been using. I could use feedback on whether we should modify the `add_parameter` API or continue using this workaround for now.
- [x] I removed TDB read/write support for the older style of MQMQA parameters, so there's no officially supported way to get the MQMQA model out. Do we need basic XML or alt-database support for this PR?

## Things I'd want to defer:
- Implementing other models (~e.g. generalized polynomial Kohler-Toop in CEF models~ some QKTO support added) (these are the key missing aspects in coverage of the parser: other models and representations of the energy: heat capacity coefficients, magnetic parameters, and other models (real gas, Pitzer, etc.)
~- Any conflicts with SymEngine readiness in the parser or model (#376)~ (symengine support added)


richardotis:
Tremendous work by the team. XML is probably the nearest term solution for MQMQA database serialization. Before I advise on whether XML is blocking this PR, it'd be good if you could assess this WIP branch for an XML database schema to comment on what revisions would need to be made to bring MQMQA parsing support into consistency with this work: https://github.com/richardotis/database-concepts/blob/main/xml2/database.rng
I could do most of the "plumbing" on the parser side but I'd prefer someone with in-depth knowledge on MQMQA to draft the relevant portion of the schema.

bocklund:
For specifically the MQMQA aspects, there's basically three new parameter types in the database (`MQMG` (pair energies), `MQMX` (excess energies), and `MQMZ` (coordination #s)) and some new model hints. In the `cs_dat` parser, each new parameter type has a `_database_add_parameter` method to add itself to a database. Mostly they look like existing parameters and the keys that overlap with our existing RK parameter names have the same meaning, but each has a little bit of new data (comments).

For `MQMG` (`SUBQPair`):

```python
        new_parameter = {
            'phase_name': phase_name,
            'constituent_array': tuple(tuple(species_dict.get(s.upper(), v.Species(s)) for s in xs) for xs in constituent_array),
            'parameter_type': param_type,
            'parameter': parameter,
            'zeta': zeta,  # floating for the value of zeta
            'stoichiometry': stoichiometry,  # list of floats
            'reference': ref,
        }
```

For `MQMZ` (`SUBQQuadrupletCoordinations`):

```python
        new_parameter = {
            'phase_name': phase_name,
            'constituent_array': tuple(tuple(species_dict.get(s.upper(), v.Species(s)) for s in xs) for xs in constituent_array),
            'parameter_type': param_type,
            'coordinations': coordinations,  # list of 4 floats for coordination numbers
            'reference': ref,
        }
```

For `MQMX` (`SUBQExcessQuadruplet`):

```python
        new_parameter = {
            'phase_name': phase_name,
            'constituent_array': tuple(tuple(species_dict.get(s.upper(), v.Species(s)) for s in xs) for xs in constituent_array),  # must be hashable type
            'parameter_type': param_type,
            'exponents': exponents,  # list of 4 exponents (one for each species in a quadruplet)
            'parameter': param,
            'additional_mixing_constituent': additional_mixing_constituent,  # Species
            'additional_mixing_exponent': additional_mixing_exponent,  # An exponent
            "mixing_code" : mixing_code,  # String for the type of mixing, e.g. "G", "Q", or "B"
            'reference': ref,
        }
```

model hints:

```python
        model_hints = {
            "mqmqa": {
                "type": self.phase_type,  # either SUBG or SUBQ for the two different versions of the MQMQA
                "chemical_groups": {
                    "cations": dict(zip(map(_species, cation_el_chg_pairs), self.subl_1_chemical_groups)),  # mapping of species to an integer ("chemical group") for each cation species
                    "anions": dict(zip(map(_species, anion_el_chg_pairs), self.subl_2_chemical_groups)),  # same for anions
                }
            }
        }
```

richardotis:
After chatting offline, I agree with deferring XML support to the future.

bocklund:
Some updates since the last comments in January:

1. SymEngine support is merged in.
2. MQMQA implementation is complete to my knowledge. We're in complete agreement with Thermochimica for the collection of databases I have
    - I don't have a database that has 5 component SUBQ model, (A, B)(X, Y, Z) or (A, B, C)(X, Y), which would be needed to test the SUBQ modification to Chi mixing works. The term passes coverage but is technically not tested.
    - Reciprocal quadruplet excess terms are not explicitly tested because I don't have a suitable database, but the term is super simple (there's no "mixing" contribution: `mixing_term = S.One`) and I tested a modified database against Thermochimica
4. QKTO support is added, but only handles one sublattice currently, but that matches the available implementation in DAT files
3. DAT files allow for giving an H298, S298, and Cp polynomial terms to construct the Gibbs energy. When there are multiple temperature intervals, each new interval needs to provide a H_transformation and S_transformation as well as the Cp polynomial terms. Two issues here:
    - I had some rudimentary support for these types of terms with SymPy using `integrate`, but SymEngine does not provide an `integrate` implementation. For the DAT-prescribe polynomial terms (`CP_TERMS = (S.Zero, S.One, v.T, v.T**2, v.T**(-2))`), we should be able to hardcode analytical integrals of the Cp temperature terms (particularly where the power rule does not apply, e.g. the  `S.One` term that gives a `T**(-1)` term in the entropy integral) as they relate to Gibbs energy. The DAT file also allows users to provide their own coefficient-exponent pairs that we won't be able to hardcode integrals for, but the exponents will be a real number so we should be able to apply the power rule on the exponents and "evaluate" the integrals symbolically by plugging in the integration limits to the hardcoded+power-rule terms. So all this is theoretically simple to do right now, just a little bit of a pain.
    - The second issue is that when there are multiple temperature limits, we need to be more careful and pedantic about doing the integration. The way the code is written right now, the energy contribution for each temperature interval is computed in isolation. For the Cp-type parameters in the DAT file, each interval would have to know about all the preceding intervals to be able to integrate properly from 298 K up until the temperatures in the current interval. Some things may have to be re-designed a little regarding how piecewise Gibbs energy expressions are constructed, which is why I haven't implemented this yet. I have two databases locally that use Cp terms, but neither use MQMQA for the liquid and I'm not sure how prevalent Cp terms are in databases in the wild. It should always be possible (but maybe unpleasant) to manually re-write a DAT file to not use Cp terms.

I think this is now ready for review (likely in multiple review passes 😄)

pycalphad/io/cs_dat.py
```diff
@@ -0,0 +1,1152 @@
...
+@dataclass
+class Phase_Stoichiometric(PhaseBase):
+    magnetic_afm_factor: Optional[float]
+    magnetic_structure_factor: Optional[float]
+
+    def insert(self, dbf: Database, pure_elements: List[str], gibbs_coefficient_idxs: List[int], excess_coefficient_idxs: List[int]):
+        model_hints = {}
+        if self.magnetic_afm_factor is not None or self.magnetic_structure_factor is not None:
+            # This follows the Redlich-Kister Muggianu IHJ model. The ChemSage
+            # docs don't incidate that it's an IHJ model, but Eriksson and Hack,
```
richardotis, line 404: 'indicate'

pycalphad/io/cs_dat.py
```diff
@@ -0,0 +1,1152 @@
...
+    endmember_constituent_idxs: List[List[int]]
+    excess_parameters: List[ExcessBase]
+    magnetic_afm_factor: Optional[float]
+    magnetic_structure_factor: Optional[float]
+
+    def insert(self, dbf: Database, pure_elements: List[str], gibbs_coefficient_idxs: List[int], excess_coefficient_idxs: List[int]):
+        model_hints = {}
+        if self.magnetic_afm_factor is not None and self.magnetic_structure_factor is not None:
+            # This follows the Redlich-Kister Muggianu IHJ model. The ChemSage
+            # docs don't incidate that it's an IHJ model, but Eriksson and Hack,
```
richardotis, line 440: 'indicate'

pycalphad/io/cs_dat.py
```diff
@@ -0,0 +1,1152 @@
...
+        formula = formula[:match.start()]
+        charge = int(f'{match.groups()[1]}{match.groups()[0] or 1}')
+    else:
+        charge = 0
+    # assumes that the remaining formula is a pure element
+    constituents = dict(parse_chemical_formula(formula)[0])
+    return v.Species(name, constituents=constituents, charge=charge)
+
+
+class TokenParser(deque):
```
richardotis, line 38: Nice use of deque here.

pycalphad/io/cs_dat.py
```diff
@@ -0,0 +1,1152 @@
...
+    subl_2_pair_idx = toks.parseN(num_subl_1_const*num_subl_2_const, int)
+    subl_const_idx_pairs = [(s1i, s2i) for s1i, s2i in zip(subl_1_pair_idx, subl_2_pair_idx)]
+    quadruplets = [parse_quadruplet(toks) for _ in range(num_quadruplets)]
+    excess_parameters = []
+    while True:
+        mixing_type = toks.parse(int)
+        if mixing_type == 0:
+            break
+        elif mixing_type == -9:
+            # some garbage, like 1 2 3K 1 2K 1 3K 2 3 6, 90 of them
```
richardotis, line 909: Can you explain this code path? Tests don't hit it. Is this another TODO for the future?

pycalphad/io/cs_dat.py
```diff
@@ -0,0 +1,1152 @@
...
+        # a two sublattice phase with (4, 1) constituents
+        # we want to invert this so each endmember is a pair, i.e.
+        # endmember_constituents = [[1, 1], [2, 1], [3, 1], [4, 1]]
+        endmember_constituent_idxs = list(zip(*endmember_constituent_idxs))
+    elif sanitized_phase_type in ('IDMX', 'RKMP', 'QKTO', 'PITZ'):
+        subl_ratios = [1.0]
+        constituent_array = None  # implictly defined by the endmembers
+        endmember_constituent_idxs = None
+    else:
+        raise NotImplemented(f"Phase type {phase_type} does not have method defined for determing the sublattice ratios")
```
richardotis, line 1028: should raise `NotImplementedError` here

pycalphad/io/grammar.py
```diff
@@ -6,12 +6,12 @@
...
 pos_neg_int_number = Word('+-' + nums).setParseAction(lambda t: [int(t[0])])  # '+3' or '-2' are examples
 # matching float w/ regex is ugly but is recommended by pyparsing
-regex_after_decimal = r'([0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?)'
+regex_after_decimal = r'([-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?)'
 float_number = Regex(r'[-+]?([0-9]+\.(?!([0-9]|[eE])))|{0}'.format(regex_after_decimal)) \
     .setParseAction(lambda t: [float(t[0])])
 
 chemical_formula = Group(OneOrMore(Word(alphas, min=1, max=2) + Optional(float_number, default=1.0))) + \
-                   Optional(Suppress('/') + pos_neg_int_number, default=0)
+                   Optional(Suppress('/') + float_number, default=0)
```
richardotis, line 14: Does your out-of-suite set of TDBs still parse the same with this grammar change?

pycalphad/tests/fixtures.py
```diff
@@ -0,0 +1,30 @@
...
+    Decorator to facilitate safe, fast loading of database objects. Use as
+
+    ```
+    @select_database(\"filename.tdb\")  # matches a file in the pycalphad.test.databases directory
+    def test_name_of_my_test(load_database):
+        dbf = load_database()  # equivalent to `dbf = Database("filename.tdb")
+        # ... implement test below
+    ```
+    """
+    return pytest.mark.parametrize("load_database", [path], indirect=True)
```
richardotis, line 30: Needs a newline at the end of the file

pycalphad/tests/test_equilibrium.py
```diff
@@ -703,6 +704,207 @@ def test_eq_associate():
     eq = equilibrium(dbf, ['A', 'Q'], phases, conds)
     assert_allclose(eq.GM.values, -1736.981311)
 
+@select_database("Ocadiz-Flores.dat")
+def test_MQMQA_energy_K_F_NI(load_database):
+    dbf = load_database()
+    comps = ['K', 'F', 'NI']
+    model = {'LIQUID2': ModelMQMQA}
```
richardotis, line 711: Anything we can do to avoid requiring users to explicitly pass `ModelMQMQA`?

setup.py
```diff
@@ -54,6 +55,7 @@ def read(fname):
         # gives the C++ SymEngine library, while conda-forge/python-symengine
         # provides the Python package called `symengine`.
         'importlib_metadata',  # drop when pycalphad drops support for Python<3.8
+        'importlib_resources',  # drop when pycalphad drops support for Python<3.9
```
richardotis, line 58: Should also be added to `requirements-dev.txt`

pycalphad/io/cs_dat.py
bocklund: I have no idea what these mean and wrote this code really early in the parser implementation. The meaning isn't documented in the ChemApp documentation that I can tell. 

I have a database for Fe-Ti-V-O that ends a `SUBQ` block with 

```
   3
 G   1   4   6   6   0   1   0   0
 0.00000000       1.00 0.00000000       1.00 0.00000000       1.00
 0.00000000       0.00 0.00000000       0.00 0.00000000       0.00
   0   0 0.00000000     0.00000000     0.00000000     0.00000000
 0.00000000     0.00000000
  -9
    1    2    3K        1    2K        1    3K        2    3    6
    1    2    4K        1    2K        1    4K        2    4    6
    1    2    5K        1    2K        1    5K        2    5    6
    1    3    4K        1    3K        1    4K        3    4    6
    1    3    5K        1    3K        1    5K        3    5    6
    1    4    5K        1    4K        1    5K        4    5    6
    2    3    4K        2    3K        2    4K        3    4    6
    2    4    5K        2    4K        2    5K        4    5    6
    3    4    5K        3    4K        3    5K        4    5    6
```

The database from Starykh et al. Calphad 74 (2021) (https://doi.org/10.1016/j.calphad.2021.102312) ends a `SUBQ` block with:

```
 G   1   2   5   5   0   2   0   0
     0.00000000   1.00     0.00000000   1.00     0.00000000   1.00
     0.00000000   0.00     0.00000000   0.00     0.00000000   0.00
   0   0     0.00000000     0.00000000     0.00000000     0.00000000
     0.00000000     0.00000000
  -1
    3    1    4T3       3    1K        3    4T4       1    4    5
```

So at least I think we can infer that there's some special data here and `-N` means that there are `N` instances/lines worth of these special data.

Maybe @jorgepazsoldanpalma or @maxposchmann know what these mean?

pycalphad/tests/test_energy.py
```diff
@@ -665,3 +667,570 @@ def test_order_disorder_magnetic_ordering():
...
+                }
+
+    check_energy(mod, subs_dict, -1.47867E+05, mode="sympy")
+    assert np.isclose(float(mod.moles("CU").subs(subs_dict)), 0.07692307692,1e-5)
+    assert np.isclose(float(mod.moles("CL").subs(subs_dict)), 0.6923076923,1e-5)
+    assert np.isclose(float(mod.moles("ZN").subs(subs_dict)), 0.07692307692,1e-5)
+    assert np.isclose(float(mod.moles("FE").subs(subs_dict)), 0.15384615384,1e-5)
+
+
+# TODO: remove this test and the database once real version (below) works
```
richardotis, line 709: Should this test now be removed?

pycalphad/models/model_mqmqa.py
```diff
@@ -0,0 +1,761 @@
...
+
+    Canonicalizes the Species by sorting among the cation and anion sublattices.
+
+    Parameters
+    ----------
+    constituents : List[Tuple[str, v.Species]]
+
+    Examples
+    --------
+        get_species('A_1.0', 'B_1.0')
```
richardotis, line 25: Example has the wrong number of arguments

pycalphad/io/grammar.py
bocklund: There's no TDB support for any custom TDB syntax in this PR, so this can probably be safely reverted. 

I guess the question is whether it's valid to have floating point charges. My understanding is that the charges are treated like pure elements of `/-` in TDB `SPECIES`, so are floating point charges just as valid as floating point amounts of other pure elements? 

Either way, it's out of scope for this PR, so I'll revert.

richardotis:
Re: `_database_add_parameter` static method: Can we change `Database.add_parameter` to accept an additional dictionary as a keyword argument? Would that eliminate the need for a custom method?

richardotis:
Re: name mangling, it's not ideal from a semantic perspective, but the alternative would require the Species object and a lot of dependent code to be reworked. I think this approach is the least-bad solution.

richardotis (approved):
My review is complete. Pending your responses to the remaining items, this can be merged. The coverage issue is a minor concern but appears to mostly be due to implementation stubs for additional DAT parsing support, which is okay to defer.

pycalphad/tests/test_equilibrium.py
bocklund: When I was working on the `ModelProtocol` implementation a while ago I came up with a `ModelDispatcher` class (a function would be fine too) that has the same API as `Model.__init__` that dispatches on the correct class depending on `model_hints`. A possible solution would be to have this class be the default argument for `unpack_kwarg` for the `"model"` kwarg. 

For context, I ultimately decided against this kind of solution because I thought it would be easier for users to be comfortable with passing non-`Model` models explicitly. The two options I came up with for being able to hook into this dispatcher weren't very good: 
1. Users register their custom models with pycalphad to be able to be picked up by the dispatcher, which is kind of an annoying gotcha and requires keeping state
2. The `ModelProtocol` that users would subclass uses metaclass magic to automatically register the class, so any subclass would be transparently registered for users, but this also created some state and seemed like too much magic and clunkiness.

Some other options (that, IMO, are worse than the manual or dispatcher approaches):
- Add logic to `calculate` or `equilibrium` to pick the right class based on `model_hints`. I'd be worried about implementing this in a way that doesn't override user-specified models - a solution that worked around this issue would probably just look like the dispatcher approach.
- Add some logic in the `Model` initializer to instantiate the `ModelMQMQA` class instead. I think it would be bad form to give a caller of `Model` a different class than the one they asked for.

setup.py
bocklund: Currently `requirements-dev.txt` only has the _additional_ dependencies required for building the package and building the documentation, the runtime requirements are not included and everything in the setup.py install requires is installed if users follow our development version install procedure. I don't think this is a build time requirement (otherwise our CI should be telling us that by needing this package in our pyproject.toml build system requirements).

pycalphad/tests/test_energy.py
bocklund: Removed.

setup.py
richardotis: Oh, okay.

pycalphad/tests/test_equilibrium.py
richardotis: What about something like

```python
class Model:
    def __new__(cls, *args, **kwargs):
        target_cls = cls.dispatch_on(*args, **kwargs)
        instance = object.__new__(target_cls)
        return instance

    @classmethod
    def dispatch_on(cls, dbe, comps, phase_name, parameters=None):
        phase = dbe.phases[phase_name.upper()]
        target_cls = cls
        if 'mqmqa' in phase.model_hints.keys():
            from pycalphad.models.model_mqmqa import ModelMQMQA
            target_cls = ModelMQMQA
        return target_cls

    def __getnewargs_ex__(self):
        return ((self._dbe, self.components, self.phase_name,), {'parameters': self._parameters_arg})
[snip]
```
When I test locally it does the correct thing (after I remove the model keywords from the MQMQA tests of `equilibrium`). If the user truly wanted to override the behavior for MQMQA, then in their subclass they can just write
```python
class MyModelSubclass(Model):
    @classmethod
    def dispatch_on(cls, *args, **kwargs):
        return cls
[snip]
```
This would still require them to pass the subclass in the model keyword argument. If they want ModelMQMQA to continue overriding for MQM phases, they would just not define `dispatch_on`, which seems like a sane default to me. It's not as fancy as a metaclass approach, but it's comparatively simple, preserves backwards compatibility, and provides good UX from a least-surprise perspective.

bocklund:
> Re: _database_add_parameter static method: Can we change Database.add_parameter to accept an additional dictionary as a keyword argument? Would that eliminate the need for a custom method?

I just had it accept `kwargs` instead, which seemed logical based on how the function works otherwise. This change is kind of nice, because it allows for users to develop their own parameter dictionaries for their own models that may need additional custom metadata. 

If the `kwargs` approach is acceptable, we should consider changing the API of `add_parameter` in the future so that the arguments match the names of the arguments in the parameter dictionaries (e.g. `param_order -> parameter_order`). And we should also consider thinking about which arguments are required and which are not. 

> Re: name mangling, it's not ideal from a semantic perspective, but the alternative would require the Species object and a lot of dependent code to be reworked. I think this approach is the least-bad solution.

Agreed. It's mostly a cosmetic issue for  the string representations, so it should be possible to revisit without significant breakage.

pycalphad/tests/test_equilibrium.py
bocklund: My nit with this way of doing it is that now `Model` has to know about all other models to get the magic to work and custom user models will never be able to hook in and get that magic without changing `Model` (and that's true for any dispatcher without registration functionality). All the approaches have tradeoffs.

I didn't want to commit to introducing any magic in this PR, because that sets the precedent for how users should expect to interact with new models. But my experience shows that you have good judgment about how APIs should look, so I'm happy to defer to whatever option you prefer 😄

pycalphad/io/cs_dat.py
maxposchmann: Those lines override the interpolation schemes selections that are normally determined by the chemical groups. We still haven't implemented this in Thermochimica, because it is a pain and not often used.

I agree with this:

    So at least I think we can infer that there's some special data here and -N means that there are N instances/lines worth of these special data.

The rest of the scheme I am not 100% sure about. I have some guesses though. For example, consider
`1    2    3K        1    2K        1    3K        2    3    6`
which I read as "for the 1-2-3 ternary, treat 1-2 as Kohler, 1-3 as Kohler, and 2-3 as Kohler". The 6 at then end indicates the constituent on the opposite sublattice.

Or in a case that uses Toop
`3    1    4T3       3    1K        3    4T4       1    4    5`
I think this means "for the 3-1-4 ternary, treat 3-1 as Toop with 3 as Toop component, 3-4 as Kohler, and 1-4 as Toop with 4 as Toop component", where now 5 is the anion constituent.

I guess this is necessary if the desired selection of interpolation schemes does not follow the chemical group rules.

pycalphad/io/cs_dat.py
```diff
@@ -530,61 +511,32 @@ class SUBQPair(Endmember):
...
-        if force_insert:
-            self._parameters.insert(new_parameter)
-        else:
-            self._parameter_queue.append(new_parameter)
-
     def insert(self, dbf: Database, phase_name: str, constituent_array: List[str], gibbs_coefficient_idxs: List[int]):
         # Here the constituent array should be the pair name using the corrected
         # names, i.e. CU1.0CL1.0
-        self._database_add_parameter(dbf, 'MQMG', phase_name, constituent_array, self.expr(gibbs_coefficient_idxs), self.zeta, self.stoichiometry_quadruplet, force_insert=False)
+        pass
```
richardotis, line 517: Unnecessary `pass` here

pycalphad/tests/test_equilibrium.py
richardotis: I'm also open to renaming to something like `_dispatch_on` to denote a private API, if that makes this approach easier to accept for now.

pycalphad/io/cs_dat.py
bocklund: Thanks, @maxposchmann. I am following along and that would seem to make sense. For the SUBQ cases where there is mixing in the cation sublattice and there is a summation over the anions (e.g. your 2021 paper Eq. 22), you are using the these relationships for the anion in the interaction parameter - i.e. anion `k`, not `x` or `y` (using your notation) - right?

@richardotis: I already pushed some changes so that we can parse successfully both cases, but we currently just throw away this metadata.

I think I have some idea on how to handle this in pycalphad. In the `ModelMQMQA._chemical_group_filter`, we could have an option argument (or some state stored) that contains "overrides" that use the specified relationships rather than the chemical group relationships with the behavior here. The painful part is that the filter function now needs to know what anions is being considered. Since we have some databases, but Thermochimica does not yet have this feature, we probably need FactSage to test and compare against if we want the true answer. Some options:

1. Implement in this PR and have someone check some test calculations against FactSage to add to the test suite
2. Defer implementing and raise an error if these overrides are parsed. The rationale for raising an error is that incorrect results should be impossible
3. Defer implementing and issue a warning if these overrides are parsed. The rationale for a warning is that these overrides only impact the calculations and extrapolation on some specific subsystems, so a user may want to do calculations in a subsystem where these overrides do not apply and our calculations will be correct.

richardotis: Defer and warn is fine. I think our DAT support is going to have to be considered preliminary at this point, given the lack of a real specification, but it's a whole lot better than nothing.

pycalphad/tests/test_equilibrium.py
richardotis: My proposed changes have been pushed to the branch in [caa1520](https://github.com/pycalphad/pycalphad/pull/389/commits/caa1520dbd5d11f2efbfb4ad59c599c90a9cf87d).

pycalphad/io/cs_dat.py
bocklund: Sounds good. I did some of the legwork to parse the overrides into a dictionary form, enough to be able to warn users which sets of elements may be incorrect in 0689134a5d8e1b278502aabe3ee9e9312ed3bfd2. For example:

- The string `1 2 3K 1 2K 1 3K 2 3 6` gives a warning:

```python
<snip>/pycalphad/pycalphad/io/cs_dat.py:721: UserWarning: Phase SLAG-LIQ overrides the ternary extrapolation models for the system ['CU1+', 'SI', 'SN4+', 'O']. Use caution as extrapolated energies may be incorrect.
```

- The string `3 1 4T3 3 1K 3 4T4 1 4 5` gives a warning:

```python
<snip>/pycalphad/pycalphad/io/cs_dat.py:721: UserWarning: Phase SLAGBSOLN overrides the ternary extrapolation models for the system ['FE[3+]', 'FE[2+]', 'V2O3', 'O']. Use caution as extrapolated energies may be incorrect.
```

bocklund:
@richardotis I think everything is addressed and this is ready to merge. Thanks for your review!

Huge thanks to @jorgepazsoldanpalma for leading the initial implementation and to @maxposchmann for all the guidance you provided along the way!

pycalphad/io/cs_dat.py
maxposchmann: I'm not sure I followed your question correctly. I'll set up an example that matches my understanding of what you asked, and you can let me know if I missed the mark.

Let's say we have cations 1, 2, 3, and anions 4, 5. I won't assign then chemical groups, but suppose we overrode them for {1,2,3,5} but not {1,2,3,4}. Your question is then, how do we compute the result of equation 22 for say {1,2/4,4}?

If so, I agree that this is not well-defined in my notation. I think your suggested approach of using the mixing as defined for cation k is likely correct. After all, given the deltas, we only accumulate into the sum for x or y equal to k.

bocklund:
Sorry, I made a few small changes to the `TokenParser` if you want to take one last look at the recent changes.

Now there is support for parsing tokens files line-by-line. The line-by-line parsing is completely transparent to `TokenParser` callers, but allows us to raise more informative error messages when there's a parsing failure. We're not able to do any detailed explanations, but it will at least now explain to users which line we were on when the parsing error was encountered. For example, some output looks like this:

```
pycalphad.io.cs_dat.TokenParserError: Error at line number 1805: ("invalid literal for int() with base 10: 'G'",) for line:
     G   1   6  17  17   0   1   0   0
```

In the database in question, there should be a `3` between lines 1804 and 180-5

```
   3                                                                          # line 1799
 G   1   6  17  17   0   0   0   0                                            # line 1800
     0.000000000   1.00     0.000000000   1.00     0.000000000   1.00         # line 1801
     0.000000000   0.00     0.000000000   0.00     0.000000000   0.00         # line 1802
   0   0 -2639.40000     0.921000000         0.000000000     0.000000000      # line 1803
     0.000000000     0.000000000                                              # line 1804
 G   1   6  17  17   0   1   0   0                                            # line 1805
     0.000000000   1.00     0.000000000   1.00     0.000000000   1.00         # line 1806
     0.000000000   0.00     0.000000000   0.00     0.000000000   0.00         # line 1807
   0   0  50.0000000         0.000000000     0.000000000     0.000000000      # line 1808
     0.000000000     0.000000000                                              # line 1809
   3                                                                          # line 1810
 G   6  13  17  17   0   0   0   0                                            # line 1811
     0.000000000   1.00     0.000000000   1.00     0.000000000   1.00         # line 1812
     0.000000000   0.00     0.000000000   0.00     0.000000000   0.00         # line 1813
   0   0 -12087.0000     -5.00000000         0.000000000     0.000000000      # line 1814
     0.000000000     0.000000000                                              # line 1815

```

It will still probably require someone experienced with DAT files to understand what causes the error, but hopefully it makes our lives easier as developers to be able to track down where errors are happening and if it's a pycalphad parsing issue or a real issue with the database.

PR #392, opened 2022-01-29
FIX: Fix TDB line length bugs

bocklund:
This fixes two bugs related to writing TDB files and reflowing lines:

1. Fixes a bug in `reflow_text` where lines that are too long and contain no valid linebreak characters would result in an infinite loop of finding the same `linebreak_idx` and emitting the same line until running out of memory. Now lines that cannot be reflowed to under the `linewidth` raises an error.
2. Adds a whitespace to the delimiter between species in the `CONSTITUENTS` line (it's now `, ` instead of `,`) so `reflow_text` can reflow the line at any element. This fixes a bug where phases with many constituents would trigger the infinite loop (and now error) from (1).

A test is added for each case.

### DIFF
--- a/pycalphad/io/tdb.py
+++ b/pycalphad/io/tdb.py
@@ -620,7 +620,7 @@ def reflow_text(text, linewidth=80):
     reflowed_text : str
     """
     lines = text.split("\n")
-    linebreak_chars = [" ", "$", ","]
+    linebreak_chars = [" ", "$"]
     output_lines = []
     for line in lines:
         if len(line) <= linewidth:
--- a/pycalphad/io/tdb.py
+++ b/pycalphad/io/tdb.py
@@ -619,18 +619,22 @@ def reflow_text(text, linewidth=80):
     -------
     reflowed_text : str
     """
-    ""
     lines = text.split("\n")
-    linebreak_chars = [" ", "$"]
+    linebreak_chars = [" ", "$", ","]
     output_lines = []
     for line in lines:
         if len(line) <= linewidth:
             output_lines.append(line)
         else:
+            if set(line).isdisjoint(linebreak_chars):
+                raise ValueError(f"Unable to reflow the following line of length {len(line)} below the maximum length of {linewidth}: \n{line}")
             while len(line) > linewidth:
-                linebreak_idx = linewidth-1
+                linebreak_idx = linewidth - 1
                 while line[linebreak_idx] not in linebreak_chars:
                     linebreak_idx -= 1
+                # Need to check 2 (rather than zero) because we prepend newlines with 2 characters
+                if linebreak_idx == 2:
+                    raise ValueError(f"Unable to reflow the following line of length {len(line)} below the maximum length of {linewidth}: \n{line}")
                 output_lines.append(line[:linebreak_idx])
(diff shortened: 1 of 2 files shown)


pycalphad/tests/test_database.py
```diff
@@ -684,4 +684,104 @@ def test_load_database_when_given_in_lowercase():
...
+    ELEMENT ZN   HCP_A3                    6.5390E+01  5.6568E+03  4.1631E+01 !
+    ELEMENT ZR   HCP_A3                    9.1224E+01  5.5663E+03  3.9181E+01 !
+
+    PHASE LIQUID % 1 1 !
+    CONSTITUENT LIQUID : AG,AL,AM,AS,AU,B,BA,BE,BI,C,CA,CD,CE,CO,CR,CS,CU,DY,ER,
+    EU,FE,GA,GD,GE,HF,HG,HO,IN,IR,K,LA,LI,LU,MG,MN,MO,N,NA,NB,ND,NI,NP,O,OS,P,PA,
+    PB,PD,PR,PT,PU,RB,RE,RH,RU,S,SB,SC,SE,SI,SM,SN,SR,TA,TB,TC,TE,TH,TI,TL,TM,U,V,
+    W,Y,YB,ZN,ZR : !
+    """
+    dbf = Database(TDB)
```
richardotis, line 786: Can we assert that the expected number of components is in the Database prior to `to_string`? I'm imagining a case where a parse error inadvertently drops the constituents broken over subsequent lines, and this test incorrectly passes.

richardotis (changes requested):

pycalphad/io/tdb.py
```diff
@@ -619,18 +619,22 @@ def reflow_text(text, linewidth=80):
...
+            if set(line).isdisjoint(linebreak_chars):
+                raise ValueError(f"Unable to reflow the following line of length {len(line)} below the maximum length of {linewidth}: \n{line}")
             while len(line) > linewidth:
-                linebreak_idx = linewidth-1
+                linebreak_idx = linewidth - 1
                 while line[linebreak_idx] not in linebreak_chars:
                     linebreak_idx -= 1
+                # Need to check 2 (rather than zero) because we prepend newlines with 2 characters
+                if linebreak_idx == 2:
+                    raise ValueError(f"Unable to reflow the following line of length {len(line)} below the maximum length of {linewidth}: \n{line}")
```
richardotis, line 637: It's minor, but if you wanted to get complete test coverage in the diff you could combine the `raise ValueError` into a single exception by setting `linebreak_idx = 0` on line 630 instead of raising, then change to `if linebreak_idx <= 2`

pycalphad/tests/test_database.py
bocklund: Good idea. I added some more rigorous checks.

pycalphad/io/tdb.py
bocklund: I think I captured the spirit of what you mean. We have to enforce `linebreak_idx > 0`, otherwise `linebreak_idx` will go negative until `linebreak_idx < -(len(line))`, which raises an `IndexError`.

richardotis (approved):

PR #397, opened 2022-02-20
FIX: Handle DIFFUSION and ZEROVOLUME_SPECIES Thermo-Calc commands.

mfrichtl:
Minor fix to prevent errors if users have DIFFUSION or ZEROVOLUME_SPECIES commands in a database file.

Checklist
No items in the checklist are relevant to this minor fix.


richardotis:
Thanks for the contribution, @mfrichtl ! Can you add a test to `pycalphad/tests/test_database.py` that fails without your changes, but passes with them? It just needs a minimal TDB that uses the added commands, something like:
```
def test_database_passes_with_diffusion_commands():
    "Diffusion commands do not raise (doesn't test implementation)"
    tdb_string = """
     ELEMENT /-   ELECTRON_GAS              0.0000E+00  0.0000E+00  0.0000E+00!
     ELEMENT VA   VACUUM                    0.0000E+00  0.0000E+00  0.0000E+00!
     ELEMENT O    1/2_MOLE_O2(G)            0.0000E+00  0.0000E+00  0.0000E+00!

     DIFFUSION [something that is valid syntax] !
     ZEROVOLUME_SPECIES [something that is valid syntax] !
     """

    dbf = Database.from_string(tdb_string, fmt='tdb')
```

mfrichtl:
@richardotis I believe I added an appropriate test per your example and pushed it to this PR. Also believe I resolved a conflict that popped up.

richardotis (approved):

PR #398, opened 2022-02-21
BLD/DOC: Minor fixes

richardotis:
- Change many sympy references in comments and code to symengine
- Add Python 3.10 to supported Python versions metadata in setup.py
- Bump cibuildwheel version (especially to ensure the in-tree builds work)

Checklist
  * [x] `pyproject.toml` (build requirements)
  * [x] `requirements-dev.txt` (build and development requirements)


bocklund:
Some more instances of SymPy:

```
examples/ChargedPhases.ipynb
170:    "# Plot site fractions by index and label them using the ._latex() feature of SymPy symbols\n",

pycalphad/core/calculate.py
453:    # If we don't do this, sympy will get confused during substitution

pycalphad/tests/test_codegen.py
2:Tests for code generation from SymPy/SymEngine objects.

docs/examples/ViscosityModel.rst
115:    import sympy

pycalphad/tests/test_database.py
70:    # Won't work until sympy/sympy#10560 is fixed to prevent precision loss
125:    """A map of old symbol names to new symbol names should propagate through symbol and parameter SymPy expressions"""
```

We should double check the examples still run without sympy in the environment

richardotis:
> We should double check the examples still run without sympy in the environment

I checked this before I merged the PR.

bocklund:
We need to regenerate then because the viscosity example still has a lingering `import sympy`

bocklund (approved):

PR #404, opened 2022-03-23
TST: Use `importlib_resources` for test databases

richardotis:
This PR completes work (started in #389) to move all non-trivial databases from string representations in `pycalphad/tests/datasets.py` to standalone files in the `pycalphad/tests/databases/` directory. This is done in a portable, platform-independent way using `importlib_resources`, which is a backport of the recent-ish `importlib.resources` module of the standard library.

There are a few reasons to do this:
- `datasets.py` was approaching 10000 lines of code, and was getting difficult to read. Arbitrary Python code could have inadvertently slipped in and be missed in the (sometimes) very large diffs. Standalone files will only be read by the parser, which is much harder to induce arbitrary code execution without getting an error.
- Only databases needed for a particular test will be loaded, so partial runs of the test suite will now be faster. Before, almost all databases were instantiated even when running a single test.

The new pattern is
```python
# Don't let your IDE fool you, you must import load_database here
from pycalphad.tests.fixtures import select_database, load_database

@select_database("a_database_in_the_databases_folder.tdb")
def test_my_function(load_database):
    dbf = load_database()
```

There are a few tests that either use multiple databases at once, or explicitly test the string representation of the database. For those cases, we've hacked together a solution with something like
```python
from importlib_resources import files
import pycalphad.tests.databases

def read_database_as_string(fname):
   return open(str(files(pycalphad.tests.databases).joinpath(fname)), "r").read()
```

Checklist
* [ ] The documentation examples have been regenerated if the Jupyter notebooks in the `examples/` have changed. To regenerate the documentation examples, run `jupyter nbconvert --to rst --output-dir=docs/examples examples/*.ipynb` from the top level directory)
* [ ] If any dependencies have changed, the changes are reflected in the
  * [ ] `setup.py` (runtime requirements)
  * [ ] `pyproject.toml` (build requirements)
  * [ ] `requirements-dev.txt` (build and development requirements)


bocklund (approved):
LGTM.

I also noticed that all the databases are used by the tests except `cfe_brosh.tdb` and `diffusion.tdb`.

PR #405, opened 2022-03-24
BUG: Changes to the ChemSage DAT parser to be more consistent with the TDB parser

bocklund:
- Floating point Gibbs energy exponents (for fixed T/P terms) for read-write equality comparison support
- Use Database.add_structure_entry
- Use phase "BLANK" instead of None (needs a string for type stability) and "BLANK" is consistent with how pycalphad treats TDBs
- Use a global default minimum temperature for piecewise intervals of 0.01 K, in agreement with how we handle TDBs
- Some DAT files have two temperature intervals in a row with the same upper limit. I am interpreting that as an implication that the function should extrapolate beyond that point, so we choose a default upper limit of 10000 K, in agreement with TDBs
- Fix a bug in magnetic parameters: `BMAG` should have been `BMAGN`.
- Improve error messages for `pycalphad.io.tdb.to_interval` that users may find more helpful when debugging.

### DIFF
--- a/pycalphad/io/cs_dat.py
+++ b/pycalphad/io/cs_dat.py
@@ -16,9 +16,10 @@ from .grammar import parse_chemical_formula
 # From ChemApp Documentation, section 11.1 "The format of a ChemApp data-file"
 # We use a leading zero term because the data file's indices are 1-indexed and
 # this prevents us from needing to shift the indicies.
-GIBBS_TERMS = (S.Zero, S.One, v.T, v.T*log(v.T), v.T**2, v.T**3, 1/v.T)
-CP_TERMS = (S.Zero, S.One, v.T, v.T**2, v.T**(-2))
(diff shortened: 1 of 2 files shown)


richardotis:
Related: I'm second-guessing the changes in #400 now. There's no physical reason the exponents have to be integer, right? I have it in my head from something a while back that non-integer exponents would cause imaginary derivatives, but I haven't been able to produce an example where that's the case. I'm wondering now how I got that mistaken impression, and if it's nonetheless important for some other reason.

richardotis:
I think I figured it out: consider functions of the form `f(x,y) = (x-y)**n`. When n is an integer, the derivative of f(x,y) with respect to x is real as long as x and y are real. When n is not an integer, the derivative can have a nonzero imaginary component. This would pose a problem for Redlich-Kister interaction terms, for example.

I don't think anything physically constrains pure-temperature terms from having non-integer exponents, though. The derivative of `x**n` is always real as long as x and n are real.

richardotis (approved):

PR #408, opened 2022-04-15
ENH: Compatibility fixes for pyodide

richardotis:
These are some small fixes which enable compatibility with the WebAssembly-based [pyodide](https://pyodide.org/en/stable/ ) Python distribution. The changes here will only affect environments where (a) SymEngine is compiled without LLVM support (like Pyodide); or (b) the platform is 32-bit.

All officially supported platforms for pycalphad have been 64-bit for a while, so both of these changes are backward compatible. In addition, they are also forward compatible if LLVM support for SymEngine in pyodide is enabled, or (unlikely in the short term) pyodide adds wasm64 support.

### DIFF
--- a/pycalphad/codegen/sympydiff_utils.py
+++ b/pycalphad/codegen/sympydiff_utils.py
@@ -29,13 +29,16 @@ functions or PhaseRecords. The following issues track this behavior:
 """
 from pycalphad.core.cache import cacheit
 from pycalphad.core.utils import wrap_symbol
-from symengine import sympify, lambdify, zoo, oo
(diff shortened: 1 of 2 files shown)


richardotis:
Note that the packaging ecosystem for pyodide is still emerging, so at the moment pycalphad _and all of its dependencies_ have to be cross-compiled specifically for the wasm32 platform in order to use with pyodide. That activity is currently ongoing at https://github.com/materialsgenomefoundation/mgf-dist-pyodide/

bocklund (approved):

PR #412, opened 2022-04-27
ENH: Enable option B and option F support for TDBs

### LINKED ISSUE #202: TDB symmetry options
It would be great if we could add support for the TDB four-sublattice symmetry options, called option "B" and option "F".

bocklund:
Move over `generate_symmetric_group` from https://github.com/PhasesResearchLab/ESPEI/pull/229 with tests.

Add Al-Fe database from Sundman et al. 2009 that uses option B and option F. The option B phase also has a non-option B alternative, so we can compare energies.

This PR closes #202.


bocklund:
This implements the symmetry options at the level of the database parameters. Maybe one thing to discuss is whether it makes more sense to do this operation at the level of a database or to only do it inside the model. Some considerations:
- with the current approach, `pycalphad-xml` also should have support added for generating parameters since presumably we'll have these phase options implemented there as well (we should IMO).
- Most model hints we treat inside models. Symmetry phase options are a special case because they correspond virtually to new parameters. However, we do have some precedent for handling the virtual symmetric parameters for ternary interactions inside the model at model-build-time.

richardotis:
I was prepared to argue against implementing the parameter augmentation at the Database level because I had concerns about round-tripping. However, I see that the approach here (marking the parameters with special metadata) makes it easy to preserve the round-trip property without intruding on Model definition. Also, it keeps Model objects from needing to know about symmetry.

One concern I have is with performance. Does a multicomponent 4SL BCC or FCC phase with the symmetry options blow up in time or memory? My intuition is 'no' but I'd be interested to know how it behaves with Bengt's MPEA database, for example.

richardotis:
Another question: Option B/F apply to all of a phase's model parameters (how it's implemented here), not just the energy, correct?

bocklund:
> One concern I have is with performance. Does a multicomponent 4SL BCC or FCC phase with the symmetry options blow up in time or memory? My intuition is 'no' but I'd be interested to know how it behaves with Bengt's MPEA database, for example.

Looking at databases I have access to, the biggest with symmetry options are `mpea-02b.tdb` (which has the most symmetry parameters) and `iron-04.tdb` (which has the most constituents in phases with symmetry options).

On average over 20 runs:

| Database  | # symmetry ordering parameters |  Time to load database | Time to add ordering parameters | % time adding parameters |  
| ------------- | ------------- | ------------- | ------------- | ------------- |
| `mpea-02b.tdb`  | 63 | 1.64s  | 0.17s | 11% | 
| `iron-04.tdb`  | 24  | 3.59s  | 0.10s | 3% |

The computational complexity of `generate_symmetric_group` only depends on the number of sublattices that are participating in symmetry operations and the computational time is independent of the number of constituents. Time and memory should not be an issue. 

The way `add_phase_symmetry_ordering_parameters` is implemented, we waste some time by recomputing essentially the same group. Since `generate_symmetric_group` operates only on the sublattices, what is inside the sublattices doesn't matter except for the set operations to determine the uniquely generated arrays. That is, constituent arrays of `(("AL",), ("NI",), ("NI",), ("NI",))`, `(("AL", "NI"), ("*",), ("*",), ("*",))`, `(("FE",), ("TI",), ("TI",), ("TI",))`, `("A", "B", "B", "B")`, `(0, 1, 1, 1)`, ... all will generate essentially the same symmetry group (for FCC: 4 different, symmetrically equivalent permutations). Theoretically, we could go through and find all the abstract unique symmetry patterns, call `generate_symmetric_group` once for each pattern, then apply that pattern to construct concrete constituent arrays for each parameter. It's unclear whether there would be a meaningful benefit 

> Another question: Option B/F apply to all of a phase's model parameters (how it's implemented here), not just the energy, correct?

Yes, I believe so. The Al-Fe database added to the tests has only the symmetrically distinct `BMAGN` parameter given for the option B phase, but has all parameters listed in the version of the 4SL bcc phase without option B. That is consistent with how partitioned parameters work.

bocklund:
With the latest changes, you can now delete a model hint and it will effectively convert the database to not use the symmetry option.

```python
dbf = Database("mpea-02b.tdb")
del dbf.phases["FCC_4SL"].model_hints["symmetry_FCC_4SL"]
dbf.to_file("mpea-02b-noF.tdb")
```

However, when testing this, I noticed that calling `remove_phase_symmetry_ordering_parameters` has a side effect of modifying the database, so `write_tdb` is impure. Calling the `write_tdb` function will remove the symmetry parameters, which is a bug. To fix, how about either:
1. `write_tdb` makes a copy of the database
2. `write_tdb` doesn't remove parameters, but does a check internally to skip writing a parameter, something like if `phase_obj.model_hints.get(symmetry_hint, False) and param.get("_generated_by_symmetry_option", False)`

I'm leaning towards (2), but not sure. Is `remove_phase_symmetry_ordering_parameters` still useful to keep around if it's not used?

richardotis:
I think (2) is a good approach. `remove_phase_symmetry_ordering_parameters` can stick around if you add a test to make sure it doesn't break over time, or you can just remove it.

pycalphad/core/utils.py
```diff
@@ -428,3 +428,102 @@ def wrap_symbol(obj):
     else:
         return Symbol(obj)
 
+
+def recursive_tuplify(x):
```
richardotis, line 432: Duplicate of https://github.com/pycalphad/pycalphad/blob/56201bfb10372fb7ef90efde4f3fa431781c36bc/pycalphad/io/database.py#L20-L22
We should pick one to keep and remove the other.

pycalphad/tests/test_utils.py
```diff
@@ -59,3 +59,127 @@ def test_instantiate_models_only_returns_desired_phases(load_database):
...
+    assert symm_groups == [("A", "A", "A", "B")]
+
+    symm_groups = generate_symmetric_group((("CR", "FE"), "VA"), None)
+    assert symm_groups == [
+        (("CR", "FE"), "VA")
+    ]
+
+
+def test_generating_symmetric_group_bcc_4sl():
+    """Binary BCC 4SL ordered symmetric configurations can can be generated"""
```
richardotis, line 90: extra "can"

richardotis (changes requested):

pycalphad/core/utils.py
bocklund: Going to keep `recursive_tuplify` as it will descend into tuples and tuplify them, for example:

`[ ([], []), ([], []) ]`:

```python
>>> _to_tuple( [ ([], []), ([], []) ] )
(([], []), ([], []))

>>> recursive_tuplify( [ ([], []), ([], []) ] ) 
(((), ()), ((), ()))
```

For the current use case of `_to_tuple` (only in `Database.__hash__` for model hints), `recursive_tuplify` should be more correct.

pycalphad/tests/test_calculate.py
```diff
@@ -240,3 +240,15 @@ def test_charged_infeasible_minimum_norm():
     # Check that the point sample didn't get 'stuck' in part of the space
     assert np.any(np.logical_and(output[:, 1] > 0.05, output[:, 1] < 0.15))
     assert np.any(np.logical_and(output[:, 1] > 0.25, output[:, 1] < 0.35))
+
+
+@select_database("Al-Fe_sundman2009.tdb")
+def test_BCC_phase_with_symmetry_option_B(load_database):
```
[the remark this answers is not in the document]
bocklund: I don't have a database like this one that has both models or easy access to another software for a test. I'm not opposed to the test.

Strictly, it shouldn't be necessary because this test shows that option B is correct and we have orthogonal tests that the symmetry groups of option F are generated correctly.

I'll note that I did an eye test on the Al-Ni phase diagram from `mpea-02b.tdb` database and it is clearly incorrect when plotted using `develop` and correct when plotted using this branch.

pycalphad/io/tdb.py
```diff
@@ -572,6 +574,43 @@ def _apply_new_symbol_names(dbf, symbol_name_map):
...
 
 
+KNOWN_SUBLATTICE_SYMMETRY_RELATIONS = {
+    # Keys should correspond to the model hints added via the `phase_options` dict
+    "symmetry_FCC_4SL": [[0, 1, 2, 3]],
+    "symmetry_BCC_4SL": [[0, 1], [2, 3]],
+}
+
+
+def add_phase_symmetry_ordering_parameters(dbf):
```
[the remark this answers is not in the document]
bocklund: Yes, `pycalphad-xml` can and should be updated with this feature. I agree that there's some need for a place to put free functions that operate on `Database` objects or I/O above the level of any particular on-disk representation

richardotis (approved):

PR #416, opened 2022-05-18
FIX: Canonicalize coordination number sorting in DAT reader

jorgepazsoldanpalma:
(This summary added from the comments in the PR)

Previously PyCalphad would alphabetically express the quadruplets and because of this if the coordination numbers of a binary quadruplet in the database were not expressed in alphabetic order in the database parameters, then the coordination numbers would be incorrectly ordered. Sometimes this would trigger a calculation of coordination number, instead of a lookup. A fix has been added to the cs_dat parser to canonically sort coordination numbers when DAT files are read.

The test makes sure that the order in which the species are entered in the DAT file does not affect the calculated energy. In this specific test the Ni is entered before the K species.


jorgepazsoldanpalma:
Yup

bocklund (approved):

jorgepazsoldanpalma:
My pleasure!

jorgepazsoldanpalma:
yup

PR #422, opened 2022-06-17
ENH: Implement writing of ChemSage DAT files

### LINKED ISSUE #413: ChemSage DAT support: DAT writing
Writing of ChemSage-style DAT files has not yet been implemented. Adding this would close the TDB/XML/DAT loop (at least for some subset of features) not just in pycalphad, but for the CALPHAD community in general.

I don't believe pycalphad is picky about the widths of values in a DAT file when reading, and Thermochimica certainly is not. However, for generality, it is probably best to output DAT files readable by FactSage (i.e. to fix the width of every written value). A DAT file validation routine would be useful for verifying correct implementation.

Also note that FactSage has updated the DAT format as of version 8.1, and databases written in the new format aren't readable by 8.0 and earlier, Thermochimica, or pycalphad (as far as I know). For now, the previous format has not been deprecated in FactSage. I suggest targeting the FactSage 8.0 and earlier format to start.

maxposchmann:
Adds support for writing of ChemSage DAT files with IDMX, QKTO, RKMP, SUBL, SUBG, SUBQ phase models, plus magnetic variants of RKMP and SUBL.

Notwithstanding open issues on DAT parsing, DAT files can be read and an equivalent DAT produced for cases I've tested. These include the databases `Kaye_NobleMetals.dat`, `FeCuCbase.dat`, and `FeTiVO.dat` available on the [Thermochimica repo](https://github.com/ORNL-CEES/thermochimica/tree/master/data).

Reasonable DAT files have been produced by converting TDBs `Al-Mg_Zhong.tdb`, `Al-Cu-Y.tdb`, and `alfe_sei.TDB` from the `examples` directory.

I've tried to be verbose in comments, but the code could probably be factored better. Input on this is desired.

This is intended to close https://github.com/pycalphad/pycalphad/issues/413, no doubt with many enhancements requiring new issues in the future.


pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+                    z_string = '      '.join([f"{z:.7f}" for z in param['coordinations']])
+                    output += f'{con_string}  {z_string}\n'
+
+        # Write magnetic excess mixing data
+        if phase_model in ('RKMPM', 'SUBLM'):
+            # Get excess magnetic parameters
+            for tc in tcs:
+                tc_value = tc['parameter']
+                tc_constituents = tc['constituent_array']
+                tcs.remove(tc)
```
richardotis, line 1636: Mutating a list while you're iterating over it can have unexpected results in Python. I think you could sort both lists of parameters by the `constituent_array` key. Something like (may need more error checking)
```python
from collections import defaultdict
magnetic_by_constituent_array = defaultdict(list)
for tc in tcs:
    magnetic_by_constituent_array[tc['constituent_array']].append(tc)
for bmagn in bmagns:
    magnetic_by_constituent_array[bmagn['constituent_array']].append(bmagn)

for tc, bmagn in magnetic_by_constituent_array.values():
    print(tc, bmagn) # etc
```

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+                                        name += f'{int(stoich)}'
+                                    else:
+                                        name += f'{stoich:.2g}'
+                        try:
+                            indices.append(1 + endmember_names.index(name))
+                        except ValueError:
+                            print(f'Endmember {name} not found in phase {phase_name}')
+                            print(f'List of known endmembers: {endmember_names}')
+                            print('Write aborted')
+                            return
```
richardotis, line 1669: This should call `raise ValueError` again with the error message so the exception is propagated correctly to the caller.

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+                                indices.append(1 + flat_constituents.index(species.name))
+                            except ValueError:
+                                print(f'Can\'t find constituent {species.name}')
+                # Now write excess magnetic terms for current constituent_set
+                output += f'{len(indices):4}\n'
+                # TODO: Get order properly if possible
+                order = 1
+                output += f'{"".join([f"{ind:4}" for ind in indices])}{order:4}\n'
+                # Check if magnetic parameters are provided as piecewise
+                if type(tc_value) == type(Piecewise([0,True])):
```
richardotis, line 1685: Is type checking what we want to do here, or should it just be value comparison?

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+                    print(f'Piecewise Curie temperature in species {name} of phase {phase_name} not permitted.')
+                    print(f'Value from first temperature interval will be used.')
+                    print(f'Check original expression to determine if this is acceptable:')
+                    print(tc_value)
+                    print()
```
richardotis, line 1690: This should warn or raise instead of printing

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+                    tc_value = float(tc_value.args[0])
+                if type(bmagn_value) == type(Piecewise([0,True])):
+                    print(f'Piecewise magnetic moment in species {name} of phase {phase_name} not permitted.')
+                    print(f'Value from first temperature interval will be used.')
+                    print(f'Check original expression to determine if this is acceptable:')
+                    print(bmagn_value)
+                    print()
```
richardotis, line 1697: This should warn or raise instead of printing

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+                                if stoich % 1 == 0:
+                                    name += f'{int(stoich)}'
+                                else:
+                                    name += f'{stoich:.2g}'
+                    try:
+                        indices.append(1 + endmember_names.index(name))
+                    except ValueError:
+                        print(f'Endmember {name} not found in phase {phase_name}')
+                        print(f'List of known endmembers: {endmember_names}')
+                        print('Write aborted')
```
richardotis, line 1738: Same comment about warning or raising

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+                        return
```
richardotis, line 1786: Same comment about warning or raising

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+            unique_constituent_sets = set([param['constituent_array'] for param in excess_params])
+            for constituent_set in unique_constituent_sets:
+                # Get indices of participating constituents in phase (order of printed endmembers)
+                indices = []
+                for sublattice in constituent_set:
+                    for species in sublattice:
+                        try:
+                            indices.append(1 + flat_constituents.index(species.name))
+                        except ValueError:
+                            print(f'Can\'t find constituent {species.name}')
```
richardotis, line 1835: We should probably re-raise here since this exception means something is wrong

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+
+    fd.write(reflow_text(output, linewidth=maxlen))
+
+def write_stoichiometry(stoichiometry):
+    # Create the stoichiometry segment of a species line
+    stoichiometry_string = ''
+    for stoich in stoichiometry:
+        if np.isclose(stoich % 0.1, 0.0) or np.isclose(stoich % 0.1, 0.1):
+            # If stoichiometry fits neatly in one decimal place, use that
+            stoichiometry_string += f'{stoich:7.1f}'
```
richardotis, line 1961: Loving the efficient use of number format strings here and elsewhere

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+            if len(splitOrder) == 2:
+                if splitOrder[0] == 'T':
+                    try:
+                        orderStrip = splitOrder[1].replace('(','').replace(')','')
+                        extra_parameters.append((coeff_string,f'{float(orderStrip):.2f}'))
+                        has_extra_parameters = True
+                        continue
+                    except ValueError:
+                        pass
+            print(f'WARNING: Skipped parameter with order {t_order} and coefficient {coeff_string}')
```
richardotis, line 2076: Should probably call `warn` explicitly instead of printing

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+
+    return coefficients, extra_parameters, has_extra_parameters
+
+def format_coefficient(coeff):
```
richardotis, line 2080: Figuring these rules out looks like it was painful

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+    while requires_substitution:
+        requires_substitution = False
+        param_symbols = [s.name for s in subs_param.free_symbols]
+        # Check if any of the symbols in symbols are still in the parameter
+        common_symbols = [s for s in symbols if s in param_symbols]
+        if common_symbols:
+            # Do another round of substitutions if so
+            subs_param = piecewise_fold(simplify(subs_param.subs(symbols)))
+            requires_substitution = True
+    return subs_param
```
richardotis, line 2143: I think you can replace this function with `Model.symbol_replace` and eliminate the sympy dependency:
```python
# Maximum number of levels deep we check for symbols that are functions of
# other symbols
_MAX_PARAM_NESTING = 32
def symbol_replace(obj, symbols):
    """
    Substitute values of symbols into 'obj'.
    Parameters
    ----------
    obj : SymEngine object
    symbols : dict mapping symengine.Symbol to SymEngine object
    Returns
    -------
    SymEngine object
    """
    try:
        # Need to do more substitutions to catch symbols that are functions
        # of other symbols
        for iteration in range(_MAX_PARAM_NESTING):
            obj = obj.xreplace(symbols)
            undefs = [x for x in obj.free_symbols if not isinstance(x, v.StateVariable)]
            if len(undefs) == 0:
                break
    except AttributeError:
        # Can't use xreplace on a float
        pass
    return obj
```

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+    # Is this monstrosity necessary? Yes, it would seem so.
+    simplified_parameter = iterative_substitution(equation,symbols)
+    gibbs_equation = expand(simplify(simplify(simplified_parameter))).args
+    if not gibbs_equation:
+        # If all ranges have 0 value, simplify returns empty set, so revert
+        simplified_parameter = equation.args
+        gibbs_equation = []
+        for i in range(int(len(simplified_parameter)/2)):
+            gibbs_equation.append((simplified_parameter[i*2],simplified_parameter[i*2+1]))
+    return gibbs_equation
```
richardotis, line 2156: There's probably another way to accomplish the goal here, but I don't fully understand the purpose of this function yet. I'd like to avoid reintroducing sympy as a dependency as much as possible

richardotis (changes requested):
Max, this is looking great! Thanks so much for working on this. Besides the specific comments, in general this PR is going to need several tests. Any DAT files you're able to contribute to the test suite would be highly appreciated. In particular I think we'll need tests for:
1. DAT read/write/read (round-trip) Database equality comparison, for a few cases
3. TDB->DAT conversion, where supported (can we achieve Database object equality on round-trip conversion?)

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+    supported_phase_types = ['IDMX','QKTO','RKMP','SUBL','RKMPM','SUBLM','SUBG','SUBQ'] # TODO: 'SUBI'
+
+    # Get numbers of solution phases (and type/species for each) and pure condensed phases
+    solution_phases = []
+    stoichiometric_phases = []
+    solution_phase_types = []
+    solution_phase_species = []
+
+    # DAT *always* includes ideal gas phase (gas_ideal) in header (in first solution phase position), even if not used
+    solution_phases.insert(0,'GAS_IDEAL')
```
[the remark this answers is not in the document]
maxposchmann: Certainly works with an ideal gas in. Frankly I don't yet understand other gas models well enough yet, I should probably read up on the options in both the TDB and DAT formats before deciding how to handle.

In my experience with other TDB -> DAT converters, they often break on pressure-dependent terms. This might be a tricky spot.

pycalphad/io/cs_dat.py
```diff
@@ -1159,6 +1162,999 @@ def parse_cs_dat(instring):
...
+
+        # Write endmember data
+        endmember_names = []
+        for endmember in endmember_params:
+            # Generate species names and stoichiometries
+            name = ''
+            stoichiometry = [0 for _ in elements_ordered]
+            # Overwrite sublattice_weights for MQM phases
+            if phase_model in ('SUBG', 'SUBQ'):
+                sublattice_weights = itertools.cycle(endmember['stoichiometry'])
```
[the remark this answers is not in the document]
maxposchmann: `cycle` is convenient, but no doubt is not the only option. For some phase models we have sublattices 1 through N each with a weight, and need to iterate through them, in order, repeatedly. For others the weights are all 1. This seemed to capture both cases neatly.

Is there any chance of the sublattice order getting shuffled? This would break if some endmembers have their constituents listed out of order.

pycalphad/io/cs_dat.py
maxposchmann: The calls to `piecewise_fold` and `simplify` here are crucial to getting the equations formatted into simple intervals that can be written to a DAT. I tried switching to `symbol_replace` as you highlighted, but it returns a nested piecewise equation that can't be used.

For more details of what needs to happen, you might look at my recent question on [StackExchange](https://stackoverflow.com/questions/72623272/simplifying-nested-piecewise-equations-in-sympy).

pycalphad/io/cs_dat.py
maxposchmann: Deliberately checking the type. Usually these are floats, but I found some come in as Piecewise instead.

The Piecewise terms in `alfe_sei.TDB` are equivalent to floats (so I assume this will happen in other cases too). Looks like a good place to use the user-selected `incompatibility` handler.

pycalphad/io/cs_dat.py
maxposchmann: This isn't a great function. Basically I ran all these lines together in a couple of places, so I just grouped them into a function. All it is attempting to do is get the Gibbs reference energy equation into the same format whether it originated from a TDB (with nested piecewise) or a DAT (with just simple coefficients).

Piecewise equations have to be simplified as far as possible, but then I needed `expand` to get coefficients of simple powers of `T`.

maxposchmann: After playing around with it, I'll add that `simplified_parameter.simplify()` doesn't yield the same results as `simplify(simplified_parameter)` (and the difference results in an error).

bocklund:
@maxposchmann sorry for my delay, this is still high on my list, but it's been a busy month or two for me. I'm hoping to get back to properly reviewing and trying to help avoid re-introducing SymPy soon.

PR #423, opened 2022-06-29
ENH: Initial volume model support

DJLicia:
This is a volume model for V0 parameters and includes a test.


richardotis:
Thanks again for your work! We're continuing this work in #466

Issue #419, opened 2022-06-10
ChemSage DAT support: Constituent parsing bug


maxposchmann:
The constituents of a multi-sublattice phase listed in a DAT can be chemically complex or contain information about the charge state of the constituent, or really anything else at all. [See here for an example](https://github.com/ORNL-CEES/thermochimica/blob/f207c6a4daeab340543f7e093692ac22dbaa6314/data/FeTiVO.dat#L157-L159):
```
 Fe[3+]                   Fe[2+]                   V2O3
 Ti[4+]                   Ti[3+]
 O
```
`Fe[3+] ` contains a number that tells you about the charge state, while `V2O3` contains numbers that tell you about the stoichiometry of the constituent.

It appears that currently, pycalphad counts on the square brackets being there around the charges. So things work fine for `Fe[3+]`, but break down if one has `Fe3+` instead, as it reads this as 3 irons. **Due to this issue, incorrect results will be computed for many DAT files.**

However, you cannot count on this convention being followed in DAT files, and in fact it rarely is. Further, one really ought not to be using the constituent names for anything. They are there for readability only, and follow no rules at all. Renaming all five constituents to `Bob` _should_ result in no changes to the calculation.

There is in fact no way at all to determine the stoichiometry of individual constituents in an MQMQA phase from a DAT file. Fortunately, these aren't necessary as the quadruplet stoichiometries can be determined from the stoichiometries of the endmembers, the coordinations, and the arrays that link the endmembers to the constituents.

bocklund:
There may be some bugs hiding, having a reproducible example would be helpful if one is found.

`Species` objects have a `name` and a `constituents` attributes. The name is supposed to be arbitrary and the constituents are what matters. As long as pycalphad gets the `constituents` correctly, everything else should _just work_. `Species` also have the feature of being capable of parsing the `constituents` if constructed from only a `name` if the constituents are not given, but I don't remember exactly how this is used in practice, except as a user-facing shortcut to make `v.Species("CU")` work as expected. 

The intent of `pycalphad.io.cs_dat.Endmember` is to construct `Species` objects using the pure element constituents according to the stoichiometry in the DAT, not to rely on the names in the models. 

IIRC, I think the parsed names of `Species` (e.g. `Species("Bob")`) do indeed get added to the `Database` object, but that should not affect anything in practice. If that is what's happening, we should probably avoid doing that.

maxposchmann:
[...]

PR #426, opened 2022-08-07
FIX: ChemSage DAT: Case sensitivity of compound names in endmembers

### LINKED ISSUE #425: Duplicate Species name error when parsing FactSage DAT into Database
When trying to parse in a DAT file which uses Cobalt 'Co' as system component and contains carbon monoxide 'CO' as gas phase, then a ValueError for using duplicate species name will be raised:

```console
db = pycalphad.Database('Bugged_DAT.dat')

  File "C:\Users\timm\anaconda3\envs\factsage-data-extraction-3-8\lib\code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "C:\Users\timm\anaconda3\envs\factsage-data-extraction-3-8\lib\site-packages\pycalphad\io\database.py", line 115, in __new__
    return cls.from_file(fname, fmt=fmt)
  File "C:\Users\timm\anaconda3\envs\factsage-data-extraction-3-8\lib\site-packages\pycalphad\io\database.py", line 220, in from_file
    format_registry[fmt.lower()].read(dbf, fd)
  File "C:\Users\timm\anaconda3\envs\factsage-data-extraction-3-8\lib\site-packages\pycalphad\io\cs_dat.py", line 1196, in read_cs_dat
    parsed_phase.insert(dbf, header.pure_elements, header.gibbs_coefficient_idxs, header.excess_coefficient_idxs)
  File "C:\Users\timm\anaconda3\envs\factsage-data-extraction-3-8\lib\site-packages\pycalphad\io\cs_dat.py", line 504, in insert
    raise ValueError(f"A Species named {sp.name} (defined for phase {self.phase_name}) already exists in the database's species ({dbf.species}), but the constituents do not match.")
ValueError: A Species named CO (defined for phase GAS_IDEAL) already exists in the database's species ({Species('CO', 'CO1.0'), Species('N', 'N1.0'), Species('H', 'H1.0'), Species('O', 'O1.0'), Species('NI', 'NI1.0'), Species('FE', 'FE1.0'), Species('CU', 'CU1.0'), Species('B', 'B1.0'), Species('SC', 'SC1.0'), Species('C', 'C1.0'), Species('MN', 'MN1.0'), Species('K', 'K1.0'), Species('CL', 'CL1.0'), Species('S', 'S1.0'), Species('P', 'P1.0'), Species('MG', 'MG1.0'), Species('CA', 'CA1.0'), Species('SI', 'SI1.0')}), but the constituents do not match.
```

This bug has been found while trying to parse exported DAT files from FactSage 6.2.
This has been tested under Python 3.9 and pycalphad 0.10.1.
I've attached a zip file containing a minimal DAT file, a script for reproducing the error and a list of packages versions, in case that's needed.

[script.zip](https://github.com/pycalphad/pycalphad/files/9276777/script.zip)

richardotis:
These are some minimal changes to allow for case sensitivity in DAT compound names (fixes #425). This doesn't address the fundamental issue that DAT and pycalphad think of constituents differently (#419) but it will hopefully enable a greater subset of DAT files to work without issue. In particular, we no longer convert the entire input file to uppercase before parsing. Instead, selective uppercasing is applied where it is "safer" to extract unambiguous names (though it is still not safe in general).

There is a test to make sure the database from the original issue parses, but this still needs a validation test to make sure the parsed database is correct.


bocklund:
I'm hesitant about this because (if I understand #419) I believe that in FactSage giving both `CO` and `Co` as uppercase `CO` in the DAT file would work correctly. 

As in #419:

> However, you cannot count on this convention being followed in DAT files, and in fact it rarely is. Further, one really ought not to be using the constituent names for anything. They are there for readability only, and follow no rules at all. Renaming all five constituents to Bob should result in no changes to the calculation.

It happens to be that the casing can disambiguate them here, but in general I don't think we can expect that to be true. It's not yet clear to me how reconcile the fact that the names are meaningless, but they are quite meaningful in pycalphad because we make an assumption that names are distinguishing in the `Model` and `variables.SiteFraction` symbolics.

richardotis:
Where is the stoichiometry information coming from for each phase at Model construction time?

bocklund:
I'm not sure I know what you're referring to about stoichiometry. The constituent species are known via the `Database().phases` `Phase` objects. And the mass of pure elements are coming from the species as well. My concern is that you'll have a `Y(PHASE,0,"CO")` (from `"CO"`) and `Y(PHASE,O,"CO")` (from `"Co"`), which mess up the free symbols in expressions where the site fractions are used (both `Model().moles` and `Model().GM`

richardotis:
I think I mean to ask about the chemistry of the phases (what chemical species are in solution), not the stoichiometry exactly. Can we construct meaningful constituent names from the cation/anion information we have for each phase? It seems like this information has to exist somewhere, or else it wouldn't be possible to construct a mass balance.

bocklund:
I think I understand your question now. According to #419, the DAT lets us get the pure elements constituents of any species. We could construct a name from that, rather than using the name/label in the DAT. I don’t know if the species by stoichiometry are guaranteed to be unique, but I don’t know why they wouldn’t be for in any phase in practice. 

For example, in this test database the numbers after `4 3` are the pure element constituents of this species. There are 18 pure elements in the database. 

```
gas_ideal
  IDMX
  CO
    4  3    0.0    0.0    0.0    0.0    0.0    0.0    0.0    0.0    0.0    0.0
     0.0    0.0    0.0    1.0    0.0    1.0    0.0    0.0
```

richardotis:
I think I'm understanding this issue better now. Can we discard the label at `Database` construction time, and construct a canonical `Species` name based on the pure element constituents (and charge information)? When we write DAT out, we can simply write the canonical name with no loss of physical information.

PR #429, opened 2022-09-03
[ENH] dot derivative function

DJLicia:
This is the dot derivative calculations function implementation


richardotis:
@DJLicia Thanks for all of your work. Over the past few months I rolled your work up into a larger branch along with some bigger changes, which can be found in #432 . I'm going to close this PR, but you can follow the progress in the other one. Thanks again!

PR #431, opened 2022-10-13
FIX: Model/calculate: Piecewise branching performance and extrapolation

richardotis:
Complex multi-sublattice phases with lots of parameters and using the magnetic ordering model can challenge pycalphad's Just-In-Time (JIT) compiler, especially for computation of second derivatives. In the worst case, the compiler will hang indefinitely and consume RAM until the process is killed.

The reason for this is the increase in algorithmic complexity that comes from having deep piecewise temperature branching in the `Model` object's representation of the Gibbs energy. However, a common case for the piecewise parameter description is that there is really only one nonzero "branch" for the entire temperature range. While TDB formally wraps every parameter in a piecewise, the `Model` object is free to discard trivial branches at build time. That is the approach used in the patch for this PR. 

This PR includes a test for such a difficult case, where the `Model` object for the corresponding phase has a Gibbs energy Hessian that cannot be built by the `develop` JIT compiler. The patch is able to reduce the number of Piecewise nodes in the Gibbs energy's abstract syntax tree by more than half. For the sake of efficiency, instead of a full correctness test we only test that the number of nodes is reduced by half.

In addition, this PR includes a change to the point sampling algorithm in `calculate`. Currently the sampler (when `fixed_grid` is `True`) tries to add additional points between all pairwise combinations of endmembers. For certain multi-component, multi-sublattice phases, there can be thousands of endmembers and, thus, millions of endmember pairs. The proposed change detects this case; when there are more than 100,000 points to be added, the algorithm only adds up to the maximum specified by the constant. All endmembers are still added, and this change does not affect the random sampling portion of the algorithm.

Finally, in addition to trivial branch elimination, the `Model` class is now updated to extrapolate the lower and upper temperature bounds of all piecewise expressions to negative and positive infinity. This brings pycalphad into line with how TC will extrapolate outside of temperature bounds for parameters. Note that limits specified by the `TEMPERATURE-LIMITS` TDB command are still not enforced and it is still possible for users to compute in non-physical regions of a database, but this was always possible; this change will allow for compatibility with a greater number of legacy databases that rely on the extrapolation behavior.

For certain pathological cases, this will reduce memory consumption by over 90% and resolve classes of memory errors for users attempting multi-component calculations with complex databases.


richardotis:
I didn't manage to get to root cause, but I did verify that you can unwrap Piecewise for every energy contribution except the ordering contribution without hitting the Mac-only failure in `test_eq_alni_high_temp` and `test_eq_alni_low_temp`. The problem is then you don't reduce enough of the Piecewise nodes to pass `test_model_deep_branching`. I switched the unwrapping to happen inside of `Model.symbol_replace`, and then I can pass the suite (including all new tests) on all platforms.

I have added a test, `test_calculation_jitter` (jitter isn't the right term, probably needs a rename), which should help explicitly detect this strange case in the future if it arises.

pycalphad/model.py
```diff
@@ -261,6 +261,21 @@ def __init__(self, dbe, comps, phase_name, parameters=None):
...
+        for atom in graph.atoms(Piecewise):
+            args = atom.args
+            # Unwrap temperature-dependent piecewise with zero-defaults
+            if len(args) == 4 and args[2] == 0 and args[3] == True and args[1].free_symbols == {v.T}:
+                arg0 = args[0]
+                try:
+                    arg0 = float(arg0)
+                except (ValueError, RuntimeError):
+                    pass
+                replace_dict[atom] = arg0
```
bocklund, line 276: What's the use case for coercing for float here? If we fail to coerce to float, the original value is still getting put in the replacement dict. Is that expected behavior?

pycalphad/model.py
```diff
@@ -280,6 +295,8 @@ def symbol_replace(obj, symbols):
             # of other symbols
             for iteration in range(_MAX_PARAM_NESTING):
                 obj = obj.xreplace(symbols)
+                if iteration < 3:
```
bocklund, line 298: Why `3` here?

bocklund (changes requested):

pycalphad/model.py
richardotis: I thought it was necessary to deal with a type promotion issue with non-float constants in SymEngine, but the test suite seems to pass without it (on both symengine.py 0.9.0 and 0.9.2). Not sure what's changed; maybe it was necessary before I moved some code around, and then I forgot to check. In any case, I will remove it now.

pycalphad/model.py
richardotis: This was some code that was originally operating on the entire Model AST, and then moved to be inside `symbol_replace` (see https://github.com/pycalphad/pycalphad/pull/431#issuecomment-1279630113 ). The iteration count limit was an attempt to work around an apparent bug, and also a (premature) performance optimization. It's no longer relevant in the current code and has been removed, so unwrapping will occur until the loop terminates. (In practice, there are rarely more than 3 levels deep of symbol nesting anyway.)

richardotis:
@bocklund ping

bocklund:
One hesitation I have for this PR is that the mixed behavior between expressions that do extrapolate outside of temperature limits automatically and expressions that don't will make it more difficult to debug cases where users are doing calculations outside of temperature limits. 

Is it too big of an effort or scope change to do that here? Does it need a separate issue?

richardotis:
I'd argue it's undefined behavior in pycalphad to compute outside of the temperature limits, and it'd still be undefined behavior after this PR was merged. I agree it's a footgun and the user has no way of knowing without examining the database, though.

For adjusting extrapolation behavior, there are two approaches:
- Database: modify the limits at Database construction time; pros: easy to do; cons: breaks roundtrip consistency
- Model: modify the limits at Model construction time; pros: protects Database integrity; cons: more complex algorithm, have to walk the expression tree, unclear whether we can detect all the cases, potentially a performance challenge

richardotis:
@bocklund This PR now extrapolates all temperature bounds, including multi-branch piecewise expressions. This uses the Model-based extrapolation approach referenced above. It ended up being a smaller delta than I expected, and the performance seems to be fine (though feel free to test on complex databases). The small performance impact makes sense to me in retrospect, as we expect fewer than 100 `Piecewise` atoms per model, and many of those will be trivial non-zero branches that hit the fast path in `Model.unwrap_piecewise`.

bocklund (approved):
What's here seems reasonable to me. Although I agree that it's technically correct, I haven't seen a use case in the wild that used or relied on temperature limits in a useful way. I'm not too worried about going out of our way to support fully correct non-extrapolation for now (YAGNI).

PR #432, opened 2022-10-23
ENH: Workspace API

richardotis:
- Adds a new `Workspace` object and associated imperative API. This object now underlies the `equilibrium` function, which remains for backwards compatibility. A `Workspace` object can be mutated after creation. The `Workspace.get()` function accepts `ComputableProperty` objects as input and returns arrays. Fixes #154 
- Adds a new `PhaseRecordFactory` object and associated connections to `PhaseRecord`, which enables deferred compilation of symbolically-defined model properties until they are called in the `PhaseRecord` API. The `PhaseRecord` object also gains `prop` and `prop_grad` functions for computing properties and property gradients of arbitrary symbolic attributes of `Model` instances.
- Creates a "Computable Property Framework" (CPF) for calculating properties of the system, individual phases, or some combination of these. It also provides a framework for defining custom properties.
- Adds support for Jansson derivatives (constrained total derivatives aka "dot" derivatives) of thermodynamic properties (fixes #261 )
- Adds support for T0 (tzero) calculation
- Adds support for driving force calculation, including nucleation driving forces, via the `IsolatedPhase` and `DormantPhase` meta-properties. Meta-properties are objects that know how to create new properties (as defined by the CPF). In this case, these meta-properties know how to start sub-calculations . The current way of finding starting points for these calculations is not ideal and prone to failure (basically reuses `CompositionSet` objects from the workspace equilibrium calculation), but it works well in the typical case.
- Adds physical units support via `pint`. All `ComputableProperty` objects have "implementation units" and "display units." Implementation units are what are assumed by the underlying numerical computation (which is done without units, for performance). The display units are what the user gets when `Workspace.get` is called. `ComputableProperty.__getitem__` allows the display units to be changed for individual calls to `Workspace.get`, or when setting `Workspace.conditions`.
- Also regarding physical units, a special unit conversion context is defined to allow conversions between per-mole properties and per-mass properties. There's room to add per-volume here as well, but that may not make it into this PR.
- Adds support for specifying linear combinations and ratios of mole fractions as a condition.
- Adds support for specifying weight fractions as a condition.
- Closes https://github.com/pycalphad/pycalphad/issues/133 through the multi-valued output from `Workspace.get` 
- Closes https://github.com/pycalphad/pycalphad/issues/163 because site fraction conditions can be given now
- Associated tests for these new features.
- `build_phase_records` is removed. Use the `PhaseRecord.prop_*` API in downstream functions instead.

Documentation
[...]

PR #466, opened 2023-04-07
ENH: Molar volume

mfrichtl:
This PR is a partial implementation of the molar volume model from [Lu, Selleby, and Sundman (2005)](https://doi.org/10.1016/j.calphad.2005.04.001). It includes the nonmagnetic thermal expansion, magnetic contribution to molar volume, and the contribution to the Gibbs free energy in cases where the molar volume is independent of pressure. Tests for these terms are also included. There are no required changes to the dependencies of pycalphad. This PR is inclusive of PR #423.

The pressure dependence of the molar volume can be implemented in the future via VC and VK parameters and Equations 9, 10, and 12 of the reference paper. It will require the exponential integral to be implemented into symengine.

### DIFF
--- a/pycalphad/model.py
+++ b/pycalphad/model.py
@@ -1419,17 +1419,17 @@ class Model(object):
         V_mag = G_mag.diff(v.P)
 
         self.MV = self.molar_volume = V_p0 + V_mag
-        self.volume_energy = S.Zero
+        volume_energy = S.Zero
 
         if VK == 0:
-            self.volume_energy = V_p0*(v.P-101325)
+            volume_energy = V_p0*(v.P-101325)
         else:
             warnings.warn(
                     f"The database for \"{self.phase_name}\" contains a term for the isothermal compressibility"
                     f"however the pressure dependence has not been fully incorporated into the molar volume or"
                     f"Gibbs free energy models. THE GIBBS ENERGY AND MOLAR VOLUME CALCULATIONS MAY BE INCORRECT.")
 
-        return self.volume_energy
+        return volume_energy
 
 
 class TestModel(Model):
--- a/pycalphad/model.py
+++ b/pycalphad/model.py
@@ -1416,8 +1416,7 @@ class Model(object):
 
         # magnetic contribution to volume
         G_mag = self.models.get('mag')
-        tau_curie = v.T/self.TC
-        V_mag = G_mag.diff(tau_curie)*tau_curie.diff(v.P)
+        V_mag = G_mag.diff(v.P)
 
         self.MV = self.molar_volume = V_p0 + V_mag
         self.volume_energy = S.Zero
--- a/pycalphad/model.py
+++ b/pycalphad/model.py
@@ -1426,7 +1426,7 @@ class Model(object):
         else:
             warnings.warn(
                     f"The database for \"{self.phase_name}\" contains a term for the isothermal compressibility"
-                    f"however the pressure dependence has not been incorporated into the molar volume or"
+                    f"however the pressure dependence has not been fully incorporated into the molar volume or"
                     f"Gibbs free energy models. THE GIBBS ENERGY AND MOLAR VOLUME CALCULATIONS MAY BE INCORRECT.")
 
         return self.volume_energy
--- a/pycalphad/model.py
+++ b/pycalphad/model.py
@@ -152,7 +152,7 @@ class Model(object):
     contributions = [('ref', 'reference_energy'), ('idmix', 'ideal_mixing_energy'),
                      ('xsmix', 'excess_mixing_energy'), ('mag', 'magnetic_energy'),
                      ('2st', 'twostate_energy'), ('ein', 'einstein_energy'),
-                     ('ord', 'atomic_ordering_energy')]
+                     ('vol', 'volume_energy'), ('ord', 'atomic_ordering_energy')]
 
     def __new__(cls, *args, **kwargs):
         target_cls = cls._dispatch_on(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


richardotis:
Thanks for your work. Where do you need the exponential integral function to complete the implementation? We may be able to rewrite or approximate the expression in terms of supported functions. Is it just the Debye function?

I'll start taking a look at the code now. When combined with the new dot derivative support in https://github.com/pycalphad/pycalphad/pull/432, we'll be able to compute coefficients of thermal expansion (CTEs) in the vicinity of the magnetic transition.

pycalphad/model.py
```diff
@@ -1358,6 +1358,78 @@ def _pure_element_test(constituent_array):
...
+
+        # nonmagnetic contribution to volume
+        V_p0 = V0*exp(VA)
+
+        # magnetic contribution to volume
+        G_mag = self.models.get('mag')
+        V_mag = G_mag.diff(v.P)
+
+        self.MV = self.molar_volume = V_p0 + V_mag
+        self.volume_energy = S.Zero
```
richardotis, line 1422: Change `self.volume_energy` to just be a function-local variable. As written, this overwrites the volume energy function itself, which is not intended behavior. At runtime users can directly access the volume energy contribution via `Model(...).models['vol']`.

richardotis:
If it's just the Debye function, I'll note some recent work investigating integral-free forms of the Debye function: https://www.mdpi.com/2227-7390/10/10/1745

In particular, Equation 71 in that paper shows how we could write the Debye function in terms of several polylogarithms, which isn't itself novel; the problem has always been that polylogarithms are difficult to implement numerically over the entire domain of $\Theta/T$ we care about. However, this new work shows a form where we would only need to call polylogarithms in the form $\mathrm{Li}_{s}(e^{-\Theta/T})$, which would restrict the domain to $0 < z < 1$. That would permit the use of relatively well-behaved (and simple) power series approximations of the polylogarithms.

mfrichtl:
@richardotis, the model uses the exponential integral function, actually the paper appears to use the E1. And its inverse. There may be some what to transform it into a Debye function, but there are numerical implementations of the exponential integral as well, such as this one: https://link.springer.com/article/10.1007/s11075-017-0331-z

I've played around a bit with the full implementation in Lu's paper and haven't yet found the magic combinations of the VC and VK parameters that match the experimental results he presents for iron using Scipy's exponential integral functions. It's entirely possible I'm missing something obvious, but my results thus far are incorrect. They're in the right ballpark, but I'm just not able to fit the experimental data with the parameters provided in the paper. I thought at least having the atmospheric molar volume model implemented into pycalphad was a good first step, with a note for future implementation of the pressure dependence.

richardotis:
We can proceed without adding more features to this PR. I just wanted the problem documented for future reference. Where in the paper do you see the E1 function? I keep missing it.

mfrichtl:
Equation 6 in the paper defines Ei(z), but the equation itself matches the definition of E1(z). They're pretty similar though, with Ei(-z) = -E1(z). https://dlmf.nist.gov/6.2#E1 has more. I'm not terribly familiar with this function, so I could be misinterpreting.

pycalphad/model.py
mfrichtl: @richardotis, I believe my last commit addressed this issue.

richardotis:
We're talking about this paper, right? https://dx.doi.org/10.1016/j.actamat.2005.01.049
Equation 6 in that paper looks to me like this:
image
I can't find the exponential integral function used anywhere in that paper.

richardotis:
Oh I see, you probably mean https://doi.org/10.1016/j.calphad.2005.04.001 - they were very productive that year, good reminder on the importance of DOIs!

mfrichtl:
Ahh I see the confusion. Lu was lead author on 3 different papers on this topic in 2005. I'm actually referring to this paper: https://www.sciencedirect.com/science/article/abs/pii/S0364591605000349. 
Equation 6 is:
image

which I think is actually the definition of the E1.
And the full molar volume model is:

image

pycalphad/tests/test_calculate.py
```diff
@@ -252,3 +252,69 @@ def test_BCC_phase_with_symmetry_option_B(load_database):
...
+    CONSTITUENT BCC_A2 :FE: !
+    PARAMETER G(BCC_A2,FE;0) 0.01 T;  6000.00 N !
+
+    PARAMETER V0(BCC_A2,FE;0) 0.01 1E-6; 6000 N !
+    PARAMETER VA(BCC_A2,FE;0) 0.01 1E-6*T; 6000 N !
+    """
+
+    db = Database(TDB)
+    energy = 1e-6*np.exp(1e-6*300)*(10e9-101325)
+    result = calculate(db, ['FE'], 'BCC_A2', T=300, P=10E9, N=1, output='volume_energy')
```
richardotis, line 289: To fix this test, you could define another custom property like `volume_energy_contribution`. It's just that you can't have a property with the same name as the function.

pycalphad/model.py
richardotis: I added a comment by the test with how to fix it. You just need to change the name of `self.volume_energy` to `self.something_else` so that it doesn't conflict with the name of the function.

pycalphad/tests/test_calculate.py
mfrichtl: Yeah that was  a silly oversight on my part. I just fixed the test by backing out the volume energy contribution to GM. This way the model is consistent with the way you've set up other contributions in that I didn't define a custom property just for the volumetric contribution. LMK if you would like to see any other changes.

richardotis (approved):

richardotis:
Thank you!

Issue #468, opened 2023-04-13
Error in equilibrium with pycalphad 0.10.2


tobiasspt:
Hi, 

I do get following error message, when I perform equilibrium calculations at certain conditions with pycalphad 0.10.2. (I have attached the TDB and the example code). 
The error does not occur when I remove GRAPHITE (which contains only 'C' as the component) as a phase. 
The error does occur for similar temperatures and concentrations. 
The error does not occur with pycalphad 0.10.1. 
pycalphad 0.10.1 and ThermoCalc predict, for the same conditions as in the example, a mixture of BCC_A2 and GRAPHITE. 
(['BCC_A2', 'GRAPHITE', '']   [9.99647257e-01, 3.52743135e-04,            nan])

``` 
Traceback (most recent call last):   
    File "C:\Users\spitaler.t\Desktop\example\example.py", line 25, in <module>    
       eq = equilibrium(dbf, ['FE','C','VA'], phases, {v.N:1, v.P:1e5, v.T:T, v.X('C'):0.0053})    
   File "C:\Users\spitaler.t\Miniconda3\envs\ptest2\lib\site-packages\pycalphad\core\equilibrium.py", line 281, in equilibrium   
       properties = _solve_eq_at_conditions(properties, phase_records, grid,            
   File "pycalphad/core/eqsolver.pyx", line 260, in pycalphad.core.eqsolver._solve_eq_at_conditions       
ValueError: could not broadcast input array from shape (4,) into shape (3,)         
```                
[example.zip](https://github.com/pycalphad/pycalphad/files/11219840/example.zip)

How can I avoid this error?

Cheers, 
Tobias

richardotis:
I'm able to reproduce this error, which is a bug. What's happening is that the minimizer is returning a non-physical solution of four stable phases, when only two are allowed by the Gibbs phase rule for the given conditions. (pycalphad will tolerate one extra stable phase (which is where the `3` comes from), to catch invariant reactions.)

The change in behavior is (probably) because of the bugfix in https://github.com/pycalphad/pycalphad/pull/428 which is causing the minimizer to return a different starting point than in 0.10.1. That fix has apparently uncovered another bug here. It should be impossible for the minimizer to return a solution that violates the Gibbs phase rule. I'm investigating it now.

richardotis:
As a workaround, you can try altering the sampling point density:
[...]

PR #470, opened 2023-04-22
FIX: Gibbs phase rule compliance

richardotis:
Fixes gh-468.

In some cases the outer loop of the global minimizer (`eqsolver.pyx`) will add phases that exceed the number allowed by the Gibbs phase rule. Infeasible solutions passed to the minimizer aren't, themselves, a problem, and can ultimately help us find minimum-energy solutions. However, the minimizer must ensure that the final solution returned is compliant.

Prior to this change, the minimizer could become 'stuck' with a solution that violates the Gibbs phase rule and is also metastable with respect to the sampled grid of points. The minimizer would nonetheless (falsely) converge and return, e.g., a four-phase stable equilibrium in an isothermal-isobaric binary system. It would then be unable to recover, either returning the non-physical solution or raising a difficult-to-interpret error.

The fix is two-fold: First, `change_phases` is usually only called when the convergence criteria are otherwise satisfied so, in this change, when a Gibbs phase rule violation of the candidate solution is detected, the minimum tolerated phase amount is increased by several orders of magnitude (currently from `1e-9` to `1e-4`). This will force smaller-amount composition sets to be automatically removed, resetting the convergence, and continuing the optimization. For the case reported in gh-468, this fixes the issue, and should generalize to similar scenarios.

Second, as a failsafe, Gibbs phase rule compliance is explicitly checked at the end of `run_loop` to prevent the minimizer from reporting `converged=True` for any violating (non-physical) case we've failed to catch with the revised logic.

Thanks @tobiasspt for reporting.


pycalphad/core/minimizer.pyx
```diff
@@ -377,6 +377,9 @@ cdef class SystemSpecification:
             state.increment_phase_metastability_counters()
             if not phases_changed:
                 advance_state(self, state, eq_soln, step_size)
+        if state.free_stable_compset_indices.shape[0] > self.max_num_free_stable_phases:
+            # Gibbs phase rule violation in solution
+            converged = False
```
bocklund, line 382: The idea is that we would hit this if the other change to MIN_PHASE_AMOUNT was not sufficient to remove enough phases violating Gibbs phase rule? So the other change justifies not coverage?

bocklund:
@zhyrek this may fix some issues you've been running in to

pycalphad/core/minimizer.pyx
richardotis: Correct. The challenge with writing a test to cover that branch is that, ideally, that branch should be unreachable. I don't know how to write a test case that reaches that branch; if I did, that would itself be a bug. The new MIN_PHASE_AMOUNT logic has addressed the only known case, and this branch is defensive programming. Unfortunately we don't currently have a way to bubble specific error messages up with the convergence state.

PR #472, opened 2023-04-25
BLD: Support PPC wheels

bocklund:


bocklund:
We used this to try to generate some PPC wheels. They didn't work and we ended up working around the issue, closing this for now.

PR #484, opened 2023-07-30
BLD/FIX: coverage: ignore errors to workaround Cython>=3 issue

richardotis:
Workaround for cython/cython#5581

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,4 +38,8 @@ omit = [
     "stringsource",
 ]
 
+[tool.coverage.report]
+ignore_errors = true
+# workaround for https://github.com/cython/cython/issues/5581
+
 [tool.setuptools_scm]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,6 @@ omit = [
 ]
 
 [tool.coverage.report]
-ignore_errors = true
-# workaround for https://github.com/cython/cython/issues/5581
+ignore_errors = true # workaround for https://github.com/cython/cython/issues/5581
 
 [tool.setuptools_scm]


richardotis:
Okay, so the test failures are specific to running Cython 3.0. If we force a downgrade we get normal behavior. That's consistent with the fact that these test failures started around the time Cython 3 went live.

richardotis:
Something strange is happening here. With `Cython<3` added to `pyproject.toml`, CI goes green. However, `pip list` and the debug logs inside of `coverage xml` are showing Cython 3.0.0. In fact, for the 'fail' and 'pass' cases, there is no difference in the contents of `pip list`. It looks like it's only the `python -m build wheel` step that is respecting the build dependency version limit, and the Cython the rest of the build environment is seeing is coming from elsewhere.

bocklund:
> Checklist
> 
>     * [ ]  If any dependencies have changed, the changes are reflected in the
>       
>       * [ ]  `setup.py` (runtime requirements)
>       * [ ]  `pyproject.toml` (build requirements)
>       * [ ]  `requirements-dev.txt` (build and development requirements)

Maybe try changing in `setup.py` and `requirements-dev.txt` too? I forget the details, but what might be happening is pyproject is controlling build time dependencies, which happen in an isolated environment, and `setup.py` is affecting run time dependencies. `requirements-dev.txt` is in the developer build instructions (for non-isolated build environments) and used in docs builds.

richardotis:
Thinking to go with a different workaround. Coverage.py allows us to turn this error into a non-fatal warning, by adding `ignore_errors = true` to our `pyproject.toml`. This seems preferable to pinning Cython.

bocklund (approved):

PR #496, opened 2023-11-02
FIX: MQMQA: Handle species that have mole numbers that are higher than 1

jorgepazsoldanpalma:
In the following commit, changes have been made so that the MQMQA model can handle species that have mole numbers that are higher than 1 (ie. dimers,trimers). In order to handle this correctly the _n_i function, ideal mixing function and some energy tests had to be adjusted. Additionally, two new tests were introduced which make sure the species with multiple moles and vacancies in the MQMQA model are correctly described.


bocklund:
@jorgepazsoldanpalma is the change from `Fe2+` -> `Fe+2`, `Fe3+` -> `Fe+3`, `Sb3+` -> `Sb+3` in the `Shishin_Fe-Sb-O-S_slag.dat` database significant? 

I tried to change just `Sb3+` back to `Sb+3` and the tests broke. If I'm understanding correctly, the changes in this pull request change the semantics about what is a valid name for a species (or the semantics of how a species string name is turned into a `Species` object). My understanding is that current practice in the `develop` branch is that we treat `Fe2+` in the DAT as a string name of the species and we get the charge state from the following lines.

The key question I have is: do FactSage and Thermochemica work with both versions (both `Sb3+` and `Sb+3`)? I.e., does this PR introduce a regression in the databases we can handle?

jorgepazsoldanpalma:
Hi @bocklund ! The changes I made to the species were due to the fact that when the .dat file reads the species as Fe2+ it believes that the species should have a constituent of 2 rather than 1.  That was the main reason I changed it. As you correctly pointed out, the lines below the specie labeling do in fact give information regarding the respective charge of the species.

bocklund:
> Hi @bocklund ! The changes I made to the species were due to the fact that when the .dat file reads the species as Fe2+ it believes that the species should have a constituent of 2 rather than 1.  That was the main reason I changed it. As you correctly pointed out, the lines below the specie labeling do in fact give information regarding the respective charge of the species. 

Is `Fe2+` valid in Thermochimica and FactSage?

jorgepazsoldanpalma:
@bocklund Yes, it is a valid expression in both softwares.

jorgepazsoldanpalma:
@bocklund The originaly reason I changed the lines, if I remember correctly, was because the energies were not matching due to the fact that the parser was interpreting the Fe2+ as being a species with charge 2 and a constituent of 2 rather than 1. Perhaps this bug in the parser has been addressed?

pycalphad/tests/databases/Shishin_Fe-Sb-O-S_slag.dat
```diff
@@ -67,7 +67,7 @@
   2.00000      3.00000     0.000000     0.000000     0.000000
   1.65300
    3   2
- Fe2+                     Fe3+                     Sb3+
+ Fe+2                     Fe+3                     Sb+3
  O                        S
   2.00000      3.00000      3.00000
```
bocklund, line 72: @jorgepazsoldanpalma You added some new tests for databases containing `Be2` or `Al2`, which is good. The fact that you needed to change this database to be `Fe+2` when Thermochimica/FactSage can use the `Fe2+` version means that something about our implementation is wrong. 

Both the new `Be2`/`Al2` tests that were added should pass _and_ the existing tests that use the `Fe2+`-style should work (and should agree with the output of using `Fe2+`-style database in Thermochimica/FactSage)

jorgepazsoldanpalma:
@bocklund, just to make sure I am understanding your comments. You are saying that when you change the species from Fe+2 to Fe2+ PyCalphad just raises and exception but still correctly calculates the energies. If that is what you are saying then that is not the case at all for me. I obtain an error in which the energies are not the same with the values from Thermochimica. Now this is my understanding the of the error so far on my end and in the branch I am proposing. 
when I made the first implementation to in the MQMQA model so that it can handle species whose mass was not equal to 1, I noticed that I kept failing with the Shinshin tests. Taking a closer look the error was consistently a mismatch between the energies that were in Thermochimica and PyCalphad.  Below is an example of how the the species were expressed originally

@select_database("Shishin_Fe-Sb-O-S_slag.dat")
def test_MQMQA_SUBQ_Q_mixing_Sb_O_S_1000K(load_database):
    dbf = load_database()

    FE2 = v.Species("FE2++2.0", constituents={"FE": 2.0}, charge=2)
    FE3 = v.Species("FE3++3.0", constituents={"FE": 3.0}, charge=3)
    SB3 = v.Species("SB3++3.0", constituents={"SB": 3.0}, charge=3)
    O = v.Species("O-2.0", constituents={"O": 1.0}, charge=-2)
    S = v.Species("S-2.0", constituents={"S": 1.0}, charge=-2)
    mod = ModelMQMQA(dbf, ["SB", "S", "O"], "SLAG-LIQ")

This is wrong since what this is implying is the Fe2 is in fact a dimer (2 atoms/moles of Fe) of charge 2 rather than 1 mole of Fe that has a charge of 2. Now the reason this test was not raising issues in the past was in fact the error addressed in this branch. The code assummed all species had a constituent of 1 no matter was assigned to the species itself. This is why in the LiF-BeF2 system is was incorrectly normalizing the Gibbs energy of the quadruplets with Be dimers in it. In this current implemetation I make the code read the species constituent so that it can correctly differentiate dimers and monomers, and hence the test from Shishin fails now. The reason why I changed the configuration of the species is because I know the code interprets a number next to a species to be the constituent number and instead of fixing the fundamental problem (which seems to be how the .dat file  address how mass is attributed to species ) I was lazy and quickly changed the expressions  of the species in the database. I hope that now my actions are more clear.

bocklund:
> The reason why I changed the configuration of the species is because I know the code interprets a number next to a species to be the constituent number and instead of fixing the fundamental problem (which seems to be how the .dat file address how mass is attributed to species ) I was lazy and quickly changed the expressions of the species in the database. I hope that now my actions are more clear.

Yeah, I think this is what I'm getting at. 

To merge this pull request we need to ensure that we:
1) Do not break existing correct behavior (i.e. having Shishin species of ` Fe2+`, `Fe3+`, and `Sb3+` needs to work and be in agreement with Thermochemica). 
2) Add support for the dimer species like `Be2`

Right now we are violating (1) at the expense of (2), but I we should merge the implementation that satisfies both (1) and (2). Let me know if I'm understanding correctly and whether that makes sense.

If it's helpful at all, I did the work to revert the Shishin database and tests and pushed the changes up to your branch here. We can successfully merge this pull request when the implementation makes all the tests (new and old) pass without modifying the databases.

jorgepazsoldanpalma:
Roger. I'll get to that implementation asap! Thanks!

pycalphad/io/cs_dat.py
```diff
@@ -543,6 +543,12 @@ def rename_element_charge(element, charge):
...
     elif charge > 0:
+        unwanted_characters=['(',')','[',']','{','}','+','-','.',',']  
+        element=list(element)
+        element=[i for i in element if i not in unwanted_characters]
+        element=[i for i in element if i.isdigit()==False ]
+
+        element=''.join(element)
         return f'{element}+{charge}'
     else:
         return f'{element}-{abs(charge)}'
```
bocklund, line 553:
```python
def rename_element_charge(element, charge):
    element = "".join(filter(str.isalpha, element))
    if charge == 0:
        return f'{element}'
    elif charge > 0:
        return f'{element}+{charge}'
    else:
        return f'{element}-{abs(charge)}'
```

Is this representative of what you were doing here? Filtering for only alphabetical characters?

I assume we should just blanket apply that to elements, is that a fair assumption? You see a lot more DAT files than me.

pycalphad/io/cs_dat.py
```diff
@@ -642,11 +648,29 @@ def insert(self, dbf: Database, phase_name: str, As: List[str], Xs: List[str], e
...
+        spec=[i for count,i in enumerate(spec) if count<2 and i not in unwanted_characters]
+        spec=''.join(spec)
+    else:
+        pass
+    if spec==element:
+        bool_= True
+    else:
+        bool_=False
+        
+    return bool_
```
bocklund: Same thing for here on filtering for alphabetical characters

pycalphad/io/cs_dat.py
jorgepazsoldanpalma: Yes, that is fair in my experience.

pycalphad/tests/test_equilibrium.py
```diff
@@ -911,6 +911,49 @@ def test_MQMQA_ternary_equilibrium_ideal(load_database):
...
+    """Ternary ideal"""
+    dbf = load_database()
+    comps = ['F', 'LI', 'BE', 'VA']
+    eq = equilibrium(dbf, comps, ['MSFL'], {v.N: 1, v.P: 101325, v.T: 1200, v.X('LI'):0.2, v.X('BE'):0.2}, calc_opts={'pdens': 50000})
+    print('GM', eq.GM.values.squeeze())
+    print('Y', eq.Y.values.squeeze())
+    print('Phase', eq.Phase.values.squeeze())
+    assert np.isclose(eq.GM.values.squeeze(), -3.75743E+05)  # Thermochimica result
+    assert np.all(eq.Phase.squeeze() == ['MSFL', '', '', ''])
+    assert np.allclose(eq.Y.values.squeeze()[0, :],[5.4835E-05, 6.8482E-03, 2.6583E-02, 0.21381, 0.65434, 9.8365E-02], atol=1e-5)  # Thermochimica result
```
bocklund: This one is interesting. The test does indeed fail (it's marked xfail), but only the site fractions don't match. We get the right energy and phase. Could you double check the Thermochimica values?

bocklund: I also removed the pdens override and get virtually the same answer within PyCalphad down to <5 µJ. 

If you can confirm that the Thermochimica site fractions are correct, can you check what the energy is if you plug the site fractions into PyCalphad directly (like the `test_energy` tests) and confirm that the energy of those site fractions is actually lower than the PyCalphad one (i.e. confirm that the Thermochimica result isn't the metastable one)?

pycalphad/models/model_mqmqa.py
```diff
@@ -276,6 +281,18 @@ def _X_i(self, dbe, species: v.Species):
...
+        Return the site fraction of species on it's sublattice. Poschmann Eq. 9 and 10.
+        """
+        if species in self.cations:
+            cons_spec=[cons for cons in species.constituents.values()][0]
+            return (self._n_i(dbe, species)/cons_spec) / sum((self._n_i(dbe, a)/cons) for a in self.cations for cons in a.constituents.values())
+        else:
+            assert species in self.anions
+            cons_spec=[cons for cons in species.constituents.values()][0]
+            return (self._n_i(dbe, species)/cons_spec) / sum((self._n_i(dbe, x)/cons) for x in self.anions for cons in x.constituents.values())
+
```
bocklund: Does your implementation of `_X_i_S` here simply replace `_X_i` (i.e. current `_X_i` is wrong)? There's now no place that `_X_i` is used in the code.

If so, we should just fix `_X_i` with your change instead of making a new method.

jorgepazsoldanpalma: Hi Brandon. Yes they are meant to be the same function. I made a duplicate to make as I was developing the _X_i_S function to make sure it worked and then forgot to update the original function

bocklund: Okay. I made the change.

pycalphad/io/cs_dat.py
bocklund: I made the change here

pycalphad/io/cs_dat.py
bocklund: So is it right that `element_check` basically boils down to this?

```python
def element_check(spec,element):
    spec = "".join(filter(str.isalpha, spec))
    return spec == element
```

bocklund:
Okay, so it all looks great so far. I made a bunch of mostly superficial changes and pushed them up to your branch and those are currently reflected here.

1. There are the two remaining open comments about element_check and double checking the site fractions against Thermochimica for the FLiBe test. 
2. I need to do a proper thorough read of the DAT parser. Anything you can do in there to further clean and simplify would be helpful. 
3. If you can check the [coverage](https://app.codecov.io/gh/pycalphad/pycalphad/pull/496?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=pycalphad) there a few chunks added to `cs_dat.py` that aren't covered, mainly `common_cation_check` and `multi_valence_spec_bool`. Is there any databases you have that can exercise the lines of code that you can contribute to the test suite?

I won’t get to those tonight, but these are the last things I need to see before it’s good to merge.

bocklund:
@jorgepazsoldanpalma any updates on this?

jorgepazsoldanpalma:
Hi @bocklund 
Sorry! I let this slip. I am still trying to "optimize" the code and make it neater. I will get back to you next week about it because I will be traveling this week.

bocklund:
No worries, just trying to keep stuff from getting stale

jorgepazsoldanpalma:
Aside from the single conflict with the test_energy.py file, there have been some alterations to the code that have definitely made it neater. I still have a section to go but it is moving!

PR #498, opened 2023-11-24
MAINT/FIX: Relax version pin for symengine

### LINKED ISSUE #497: Unexpected output for calculations on Python 3.11
When doing equilibrium calculations, there's some printed output that shows up only on Python 3.11 on multiple platforms (confirmed on Ubuntu and macOS in our CI runs).

```
Attribute 'uwtable' should have an Argument
ptr @0
```

Some searching seems to indicate that this is from LLVM, so probably related to symengine or our use of it. It doesn't appear to affect the correctness of any calculations, but is an annoyance if reading through printed output.

richardotis:
Relaxes the symengine version pin so that future symengine releases also satisfy the dependency requirement. SymEngine release cadence has increased and we're also seeing fewer regressions, so I think it's safe and advisable for pycalphad to relax the hard pin. Also, this will fix #497.

Checklist
  * [x] `pyproject.toml` (build requirements)
  * [x] `requirements-dev.txt` (build and development requirements)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -68,7 +68,7 @@ setup(
         'pytest-cov',
         'scipy',
         'setuptools_scm[toml]>=6.0',
-        'symengine==0.9.2',  # python-symengine on conda-forge
+        'symengine>=0.11.0',  # python-symengine on conda-forge
         'tinydb>=3.8',
         'xarray>=0.11.2',
     ],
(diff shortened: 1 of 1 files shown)


bocklund (approved):
Is unpinning the best way to ensure that the next SymEngine release is breaking? 😄 

LGTM

PR #501, opened 2023-12-11
FIX: Molar volume model fixes

bocklund:
- Fix normalization of molar volume energy and properties to be per mole of atoms
- Set molar volume properties (`V0`, `VA`, `VK`, `VC`) as attributes of `Model`. Not sure about `VK` and `VC` parameter normalization, but for now I'm assuming they are the same normalization `VA` (parameters stored per mole of atoms, rather than the usual per mole of formula)
- Set the property name of molar volume to be `VM`, rather than `MV`, which is more consistent with commercial software and handling molar quantities
- Replace molar volume property expressions with symbols in the database. If we don't do this, any `FUNCTION` TDB keywords (which become symbols) get set to zero when computing in `calculate` or `equilibrium`

### DIFF
--- a/pycalphad/model.py
+++ b/pycalphad/model.py
@@ -1411,7 +1411,7 @@ class Model(object):
             reference_contrib = Add(*terms)
             referenced_value = getattr(self, out) - reference_contrib
             setattr(self, fmt_str.format(out), referenced_value)
-    
+
     def volume_energy(self, dbe):
         """
         Return the volumetric contribution in symbolic form. Follows the approach by Lu, Selleby, and Sundman [1].
@@ -1420,7 +1420,7 @@ class Model(object):
         ----------
         dbe : Database
             Database containing the relevant parameters.
-        
+
         Notes
         -----
         The high-pressure portion of the model is not currently implemented.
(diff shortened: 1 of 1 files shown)


bocklund:
cc @mfrichtl as this might be relevant to you

mfrichtl:
@bocklund Glad to see improvements here since I'm definitely familiar enough with the nuts and bolts of pycalphad to get it right on my first try. Maybe I was just getting lucky with my testing that the site ratio normalization didn't impact me. Unfortunately I don't know more about how the other terms are defined, but I see now in one of my references that at least `V0` is defined as m^3/(mole formula unit).

bocklund:
> @bocklund Glad to see improvements here since I'm definitely familiar enough with the nuts and bolts of pycalphad to get it right on my first try. Maybe I was just getting lucky with my testing that the site ratio normalization didn't impact me. Unfortunately I don't know more about how the other terms are defined, but I see now in one of my references that at least `V0` is defined as m^3/(mole formula unit).

If you were looking at simple bcc/fcc/hcp phases, it's not too surprising to miss the normalization. I was able to at least confirm that V0 is per mole of formula (consistent with some refs I've seen with sigma phase molar volume modeling), and I don't have any databases for that have VA parameters for phases with more than 1 mole of formula units that I've found published, but I verified that Thermo-Calc treats VA parameters are per mole of atoms (which was surprising). My suspicion is VK/VC are the same, but I'm not sure

mfrichtl:
@bocklund Yes, I was only looking at BCC and FCC phases while assuming interstitials as zero volume species, so I think the site normalization would not have affected my results. I should have thought about it though since that's how CALPHAD works. 

It is odd that VA parameters are given per mole of atoms. I really don't know about the other parameters either. All of my modeling to date has been at atmospheric pressure and my limited attempts to get that portion of the model working failed.

richardotis (approved):
A new test for the symbol replacement behavior would be nice, but I think it's alright to move forward regardless, especially as how properties work will change with the Workspace API PR.

PR #507, opened 2023-12-19
FIX: Minimizer: fixed phases should not have metastable counters increase

bocklund:


richardotis (approved):
Thanks for catching this. I took a quick look at the other uses of `free_stable_compset_indices` and didn't see any where the fixed phases were being incorrectly excluded, so hopefully this was the only one.

Technically, `spec.fixed_stable_compset_indices` is the authoritative source for which compsets are fixed, but this portion of code doesn't have access to the spec, and on initialization the `SystemState` mutates all the compsets to have the `fixed` value corresponding to the spec, so I think it's safe to rely on here.

PR #508, opened 2024-01-17
MAINT/BLD: Python CI support and docs build fixes 

bocklund:
- Add support in CI for Python 3.12
- Drop support for Python 3.7 as it's EOL as of June 2023 https://devguide.python.org/versions/
- Drop a really old pin on furo because we can upgrade now (pin was forcing old sphinx versions to be installed, which is what was breaking our docs builds in CI)
- Fixes for new versions of sphix:
  1. Language is no longer allowed to be `None`, set to `'en'` which is the default
  2. Fix extlinks that requires a `%s` in the caption (added support for that in v4.0, removed support for deprecated way without the `%s` in v6.0 https://github.com/sphinx-doc/sphinx/pull/10471)

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,7 +11,7 @@ jobs:
       max-parallel: 18
       matrix:
         os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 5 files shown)


bocklund:
No changes need for our wheel building flags, which are currently including 3.12 and not including 3.7 anyway

```
      - name: Build wheels
        uses: pypa/cibuildwheel@v2.16.2
        env:
          # PyPy wheels not allowed because SciPy (build requirement) is not available
          CIBW_BUILD: cp3*-*
          CIBW_SKIP: cp35-* cp36-* cp37-* *-musllinux_*
```

richardotis (approved):

richardotis:
Coverage analysis is signalling that the `importlib_metadata` backport can be removed now. I'm fine either way. If it is removed from `__init__.py`, be sure to update the dependencies accordingly.

PR #516, opened 2024-03-06
Fix float exponents when writing a TDB file (TC ERROR DatabaseUtils)

rdamaral:
Hi, this is a small fix to '_print_Piecewise' function in the io.tdb module to improve compatibility with TC. I'm using this module to parse some of my tdbs and TC complains about float exponents (A constant integer must be used as the exponent in exponential TDB expressions). Thank you.

Checklist
* [N/A] If any dependencies have changed, the changes are reflected in the
  * [N/A] `setup.py` (runtime requirements)
  * [N/A] `pyproject.toml` (build requirements)
  * [N/A] `requirements-dev.txt` (build and development requirements)

### DIFF
--- a/pycalphad/io/tdb.py
+++ b/pycalphad/io/tdb.py
@@ -486,7 +486,7 @@ class TCPrinter(object):
     def _print_Piecewise(self, expr):
         # Filter out default zeros since they are implicit in a TDB
         filtered_args = [(x, cond) for x, cond in zip(*[iter(expr.args)]*2) if not ((cond == S.true) and (x == S.Zero))]
-        exprs = [self._stringify_expr(x) for x, cond in filtered_args]
+        exprs = [re.sub(r'\*\*\(([+-]?[0-9]+)\.0\)', r'**(\1)', self._stringify_expr(x)) for x, cond in filtered_args]
         # Only a small subset of piecewise functions can be represented
         # Need to verify that each cond's highlim equals the next cond's lowlim
         # to_interval() is used because symengine does not implement as_set()


richardotis:
Only one test is failing, and I think it's safe for you to modify the pass condition for that test to match this new behavior, since it only differs in the amount of whitespace.

rdamaral:
Hi Richard, I’ve updated ‘test_database.py’, and the test should pass now. I’ve also generalized my previous rules to allow line breaks before any number addition/subtraction that isn’t part of an EXP or LN/LOG operation. The previous rules already accounted for that, but were overlooking some permissible break points.

pycalphad/io/tdb.py
```diff
@@ -541,7 +541,7 @@ def reflow_text(text, linewidth=80):
         else:
             while len(line) > linewidth:
                 linebreak_idx = linewidth - 1
-                while linebreak_idx > 0 and line[linebreak_idx] not in linebreak_chars:
+                while linebreak_idx > 0 and not re.match(r'[^E\(][+-][0-9]', line[linebreak_idx-1:linebreak_idx+2]) and line[linebreak_idx] not in linebreak_chars:
```
richardotis, line 544: Can you clarify why this regex is needed?

rdamaral:
Hi @richardotis. This regex basically matches any addition and subtraction of any number, ie. [+-][0-9], since they are not preceded by 'E' or '('. This should prevent breaking lines inside EXP or LN/LOG or power expression: '6.14599E-07', 'T**(-3)', and 'LOG(-3)', hence splitting these could lead to errors in TC. In practice, this makes the functions being breakable at the `linebreak_chars` plus at [+-][0-9], since they are not preceded by 'E' or '('.

I'll add a comment on the code about this implementation and merge it to https://github.com/pycalphad/pycalphad/pull/522

pycalphad/io/tdb.py
richardotis: See response in https://github.com/pycalphad/pycalphad/pull/516#issuecomment-2186652180

pycalphad/io/tdb.py
```diff
@@ -466,9 +466,19 @@ def _stringify_expr(self, expr):
...
+            #For other types of expr, this shouldn't be an issue since:
+            #    Mul should not need extra parenthesis, e.g. A*(B*C) -> A*B*C
+            #    Pow will explicitly add parenthesis (in the next elif block)
+            #    Other functions such as Log, Sin, etc should
+            #        include the parenthesis when converting to string
+            
+            #All the arguments in Mul should be tested and they're all combined to a single expression by ' * '
+            #    So we could stringify each argument as a list and join them together is ' * '
+            term_list = ['(' + self._stringify_expr(arg) + ')' if isinstance(arg,Add) else self._stringify_expr(arg) for arg in expr.args]
+            terms = ' * '.join(term_list)
```
richardotis, line 481: Is it ok for the spaces in `' * '` to be there?

richardotis:
The test failures are related to https://github.com/pycalphad/pycalphad/pull/530 and can be ignored here.

Can you add a test for `reflow_text` to confirm the line-breaking behavior for floating-point exponents is correct now?

rdamaral:
Sure, I'll add it and let you know.

rdamaral:
Hi @richardotis, just added a new test ``test_reflow_text_for_line_breaks`` for testing the line breaks and updated ``test_tc_printer_exp`` for testing if floating-point exponents are being returned.

PR #523, opened 2024-03-30
BLD: deploy.yaml: PyPI trusted publisher support

richardotis:
This PR adds support for [trusted publishing](https://docs.pypi.org/trusted-publishers/) on PyPI using GitHub Actions as an OpenID Connect provider. I have already performed the setup on the PyPI side to allow for this.

After this is merged, we can revoke the dedicated PyPI upload token for pycalphad and use GitHub Actions as our sole deployment tool. We already do this, but this will prevent a leaked token from being able to manipulate release artifacts out-of-band.


richardotis:
The TestPyPI piece can only be tested on a repo branch, so I will merge to see if this works.

Issue #438, opened 2022-11-19
Pseudo binary phase diagram


XYX224:
May I ask if you have done an example of a pseudo binary phase diagram using this software?

richardotis:
The short answer is not yet. For some pseudo binaries you can hack together a solution by using 'pseudo-components' to behave like pure elements, then just using `binplot`. We do intend to implement this general capability but I don't have a specific timeline for you.

Related: gh-5

bocklund:
Keep this open as a meta issue for pseudo binaries. It's related to #5, but distinct because it's a special case with tie-lines in the plane, and requires support for that special case and special condition support

bocklund:
https://github.com/pycalphad/pycalphad/discussions/302 discussion here has some details

bocklund:
Mapping support is by #543, so what is missing to close this is:

1. The ability to define non-pure element components 
2. The ability to detect the case where for a 3 pure element system where a user has defined a pseudobinary 2 component system (3 pure element system, with 1 molar quantity and 1 potential mapping variables)  or a pseudoternary 3 element system (4 pure element system, with 2 molar quantity mapping variables), detect that the tie-lines are in the plane and use the appropriate mapping strategy

PR #543, opened 2024-07-14
Mapping

bocklund:
This PR adds support for a new mapping system in PyCalphad, in the spirit of the algorithm published by [Sundman _et al._](https://doi.org/10.1016/j.calphad.2021.102330). This PR is the culmination of significant effort and much of the credit to getting it this close to the finish line goes to @nury12n 

## Features

1. The PR adds support for mapping and plotting:
  - 1D stepping in axis variables (e.g. for property diagrams)
  - Binary phase diagrams (to supersede the current `binplot` based on `pycalphad.plot.binary.map`)
  - Ternary isothermal phase diagrams (to supersede the current `ternplot` based on `eqplot`)
  - Isopleth diagrams
2. The visual quality of diagrams is significantly improved:
  - Points on phase boundaries are now connected by solid lines
  - Invariant reactions in binary phase diagrams are computed and drawn now
  - Many pathological cases that gave previous algorithms trouble are resolved, including mapping regions that are very narrow with respect to the conditions, e.g. γ-loops and narrow liquidus/solidus regions. 
3. Provides an API to programmatically access phase boundaries and phase equilibria computed by the mapping algorithm.
4. Provide drop-in replacements to `binplot` and `ternplot` through a compatability API. Old versions of `binplot` and `ternplot` remain available via their absolute imports, but they will eventually be removed. 

## Outstanding TODO items

On the current state of this PR (somewhat outdated comments copied from #517):
- Stepping tends to work well as long as there are no miscibility gaps. With miscibility gaps, mapping can sometimes be flaky or find metastable phase equilibria, especially stepping from 1-phase to 2-phase MG regions.
- Isopleth support exists in principle, but isn't tested yet.
  - Second order transitions? Ex. BCC_A2->BCC_B2 in Fe-Al and Al-Mn. I think you get a node where the order and disorder composition sets are at the same composition but different constituent site fractions
- Global min check can fail to detect phases with high internal DOFs. Possible options:
  - Create DormantPhase for multiple phases with high driving force and then grab the phase with the highest driving force. There is a way to do it right now with the GLOBAL_MIN_NUM_CANDIDATE option, but it's a bit unintuitive where it checks the n samples, then creates a DormantPhase for all unique phases in the n samples, rather than testing n phases.
  - Keep all possible phases as a DormantPhase and update them per iteration. This can be slow though if there are a lot of phases
- Starting point generating using the StepStrategy can be a bit slow, especially for systems with a lot of phases or a lot of complex phases. And occasionally, I've seen it fail for ternaries and isopleths where no starting points can be found.
- Mapping can be sensitive to the ratio between the two axis variable step size, where the wrong direction will be proposed and the zpf line will not be able to be generated since the wrong direction will lead to an equilibrium failure. This is an unfortunate consequence of the new direction testing for binaries and ternaries where the direction is normalized to the step size. I seem to encounter this on systems with low melting points, where I will set the temperature step size too high. It could be possible to set a maximum step size based off the range of the axis variable, but I also don't want to change user inputs behind the scenes.
- Ternaries
  - If the fixed composition set leads to a miscibility gap, then it's possible that the node will be missed when we step in the free composition set and step directly over the node, and the fixed composition set will just move to the other side of the miscibility gap rather than being detected as a new global min
  - We should be able to remove the assumption of the user setting the two axis to (0, 1). This will mainly change how we do the starting points on the third axis. A example where a user may not want the entire composition range could be for a carbon containing system where the carbon content only goes up to a few percent.
- Mapping with a fixed chemical potential/activity/pressure variable?
- Isopleths
  - I have not tested isopleths where the two axes are composition
  - Mapping with linear combinations can lead to many failed equilibrium calculations.
  - A decent number of the isopleths I tested can miss an exit or two from a node
  - Axis limits should be affected by the additional fixed composition, where the axis variable will be limited between (0, 1-sum(fixed compositions)). Currently, the zpf line will stop when it passes the upper limit due to a failed equilibrium calculation (where sum(x) > 1), but it would be better to stop the zpf line before attempting equilibrium. 
- Unit tests
  - There are unit tests for the checks that are performed after each zpf line iteration
  - I suppose for the strategies, we can pick a simple system for each and make sure it generates the correct zpf lines and nodes, testing only for the phases that are present in each zpf line.
  - I'll have to see if it's possible to set up a strategy at a specific point, such as just before a node, then it would be possible to have unit tests for specific conditions (this could be like finding a node, checking the node exits, reaching an axis limit, etc)
- (from #517, maybe outdated) Giving components to `Mapper` that do not include vacancies seems to produce silent failures where no boundaries are produced.
- (from #517, maybe outdated) Automated tests exercising the basic behavior to achieve coverage need to be implemented


pycalphad/mapping/primitives.py
```diff
@@ -0,0 +1,398 @@
...
+    A Node is a special case of a Point that indicates a set of conditions
+    corresponding to a phase change.
+
+    It also keeps track of the parent Point that it was created from, so that
+    users of the class can determine which exits from the node are novel.
+
+    Compared to its parent, the set of conditions should be the same (both keys
+    and values) and the set of stable phases should differ by exactly one phase.
+
+    We"ll keep track of the axis variable and direction as well
```
bocklund, line 239:
```suggestion
    We'll keep track of the axis variable and direction as well
```

pycalphad/mapping/starting_points.py
```diff
@@ -0,0 +1,103 @@
...
+    for key in copy_conds:
+        copy_conds[key] = key.compute_property(comp_sets, copy_conds, point.chemical_potentials)
+
+    new_point = Point(copy_conds, point.chemical_potentials, [cs for cs in comp_sets if cs.fixed], [cs for cs in comp_sets if not cs.fixed])
+    return new_point
+        
+
+
+
+
```
bocklund, line 103:
```suggestion

```

pycalphad/mapping/strategy/binary_strategy.py
```diff
@@ -0,0 +1,223 @@
...
+        super().__init__(dbf, components, phases, conditions, **kwargs)
+
+    def initialize(self):
+        """
+        Searches axis limits to find starting points
+
+        Here, we do a step mapping along the axis bounds and grab all the nodes
+        The nodes of a step map is distinguished from starting points in that they have a parent
+
+        TODO: we should check performance to see if step mapping versus running multiple equilibrium along the axis then searching through the array is faster
```
bocklund, line 78: Resolve TODO

pycalphad/mapping/strategy/binary_strategy.py
```diff
@@ -0,0 +1,223 @@
...
+                step.do_map()
+                self._add_starting_points_from_step(step)
+
+    def _add_starting_points_from_step(self, step: StepStrategy):
+        """
+        Grabs starting points from a step calc
+            For stepping in a state variable (T or P), this is all the nodes of the step calc
+            For stepping in composition, this is all the 2 phase regions
+        TODO: Do we really need separate methods for whether we're stepping in a state variable or not? or
+              can we just use the method for stepping in composition?
```
bocklund, line 111: Resolve TODO

pycalphad/mapping/strategy/isopleth_strategy.py
```diff
@@ -0,0 +1,286 @@
...
+    # Only return axis variable to step against
+    if best_index == -1:
+        return options_tests[0][1]
+    else:
+        return options_tests[best_index][1]
+
+class IsoplethStrategy(MapStrategy):
+    """
+    TODO: initialize, _determine_start_direction and _test_swap_axis are the same as for binary strategy
+            We may want to merge some of these methods to a single class that isopleths and binary strategies can inherit from
```
bocklund, line 54: Resolve TODO

pycalphad/mapping/strategy/isopleth_strategy.py
```diff
@@ -0,0 +1,286 @@
+from typing import Union
+import logging
+import itertools
+import copy
+
+import numpy as np
+
+from pycalphad import Database, variables as v
```
bocklund, line 8: Isopleth needs an example

bocklund, line 8: Not sure exactly where this code should live, but I would like to be able to detect a pseudo binary case, where there's more than 2 components, but the tie-lines are in plane of the calculation (maybe we need to switch to binary strategy in that case?). 

Would be good if there's an example too. See https://github.com/pycalphad/pycalphad/issues/438 for some details

pycalphad/mapping/strategy/step_strategy.py
```diff
@@ -0,0 +1,170 @@
...
+    def __init__(self, dbf: Database, components: list[str], phases: list[str], conditions: dict[v.StateVariable, Union[float, tuple[float]]], **kwargs):
+        super().__init__(dbf, components, phases, conditions, **kwargs)
+
+    def initialize(self):
+        """
+        Adds initial starting point in middle of free condition
+        """
+        av = self.axis_vars[0]
+        mid_val = (self.axis_lims[av][0] + self.axis_lims[av][1]) / 2
+        mid_conds = copy.deepcopy(self.conditions)
```
bocklund, line 27: Need a stepping example. I'm sort of indifferent of whether we use the class API or `stepplot`, but depends on the other comment on whether we keep `stepplot`

pycalphad/mapping/strategy/strategy_base.py
```diff
@@ -0,0 +1,507 @@
...
+        """
+        Creates a copy of condition and steps in proposed direction
+
+        If the axis variable is outside the defined limits, then we adjust the conditions to be at the edge of the axis limits
+        """
+        new_conds = copy.deepcopy(point.global_conditions)
+        new_conds[axis_var] += axis_delta*direction.value
+
+        # Limit axis variable to axis limits (provide a small offset for composition)
+        # TODO: check if composition sums to more than 1, could be an override for isopleth strategy
```
bocklund, line 245: Resolve TODO

pycalphad/mapping/strategy/strategy_base.py
```diff
@@ -0,0 +1,507 @@
...
+        if direction_data is None:
+            return
+
+        # Create node from exit and start direction
+        exit_point, start_ax, start_dir, start_delta = direction_data
+        # start_node = self._create_node_from_point(self._exits[self._exit_index], None, start_ax, start_dir)
+
+        # Initialize zpf line
+        # self.zpf_lines.append(ZPFLine(start_node.fixed_phases, start_node.free_phases))
+        # self.zpf_lines[-1].points.append(start_node)
```
bocklund, line 391:
```suggestion
```

pycalphad/mapping/strategy/strategy_base.py
```diff
@@ -0,0 +1,507 @@
...
+        # Assume that we can step using the specified axis delta,
+        # but, if stepping fails, then we reduce the step size until
+        # it works. If the step size becomes too small, then we stop
+
+        # Temporary, we could default the starting delta to be the minimum, then scale up to the
+        # input delta after successful zpf line steps. This could avoid some extra equilibrum calcs
+        # If this works better then testing and scaling down the direction, then we don't need the while loop
+        #  It would also be easier to return a failure mode for when the direction testing failed (non-convergence, not global min, etc)
+
+        # curr_delta = self.axis_delta[axis_var]
```
bocklund, line 471:
```suggestion
```

pycalphad/mapping/strategy/ternary_strategy.py
```diff
@@ -0,0 +1,396 @@
...
+
+    if best_index == -1:
+        return options_tests[0][0], options_tests[0][1], options_tests[0][2], normal
+    else:
+        return options_tests[best_index][0], options_tests[best_index][1], options_tests[best_index][2], normal
+
+class TernaryStrategy(MapStrategy):
+    def __init__(self, dbf: Database, components: list[str], phases: list[str], conditions: dict[v.StateVariable, Union[float, tuple[float]]], **kwargs):
+        super().__init__(dbf, components, phases, conditions, **kwargs)
+        # TODO: I don't really like this since this assumes pure elements
```
bocklund, line 107: Resolve TODO

pycalphad/mapping/strategy/ternary_strategy.py
```diff
@@ -0,0 +1,396 @@
...
+        map_kwargs = self._constant_kwargs()
+
+        # Iterate through axis variables, and set conditions to fix axis variable at min only
+        for av in self.axis_vars:
+            conds = copy.deepcopy(self.conditions)
+            conds[av] = np.amin(self.axis_lims[av])
+
+            # other_av = self._other_av(av)
+            # av_range = np.amax(self.axis_lims[other_av]) - np.amin(self.axis_lims[other_av])
+            # conds[other_av] = (self.axis_lims[other_av][0], self.axis_lims[other_av][1], av_range/20)
```
bocklund, line 128:
```suggestion
```

pycalphad/mapping/strategy/ternary_strategy.py
```diff
@@ -0,0 +1,396 @@
...
+            step.do_map()
+            self._add_starting_points_from_step(step)
+
+        # Additional step where we switch axis conditions to the unlisted variable
+        conds = copy.deepcopy(self.conditions)
+        conds[self.all_vars[-1]] = MIN_COMPOSITION
+        del conds[self.axis_vars[0]]
+
+        # av_range = np.amax(self.axis_lims[self.axis_vars[1]]) - np.amin(self.axis_lims[self.axis_vars[1]])
+        # conds[self.axis_vars[1]] = (self.axis_lims[self.axis_vars[1]][0], self.axis_lims[self.axis_vars[1]][1], av_range/20)
```
bocklund, line 147:
```suggestion
```

pycalphad/mapping/strategy/ternary_strategy.py
```diff
@@ -0,0 +1,396 @@
...
+            If subset, then set the last zpf point as a parent and add new node like normal
+            If not, then decrease step size and try stepping again
+              If step size is too small, then fail zpf line and add node as a new starting point
+        """
+        # Do a global equilibrium check before attempting to add the new node
+        global_eq_check, test_point = self._check_full_global_equilibrium(new_node, add_global_point_if_false = False)
+        if test_point is None:
+            _log.info("Global eq check could not be determined")
+            # zpf_line.status = ZPFState.FAILED
+            # return
```
bocklund, line 326:
```suggestion
```

pycalphad/mapping/strategy/ternary_strategy.py
```diff
@@ -0,0 +1,396 @@
...
+                zpf_line_phases = zpf_line.stable_phases_with_multiplicity
+                test_node_phases = test_point.stable_phases_with_multiplicity
+                _log.info(f"Global eq check failed. Comparing zpf line {zpf_line_phases} with node phases {test_node_phases}")
+                if len(set(test_node_phases) - set(zpf_line_phases)) == 1:
+                    _log.info("Node can be added as a proper node")
+                    new_node = self._create_node_from_point(test_point, zpf_line.points[-1], None, None)
+                    return super()._process_new_node(zpf_line, new_node)
+                else:
+                    # zpf_line.status = ZPFState.FAILED
+                    # return
```
bocklund, line 353:
```suggestion
```

pycalphad/mapping/utils.py
```diff
@@ -0,0 +1,70 @@
...
+    for cs in new_point._free_composition_sets:
+        if bias_towards_free:
+            new_amt = 1 if cs.fixed else 2
+            update_cs_phase_frac(cs, new_amt/phase_sum)
+        else:
+            update_cs_phase_frac(cs, 1/phase_sum)
+        cs.fixed = False
+
+        # cs.fixed = False
+        # update_cs_phase_frac(cs, 1/len(new_point._free_composition_sets))
```
bocklund, line 67:
```suggestion
```

pycalphad/mapping/utils.py
```diff
@@ -0,0 +1,70 @@
...
+    non_va_comps = len(set(components) - {"VA"})
+    return non_va_comps + 2 - len(point.stable_composition_sets) - (2 - num_potential_conditions)
+
+def update_cs_phase_frac(comp_set: CompositionSet, phase_frac: float):
+    comp_set.update(comp_set.dof[len(STATEVARS):], phase_frac, comp_set.dof[:len(STATEVARS)])
+
+def get_statevars_array(conditions: dict[v.StateVariable, float]):
+    return np.asarray([conditions[sv] for sv in STATEVARS], dtype=np.float64)
+
+def elements_from_components(components: list[str]):
```
bocklund, line 20: There used to be pycalphad code for getting nonvacant pure elements, might be good to use that instead if we can

pycalphad/mapping/zpf_checks.py
```diff
@@ -0,0 +1,286 @@
...
+            zpf_line.status = ZPFState.FAILED
+            return None
+        zpf_line.status = ZPFState.ATTEMPT_NEW_STEP
+    return None
+
+def _check_axis_values_within_limit(zpf_line: ZPFLine, prev_point_vars: dict[v.StateVariable, float], new_point_vars: dict[v.StateVariable, float], axis_data: Mapping, **kwargs):
+    """
+    Checks that axis values are within the axis limits
+
+    TODO: should check if we really need x_offset, this could affect phases like GRAPHITE which has a composition of X(C)=1
```
bocklund, line 83: Resolve TODO

pycalphad/mapping/zpf_checks.py
```diff
@@ -0,0 +1,286 @@
...
+        return None
+
+    if not _check_axis_values_within_limit(zpf_line, prev_point_vars, new_point_vars, axis_data, **kwargs):
+        _log.info("Variable reach axis limit. Ending ZPF line")
+        zpf_line.status = ZPFState.REACHED_LIMIT
+        return None
+    # if not _check_composition_within_limit(zpf_line, step_results):
+    #    _log.info("Variable reach composition limit. Ending ZPF line")
+    #    zpf_line.status = ZPFState.REACHED_LIMIT
+    #    return None
```
bocklund, line 159: Resolve TODO

pycalphad/mapping/zpf_checks.py
```diff
@@ -0,0 +1,286 @@
...
+            else:
+                new_node = zeq.create_node_from_different_points(global_test_point, new_point.stable_composition_sets, axis_vars)
+
+            if new_node is not None:
+                new_node_vars = {av: new_node.get_property(av) for av in axis_vars}
+                # Check that new node satisfy axis limits and distance between nodes
+                # If not, then the zpf line ends unexpectedly
+                check_axis = _check_axis_values_within_limit(zpf_line, new_point_vars, new_node_vars, axis_data, **kwargs)
+                check_dist =_check_axis_values_by_distance(zpf_line, new_point_vars, new_node_vars, axis_data, **kwargs)
+                # check_dist = True
```
bocklund, line 255:
```suggestion
```

pycalphad/mapping/zpf_equilibrium.py
```diff
@@ -0,0 +1,232 @@
...
+    For each possible phase:
+        1. Sample DOF and find CS that minimizes driving force
+        2. Create a DormantPhase with CS and compute driving force with potentials at equilibrium
+        3. If driving force is negative, then new phase is stable
+        4. Check that the new CS doesn"t match with a currently stable CS
+        4. Hope that this works on miscibility gaps
+
+    This should take care of the DLASCLS error since we compute the new phase separately so if
+    the composition clashes with a fixed phase, we check that afterwards before attempting to
+    run equilibrium on two CS with the same composition
```
bocklund, line 66: Maybe revise comment since pycalphad eliminated DLASCLS errors from showing. We probably still need this mitigation in the code itself to prevent ourselves from getting into a spot that would trigger the DLASCLS error (that PyCalphad protects us from)

pycalphad/mapping/zpf_equilibrium.py
```diff
@@ -0,0 +1,232 @@
...
+    tested_phases = []
+    for i in range(np.amin([num_candidates, len(sorted_indices)])):
+        index = sorted_indices[i]
+        if phase_ids[index] not in tested_phases:
+            tested_phases.append(phase_ids[index])
+            test_cs = CompositionSet(phase_records[phase_ids[index]])
+            # TODO: on some models, I get an error of
+            # ValueError: buffer source array is read-only (line 59 in pycalphad.core.composition_set.CompositionSet.update)
+            # Databases where this occured in test suite: alcfe_b2.tdb running Al-C-Fe isopleth, alcocrni.tdb running Al-Co-Ni ternary
+            # and femns.tdb with ionic liquid model
```
bocklund, line 91: Resolve TODO if we can. I have a list of some binaries that trigger the exceptions, not sure if this one is included. 

This is a good thing to put in the test suite and fix to make sure that changes we make don't trigger known bugs in the future

pycalphad/mapping/zpf_equilibrium.py
```diff
@@ -0,0 +1,232 @@
...
+from pycalphad.property_framework.computed_property import JanssonDerivative
+
+from pycalphad.mapping.primitives import STATEVARS, Node, Point
+import pycalphad.mapping.utils as map_utils
+
+_log = logging.getLogger(__name__)
+
+def update_equilibrium_with_new_conditions(point: Point, new_conditions: dict[v.StateVariable, str], free_var: v.StateVariable = None):
+    """
+    Pretty much a copy of the old update with new conditions
```
bocklund, line 20: Update docstring

pycalphad/mapping/zpf_equilibrium.py
```diff
@@ -0,0 +1,232 @@
...
+
+    Returns
+    -------
+    new_point - Point with updated composition sets and conditions
+        Checking whether the number of composition sets has changed will be checked later
+    None - equilibrium failed
```
bocklund, line 26:
```suggestion
    new_point
        Point with updated composition sets and conditions
        Checking whether the number of composition sets has changed will be checked later
    None
        Equilibrium failed
```

pycalphad/mapping/zpf_equilibrium.py
```diff
@@ -0,0 +1,232 @@
...
+_log = logging.getLogger(__name__)
+
+def update_equilibrium_with_new_conditions(point: Point, new_conditions: dict[v.StateVariable, str], free_var: v.StateVariable = None):
+    """
+    Pretty much a copy of the old update with new conditions
+
+    Returns
+    -------
+    new_point - Point with updated composition sets and conditions
+        Checking whether the number of composition sets has changed will be checked later
```
bocklund, line 25: > Checking whether the number of composition sets has changed will be checked later

Maybe rephrase this to make it more clear that counting stable composition sets is out of scope for this function

pycalphad/mapping/compat_api.py
```diff
@@ -0,0 +1,218 @@
...
+    Returns
+    -------
+    Axes
+        Matplotlib Axes of the phase diagram
+    (Axes, BinaryStrategy)
+        If return_strategy is True.
+
+    """
+    indep_comps = [key for key, value in conditions.items() if key not in STATEVARS and len(np.atleast_1d(value)) > 1]
+    indep_pots = [key for key, value in conditions.items() if key in STATEVARS and len(np.atleast_1d(value)) > 1]
```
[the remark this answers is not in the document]
nury12n: Are you referring to pycalphad.core.utils.get_state_variables? 

There's a lot of references in the mapping code to the hard-coded STATEVARS list which I think the final solution here would be to remove STATEVARS in favor of getting state variables from get_state_variables in case a user defines a custom state variable. An example this could benefit could be interfacial contribution for Gibbs-Thompson phase diagrams, in which a hard-coded STATEVARS list would not allow for.

pycalphad/mapping/plotting.py
```diff
@@ -0,0 +1,339 @@
...
+from pycalphad.mapping.strategy.isopleth_strategy import IsoplethStrategy
+import pycalphad.mapping.utils as map_utils
+
+def _get_label(var: v.StateVariable):
+    # If user just passes v.NP rather than an instance of v.NP, then label is just NP
+    if var == v.NP:
+        return 'Phase Fraction'
+    # Otherwise, we can just use the display name
+    else:
+        return var.display_name
```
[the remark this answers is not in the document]
nury12n: For v.T and v.P, the display name returns Temperature and Pressure. For units, I may want to wait until we work out the Renderers.

If the var is something like v.X('FCC_A1', 'AL'), I suppose the phase name should remain all caps and displace X(FCC_A1, Al)?

For chemical potential, should it be MU(Al) or Chemical Potential (Al)?

pycalphad/mapping/plotting.py
```diff
@@ -0,0 +1,339 @@
...
+        x = strategy.axis_vars[0]
+    if x == strategy.axis_vars[0]:
+        x_is_global = True
+    # If y is None, then use v.NP
+    if y is None:
+        y = v.NP
+
+    step_data = _get_step_data(strategy, x, y, x_is_global)
+    data = step_data['data']
+    phases = step_data['phases']
```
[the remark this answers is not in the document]
nury12n: This is removed, since the phases are stored in step_data['data']

pycalphad/__init__.py
```diff
@@ -15,10 +15,11 @@
 from pycalphad.core.calculate import calculate
 from pycalphad.core.equilibrium import equilibrium
 from pycalphad.core.workspace import Workspace
-from pycalphad.plot.binary import binplot
-from pycalphad.plot.ternary import ternplot
+# from pycalphad.plot.binary import binplot
+# from pycalphad.plot.ternary import ternplot
```
bocklund, line 19:
```suggestion
```

pycalphad/mapping/zpf_equilibrium.py
nury12n: Note: This function returns the original list of composition sets along with the point.

pycalphad/mapping/zpf_equilibrium.py
nury12n: See above comment

pycalphad/mapping/compat_api.py
```diff
@@ -0,0 +1,218 @@
+import numpy as np
```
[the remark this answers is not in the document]
richardotis: Ok with both of these for now, but we may need to reconsider the API as we work toward an integrated approach to visualization with Workspace.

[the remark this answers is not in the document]
richardotis: Per offline discussion, these functions are pretty thin wrappers, so we should hold off until we figure out an integrated approach to visualization with Workspace.

pycalphad/mapping/zpf_checks.py
```diff
@@ -0,0 +1,466 @@
...
+                new_node = None
+            else:
+                new_node = zeq.create_node_from_different_points(global_test_point, new_point.stable_composition_sets, axis_vars)
+
+            if new_node is not None:
+                new_node_vars = {av: new_node.get_property(av) for av in axis_vars}
+                # Check that new node satisfy axis limits and distance between nodes
+                # If not, then the zpf line ends unexpectedly
+                check_axis = _check_axis_values_within_limit(zpf_line, new_point_vars, new_node_vars, axis_data, **kwargs)
+                check_dist =_check_axis_values_by_distance(zpf_line, new_point_vars, new_node_vars, axis_data, **kwargs)
```
richardotis, line 420: Can we check against previously removed nodes somehow?

pycalphad/mapping/zpf_equilibrium.py
```diff
@@ -0,0 +1,328 @@
...
+        cs.update(cs.dof[len(STATEVARS):], cs.NP, new_state_conds)
+
+    # Remove free variable condition if given - this assumes that Gibbs phase rule will be satisfy if done
+    if free_var is not None:
+        del new_conditions[free_var]
+
+    # Keep track of original composition sets (these will be updated with the solver, but the original list will remain even if a phase becomes unstable)
+    orig_cs = [cs for cs in comp_sets]
+    try:
+        solver = Solver(remove_metastable=True, allow_changing_phases=False)
```
richardotis, line 58:
```suggestion
        solver = Solver(remove_metastable=True)
```

richardotis: `allow_changing_phases` kwarg doesn't exist

pycalphad/mapping/zpf_equilibrium.py
```diff
@@ -0,0 +1,328 @@
...
+
+    # Setup conditions and remove axis variable to free it
+    solution_cs = [cs for cs in new_cs]
+    new_conditions = copy.deepcopy(new_point.global_conditions)
+    for av in axis_vars:
+        del new_conditions[av]
+
+    # Solve equilibrium with fixed CS
+    try:
+        solver = Solver(remove_metastable=True, allow_changing_phases=False)
```
richardotis, line 278:
```suggestion
        solver = Solver(remove_metastable=True)
```

pycalphad/model.py
```diff
@@ -436,6 +436,8 @@ def degree_of_ordering(self):
     #pylint: disable=C0103
     # These are standard abbreviations from Thermo-Calc for these quantities
     energy = GM = property(lambda self: self.ast)
+    energy_implementation_units = GM_implementation_units = 'J / mol'
+    energy_display_units = GM_display_units = 'kJ / mol'
```
richardotis, line 440:
```suggestion
```

pycalphad/mapping/primitives.py
```diff
@@ -0,0 +1,426 @@
...
+        curr_np = comp_set.NP
+        # Can't use map utils here due to circular dependencies
+        comp_set.update(comp_set.dof[len(STATEVARS):], 1.0, comp_set.dof[:len(STATEVARS)])
+
+        prop_value = var.compute_property([comp_set], self.global_conditions, self.chemical_potentials)
+
+        # Restore phase fraction
+        comp_set.update(comp_set.dof[len(STATEVARS):], curr_np, comp_set.dof[:len(STATEVARS)])
+
+        return np.squeeze(prop_value)
```
richardotis, line 253:
```suggestion
        return prop_value
```

richardotis (changes requested):

pycalphad/mapping/compat_api.py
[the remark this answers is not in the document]
nury12n: These are removed with examples for how to use and plot the strategies

pycalphad/mapping/zpf_checks.py
nury12n: I suppose rather than removing them, we can just flag them as bad nodes (which will be ignored during plotting) in the NodeQueue, so the NodeQueue will check against all existing nodes when a new node is attempted to be added. The only thing for binary cases is that we would need a way to determine that the node is metastable and flag them.

pycalphad/mapping/zpf_equilibrium.py
nury12n: This seems to be due to the site fraction array from calculate being non-writable for some models. I'm not sure why this occurs, but the simple fix for mapping purposes is to create a new numpy array for the site fractions before updating the composition set since the new array will be writable.

Example of the site fractions array from femns.tdb being read-only.
```python
db = Database('femns.tdb')
comps = ['FE', 'MN', 'S', 'VA']
phases = list(db.phases.keys())
points = calculate(db, comps, phases, T=1000, P=101325, to_xarray=False)
y = np.squeeze(points.Y)
print('femns.tdb')
print('From np.squeeze: writeable =', y.flags.writeable)
y = np.array(y, dtype=np.float64)
print('From np.array writeable =', y.flags.writeable)

db = Database('crtiv_ghosh.tdb')
comps = ['CR', 'TI', 'V', 'VA']
phases = list(db.phases.keys())
points = calculate(db, comps, phases, T=1000, P=101325, to_xarray=False)
y = np.squeeze(points.Y)
print('\ncrtiv_ghosh.tdb')
print('From np.squeeze writeable =', y.flags.writeable)
y = np.array(y, dtype=np.float64)
print('From np.array writeable =', y.flags.writeable)
```

Output
```
femns.tdb
From np.squeeze: writeable = False
From np.array writeable = True

crtiv_ghosh.tdb
From np.squeeze writeable = True
From np.array writeable = True
```

pycalphad/mapping/strategy/binary_strategy.py
nury12n: I think this TODO should be resolved. From the SGTE binaries, it seems like most binaries are finishing under a minute.

pycalphad/mapping/strategy/ternary_strategy.py
nury12n: I think it's okay to assume pure elements for now, and wait until the general Components are supported

pycalphad/mapping/strategy/isopleth_strategy.py
bocklund: Technically we don't need generalized components, because we can still have pseudo binary cases like (Hf,Ta)C or (Hf,Ta)B2 that don't require them. However, I think we should defer and merge and open a separate issue.

pycalphad/mapping/compat_api.py
```diff
@@ -0,0 +1,111 @@
...
+    -------
+    Axes
+        Matplotlib Axes of the phase diagram
+    (Axes, TernaryStrategy)
+        If return_strategy is True.
+
+    """
+    # remaining plot_kwargs from pycalphad.plot.eqplot
+    # x=None, y=None, z=None, tieline_color=(0, 1, 0, 1), tie_triangle_color=(1, 0, 0, 1), **kwargs
+    # kwargs passed ot ax.scatter
```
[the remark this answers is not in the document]
nury12n: I didn't even notice the z variable this whole time. 

One thing that always bugged me between binplot and ternplot was binplot used (plot_kwargs = None, **map_kwargs) and ternplot used (map_kwargs=None, **plot_kwargs) in order to retain the same API in the previous binplot and ternplot functions. It seemed inconsistent to have to input a dictionary to change the tieline colors in binplot while I only needed an keyword argument for ternplot. I suppose it's okay though if we're moving away from binplot/ternplot and towards a `Renderer` API.

Regarding this difference, ternplot will have `tieline_color` and `tie_triangle color` as arguments to the docstring, but binplot will mention these in the docstrings as variables to be added to the plot_kwargs dictionary. I will note that tie_triangle_color is a new argument for binplot since we can plot invariants now.

pycalphad/mapping/primitives.py
```diff
@@ -0,0 +1,425 @@
...
+    Exit rules
+
+    NORMAL - will search for all viable exits from node
+             ignores the exit that corresponds to the ZPF line that found the node
+    POINT_IS_EXIT - will use the point composition sets as the exit
+                    this is mainly used for starting points since the point may not necessary fit the conditions for a "node" (0 DOF)
+    FORCE_ALL_EXITS - similar to normal, but will include every possible exit from node
+                      not sure what the use case for this is, seems to have been for ad-hoc fix that is no longer needed?
+                      Only thing I can think of is for a starting point on a node for tielines or isopleths strategies
+    """
```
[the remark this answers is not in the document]
nury12n: Looks like there's no where where a Node.exit_hint is set to FORCE_ALL_EXITS. Removing this flag.

pycalphad/mapping/compat_api.py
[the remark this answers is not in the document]
bocklund: I'm not sure how I managed to make those perfectly inconsistent, but I'm happy with keeping it how it is for now, to maintain compatibility with existing code

PR #546, opened 2024-07-14
ENH: Add Component implementation in variables

bocklund:
This PR sets the groundwork for more formally distinguishing a `Component` from a `Species`. It is breaking for the public API of Workspace, and consumers of the private API for `pycalphad.core.utils.unpack_components`. We are making these breaking changes now, so future changes that implement generalized (non-pure-element) component support can be non-breaking or minimally breaking in public APIs.

The following changes are made:
1. Add `v.Component` class. This is similar to `Species`, but distinct because 1) `Species` are reserved to refer to phase constituents, and 2) `Species` are allowed to have charges
2. Rename `pycalphad.core.utils.unpack_components` to `unpack_species` to be more accurate
3. Add a `v.unpack_components` function to process components following documented rules comparable to those in commercial software packages
4. Modify `Workspace` API:
  4.1. Rename `Workspace.components`, which is a container of `Species`, objects to `Workspace.constituents`. `Workspace.constituents` automatically update when `Workspace.components` change. Currently the constituents are not used in the `Workspace` implementation and changes from users don't affect any calculations, but it might be possible to use this to suspend certain species. 
  4.2. Add `ComponentsList` and `ComponentsField` implementations `Workspace`, as a container of `Component` objects in `Workspace.components` with the ability to auto-populate from `Database` pure elements and update when the database changes.

What is _not_ in this PR is explicit support for non-pure element components as a basis for use in conditions, properties, etc. Non-pure element components is a driver for the `Component` object to exist, but support is not added here.

### DIFF
--- a/pycalphad/variables.py
+++ b/pycalphad/variables.py
@@ -3,6 +3,7 @@
 Classes and constants for representing thermodynamic variables.
 """
 
+from typing import Optional
 from symengine import Float, Symbol
 from pycalphad.io.grammar import parse_chemical_formula
 from pycalphad.property_framework.types import JanssonDerivativeDeltas
@@ -199,7 +200,7 @@ class Species(object):
         return hash(self.name)
 
 
-def unpack_components(dbf: "Database", components: "Sequence[Component | Species | str]"):
+def unpack_components(components: "Sequence[Component | Species | str]", dbf: Optional["Database"] = None):
     """
     Build a set of Components from ones provided by the caller.
 
@@ -212,10 +213,12 @@ def unpack_components(dbf: "Database", components: "Sequence[Component | Species
 
     Parameters
     ----------
-    dbf : Database
-        Thermodynamic database containing elements and species.
     comps : Sequence[Component | Species | str]
         Names of components to consider in the calculation.
+    dbf : Optional[Database]
+        Thermodynamic database containing elements and species. If no database is
+        passed, constructing Component objects via pattern matching to Species defined
+        in the Database (3a) will not be possible.
 
     Returns
     -------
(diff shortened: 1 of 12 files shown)


bocklund:
Conceptually, this PR seeks to add the left hand side of this diagram. Before this PR, PyCalphad only uses pure elements as the composition basis set

ComponentsVennDiagram

bocklund:
I think there's value to trying to get something merged that has `v.Component` support without necessarily having any linear combination stuff implemented for non-pure element `v.Component` instances.

`as_property` also has at least some limited LinearCombination support, so getting something like `v.MU(v.Component("AB2"))` does produce a linear combination of `MU(A) + 2*MU(B)` (which isn't right for the chemical potentials, which should be normalized per mole)

bocklund:
Given the context of just getting something in, I'm ready for some feedback @richardotis 

I'm not entirely happy with how things are done. Maybe we need a `ComponentBase` or something. Separating the semantics of `Species := PhaseConstituents` and `Component := CompositionBasis` throughout all of the variables isn't done here yet, but I think the types are mostly correct.

There's probably further distinction to make in `_expand_property_arguments` between properties that are composition-dependent and should be using `Component`, and properties that are site fraction-dependent, and should be using `Species`. I guess the heuristic now is that `hasattr(sublattice_index)` is `True` only for things that should be `Species`. Using `species.name` or `component.name` to check for wildcards should be fine I think.

bocklund:
The final version of this PR should close https://github.com/pycalphad/pycalphad/issues/226, which will require some changes to the API documentation and variables, and overall just thinking through the legacy code and whether it makes sense to take in `Component` or take in `Species` as the argument, and which to use inside each function/object, and make sure the corresponding calls to `unpack_components` or `unpack_species` is called.

If we're in the breaking spirit, I am also open to feedback on the renaming the current `Species` object to `PhaseConstituent` (or just `Constitutient`) and `unpack_species` to `unpack_phase_constituents` (or `unpack_constituents`). Maybe that would enable the class heirarchy to be

```python
class Species(object):
    ...

class Component(Species):
    ...

class PhaseConstituent(Species):
   ...
# or 
class Constituent(Species):
   ...
```

pycalphad/variables.py
```diff
@@ -378,6 +532,7 @@ def compute_property(self, compsets, cur_conds, chemical_potentials):
         result = np.atleast_1d(np.zeros(self.shape))
         result[:] = np.nan
         for _, compset in self.filtered(compsets):
+            print(compset.phase_record.phase_name, compset.phase_record.nonvacant_elements)
```
richardotis, line 535: print statement should be removed

pycalphad/variables.py
```diff
@@ -634,7 +789,7 @@ class ChemicalPotential(StateVariable):
     display_name = property(lambda self: f'Chemical Potential {self.species}')
 
     def __init__(self, species):
-        species = Species(species)
+        species = Component(species)
```
richardotis, line 792: We're leaving this as 'species' for the moment but will fix it in a future revision?

richardotis (approved):

pycalphad/variables.py
bocklund: I think so. I'm open to changing it

PR #549, opened 2024-07-24
FIX: tdb: Export TDB with nested powers and non-integer exponents

richardotis:
This PR will make TDBs correctly export when expressions with nested powers (e.g., `a**x**y`) are in the TDB.

Example:
```python
from pycalphad import Database
dbf = Database("""
ELEMENT A FCC_A1 0 0 0 !

PHASE FCC_A1 % 1 1 !
CONSTITUENT FCC_A1 : A : !

PARAMETER G(FCC_A1,A;0) 1 ((3.49 * (1373 * T**(-1)))**(1.778 * (1473 * T**(-1))))**(0.926); 10000 N !
"""
)
print(dbf.to_string(fmt='tdb'))
```

Current Behavior (Wrong, truncating the non-integer exponent):
```
$ [snip]
PARAMETER G(FCC_A1,A;0) 1.0 ( 4791.77*T**(-1) )**(2618.994*T**(-1))**(0);
   10000.0 N !
```

New Behavior (Correct):
```
$ [snip]
PARAMETER G(FCC_A1,A;0) 1.0 ( ( 4791.77*T**(-1) )**(2618.994*T**(-1))
   )**(0.926); 10000.0 N !

```

In addition, the behavior for non-integer exponents is now controllable via the `if_incompatible` keyword argument, e.g., `Database.to_file(..., fmt='tdb', if_incompatible=...)`. If set to `ignore`, the new behavior is produced. If set to `warn` (still the default), the new behavior is produced but a warning is emitted. If set to `raise` or `fix`, a `DatabaseExportError` is raised (since `fix` cannot do anything for this incompatibility).

### DIFF
--- a/pycalphad/io/tdb.py
+++ b/pycalphad/io/tdb.py
@@ -10,7 +10,7 @@ from pyparsing import delimitedList, ParseException
 import re
 from symengine.lib.symengine_wrapper import UniversalSet, Union, Complement
 from symengine import sympify, And, Or, Not, EmptySet, Interval, Piecewise, Add, Mul, Pow
-from symengine import Float, Symbol, RealDouble, LessThan, StrictLessThan, S, E
+from symengine import Float, Symbol, LessThan, StrictLessThan, S, E
 from tinydb import where
 from pycalphad import Database
 from pycalphad.io.database import DatabaseExportError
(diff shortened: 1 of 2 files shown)


bocklund (approved):

bocklund:
LGTM. Would it make sense to emit a warning if we can't coerce the exponent to a integer that the database may be incompatible with other software?

PR #554, opened 2024-08-03
ENH: Workspace: Suspend attribute dependency updates during init

richardotis:
This PR changes how Workspace objects are initialized, to ensure that each attribute is initialized after its dependencies (`database` before `models`, `parameters` before `phase_record_factory`, etc). It also suspends dependency callbacks inside of `Workspace.__init__` to avoid calls to the `default_factory` of attributes that will be overwritten again later in initialization.

For the worst case of a user passing in fully-initialized attributes to the constructor of a Workspace object, with pre-computed `models` and `phase_record_factory` (a common case inside of ESPEI and any other software calling pycalphad in a high-throughput loop), this PR should substantially improve performance.

We can be a little smarter still, by eliding dependency updates for cases where the change in the dependency doesn't actually affect the dependent object; for example, `phase_record_factory` only really depends on the keys of `conditions`, not the values, so we check for that in `PRFField.on_dependency_update` and avoid rebuilding it if they haven't changed.

### DIFF
--- a/pycalphad/core/workspace.py
+++ b/pycalphad/core/workspace.py
@@ -123,7 +123,8 @@ class TypedField:
 
     def on_dependency_update(self, obj, updated_attribute, old_val, new_val):
         "A callback method that can be overridden to define custom behavior when a dependent attribute is updated."
-        pass
+        if obj._suspend_dependency_updates:
+            return
 
 class ComponentsField(TypedField):
     def __init__(self, depends_on=None):
@@ -211,6 +212,8 @@ class ModelsField(DictField):
             super().__set__(obj, None)
 
     def on_dependency_update(self, obj, updated_attribute, old_val, new_val):
+        if obj._suspend_dependency_updates:
+            return
         self.__set__(obj, self.default_factory(obj))
 
 class PRFField(TypedField):
@@ -226,10 +229,14 @@ class PRFField(TypedField):
         super().__init__(default_factory=make_prf, depends_on=depends_on)
 
     def on_dependency_update(self, obj, updated_attribute, old_val, new_val):
+        if obj._suspend_dependency_updates:
+            return
         self.__set__(obj, self.default_factory(obj))
 
 class SolverField(TypedField):
     def on_dependency_update(self, obj, updated_attribute, old_val, new_val):
+        if obj._suspend_dependency_updates:
+            return
         self.__set__(obj, self.default_factory(obj))
 
 class EquilibriumCalculationField(TypedField):
@@ -243,6 +250,8 @@ class EquilibriumCalculationField(TypedField):
         return getattr(obj, self.private_name)
 
     def on_dependency_update(self, obj, updated_attribute, old_val, new_val):
+        if obj._suspend_dependency_updates:
+            return
         self.__set__(obj, None)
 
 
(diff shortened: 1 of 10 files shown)


richardotis:
In a departure from best practice, I'm including an unrelated commit here (related to errors when `phase_record_factory.keys()` is not equal to `Workspace.phases`) that fixes a bug encountered downstream in ESPEI. I'm putting it here because all of these commits are needed for ESPEI's compatibility with Workspace, and it's easier for testing if it's all contained within one branch.

richardotis:
Windows test failures are due to https://github.com/matplotlib/matplotlib/issues/28551#issuecomment-2267563049

tacaswell:
The 3.9.1 wheels on windows would, depending on import order, cause segfaults, see https://github.com/matplotlib/matplotlib/issues/28551#issuecomment-2266699619.

We deleted the bad wheels to avoid segfaults (which seemed like a good idea) but failed to anticipate that this would cause downstream projects to start trying to build mpl from source instead of pulling the 3.9.0 wheel.  We have now yanked the whole 3.9.1 release to fix these problems and will get 3.9.2 out as quick as we can.

To prevent this sort of thing in the future (trying to build something from source when you are not prepared to do so) I suggest using `matplotlib --only-binary "matplotlib" ` in your requirements file (which may be a good practice for all wheels that have extensions).

Sorry for the trouble.

pycalphad/core/workspace.py
```diff
@@ -254,25 +270,40 @@ class Workspace:
     database: Database = TypedField(lambda _: None)
     components: SpeciesList = ComponentsField(depends_on=['database'])
     phases: PhaseList = PhasesField(depends_on=['database', 'components'])
-    conditions: Conditions = ConditionsField(lambda wks: Conditions(wks))
+    conditions: Conditions = ConditionsField(lambda wks: Conditions(wks), depends_on=['components', 'models'])
```
bocklund, line 273: where does the conditions dependence on models come from?

pycalphad/core/workspace.py
```diff
@@ -254,25 +270,40 @@ class Workspace:
...
     models: Mapping[PhaseName, ModelType] = ModelsField(depends_on=['phases', 'parameters'])
     parameters: SumType([NoneType, Dict]) = DictField(lambda _: OrderedDict())
     phase_record_factory: Optional[PhaseRecordFactory] = PRFField(depends_on=['phases', 'conditions', 'models', 'parameters'])
     calc_opts: SumType([NoneType, Dict]) = DictField(lambda _: OrderedDict())
     solver: SolverBase = SolverField(lambda obj: Solver(verbose=obj.verbose), depends_on=['verbose'])
     # eq is set by a callback in the EquilibriumCalculationField (TypedField)
-    eq: Optional[LightDataset] = EquilibriumCalculationField(depends_on=['phase_record_factory', 'calc_opts', 'solver'])
+    eq: Optional[LightDataset] = EquilibriumCalculationField(depends_on=['phase_record_factory', 'conditions', 'calc_opts', 'solver'])
 
     def __init__(self, *args, **kwargs):
```
bocklund, line 283: Maybe a little unrelated, but should we change the signature to 

```suggestion
    def __init__(self, database=None, components=None, phases=None, conditions=None, *, **kwargs):
```

right now we're throwing away any positional arguments past the first four. Per [PEP-3102](https://peps.python.org/pep-3102/), we can syntactically disallow any positional arguments beyond the first four

bocklund (approved):

pycalphad/core/workspace.py
richardotis: https://github.com/pycalphad/pycalphad/pull/554/files/bc95d629328445e8b611778bc2e4c0449651a05f#diff-c8bfc00897c8c4e08c283c9f9319933a84b8d6a356d1f98f3704aa40d05ec2d2R130

pycalphad/core/workspace.py
richardotis: This turns out to be a SyntaxError as `**kwargs` is not allowed after a bare `*`

Issue #540, opened 2024-07-11
Preserve line numbers in TDB parser tracebacks?


bocklund:
Not sure if it's possible, (it probably is at least annoying to implement) but it would be cool if we could preserve line numbers in the original file when we hit parse exceptions, e.g. the lsat bit of of the Exception produced by pyparsing:

`pyparsing.exceptions.ParseException: Expected TCCommand 'BMAGN', found '('  (at char 13), (line:1, col:14)`

richardotis:
We will need to associate line numbers with each line of the database file, and somehow propagate the information about lines that are removed (e.g., comments). There may also be some kind of approach involving calculating offsets of the removed lines.

PR #558, opened 2024-08-05
FIX: Better handle convergence failures in Workspace

bocklund:
This PR adds a testing `ConvergenceFailureSolver` to the test suite that overrides `solve` to return an unconverged result. We use this solver to add test cases for situations where there are convergence failures without having to track and continually update the test suite with real convergence failures as the solver improves to fix them.

In general, the expected behavior for convergence failures is that all properties derived from equilibrium return `NaN`, which is backwards compatible with pre-Workspace xarray Datasets. 

In particular, we test and fix bugs where the following `wks.get` calls would raise errors by trying to access composition sets that don't exist, taking into account the changes in #540 for `find_first_compset`:
- `wks.get("MU(component)")`
- `wks.get("Y(phase_name, sublattice_index, species)")`
- `wks.get(IsolatedPhase("phasename", wks=wks)("property"))`
- `wks.get(DormantPhase("phasename", wks=wks)("property"))`


bocklund:
This will need a rebase on develop once #554 is merged

bocklund:
@richardotis  One thing I'm not sure about is whether there are other properties or uses of Workspace worth testing for convergence failures. I think I got most the interesting variables and `GM` should be representative of model computed properties

pycalphad/tests/fixtures.py
```diff
@@ -26,3 +28,12 @@ def test_name_of_my_test(load_database):
...
     return pytest.mark.parametrize("load_database", [path], indirect=True)
+
+
+class ConvergenceFailureSolver(Solver):
+    """Solver that is guaranteed to produce a convergence failure for any composition sets or conditions passed"""
+    def solve(self, composition_sets, conditions):
+        spec = self.get_system_spec(composition_sets, conditions)
+        self._fix_state_variables_in_compsets(composition_sets, conditions)
+        state = spec.get_new_state(composition_sets)
+        return SolverResult(converged=False, x=composition_sets[0].dof, chemical_potentials=np.full_like(state.chemical_potentials, np.nan))
```
richardotis, line 39: So we are only setting NaN in chemical potentials, but otherwise returning the minimizer solution? This also looks like it would break calling code if more than one composition set is passed in as input.

pycalphad/tests/test_workspace.py
```diff
@@ -342,3 +342,27 @@ def test_issue_503_pure_vacancy_charge_balance():
...
+
+
+@select_database("cumg.tdb")
+def test_workspace_convergence_failures(load_database):
+    """Convergence failures should propagate NaNs through to computed properties"""
+    dbf = load_database()
+    components = ["CU", "MG", "VA"]
+    phases = ["LIQUID", "FCC_A1"]
+    wks = Workspace(dbf, components, phases, {v.N:1, v.P:1e5, v.T:300, v.X("MG"): 0.5}, solver=ConvergenceFailureSolver())
+    assert len(list(wks.enumerate_composition_sets())[0][1]) == 0  # convergence failures should have no stable composition sets
```
richardotis, line 354:
```suggestion
    assert len(wks.get_composition_sets()) == 0  # convergence failures should have no stable composition sets
```

pycalphad/variables.py
```diff
@@ -244,10 +244,12 @@ def __init__(self, phase_name, subl_index, species): #pylint: disable=W0221
             self._self_without_suffix = self
 
     def compute_property(self, compsets, cur_conds, chemical_potentials):
-        state_variables = compsets[0].phase_record.state_variables
         result = np.atleast_1d(np.full(self.shape, fill_value=np.nan))
+        state_variables = None
         for _, compset in self.filtered(compsets):
             # we'll only hit this code path if this phase is present in the list of compsets
+            if state_variables is None:
+                state_variables = compsets[0].phase_record.state_variables
```
richardotis, line 252:
```suggestion
                state_variables = compset.phase_record.state_variables
```

richardotis: `self.filtered(compsets)` may not include `compsets[0]` (though in principle the state variables should all agree)

pycalphad/variables.py
```diff
@@ -651,6 +655,8 @@ def compute_property(self, compsets, cur_conds, chemical_potentials):
 
     def jansson_derivative(self, compsets, cur_conds, chemical_potentials, deltas: JanssonDerivativeDeltas):
         "Compute Jansson derivative with self as numerator, with the given deltas"
+        if len(compsets) == 0:
+            return np.full(self.shape, np.nan)
```
richardotis, line 659: Coverage is pinging here - any easy way to hit this code path? I thought one of the tests above would do it, but it looks like it's showing missing.

richardotis (changes requested):

pycalphad/tests/fixtures.py
bocklund: I think it should't matter what the chemical potentials are set to. In real non-converging cases, the chemical potentials will not be NaN and the value of `x` should also not matter (unless there are cases where we actually expect to be able to produce any meaningful result from a convergence failure, but right now I don't think that we do).

pycalphad/variables.py
bocklund: do we expect that `compset[0]` could have more correct state variables? Isn't that already applying the assumption that the state variables are the same across all the compsets? Since this is in `SiteFraction`, it's possible that `compset[0]` isn't even the phase we're interested in.

pycalphad/variables.py
bocklund: we hit the equivalent in `JanssonDerivative.compute_property` and this method never gets called if there are no compsets. This is probably a useful guard in case anyone tries to do anything fancy later. Open to other options if we need to hit the line though.

pycalphad/tests/fixtures.py
bocklund: I set them to a nonsense value instead of NaN in the `SolverResult`, as that's more representative of realistic cases.

pycalphad/variables.py
richardotis: I don't think it should matter. I just wanted to raise it to be sure.

richardotis (approved):

PR #560, opened 2024-08-05
FIX: fix wildcard support

bocklund:
Fixes the following cases with a test:
- Fixes wildcard expansion on components (for properties without a sublattice index) to expand on non-vacant pure elements. With proper component support, we would go back to using `self.components`, but currently `self.components` contain all phase constituents, including, for example, vacancies
- Adds `expand_wildcard` on components for chemical potentials
- Fixes wildcard expansion for properties where expansions are on phase constituents, such as site fractions
- Adds support for wildcard expansion on sublattice indices (required adding `sublattice index = '*'` as valid).
- Change behavior of compute_property for site fractions to return NaN if the phase is not stable (previously we were returning zero)

### DIFF
--- a/pycalphad/core/workspace.py
+++ b/pycalphad/core/workspace.py
@@ -326,7 +326,8 @@ class Workspace:
                 additional_args = args[i].expand_wildcard(phase_names=phase_names)
                 args.extend(additional_args)
             elif hasattr(args[i], 'sublattice_index') and args[i].sublattice_index == '*':
-                # We need to resolve sublattice_index before species to ensure we get the correct set of components
+                # We need to resolve sublattice_index before species to ensure we
+                # get the correct set of phase constituents for each sublattice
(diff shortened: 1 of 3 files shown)


richardotis (approved):

pycalphad/core/workspace.py
```diff
@@ -325,18 +325,23 @@ def _expand_property_arguments(self, args: Sequence[ComputableProperty]):
                 phase_names = sorted(self.phase_record_factory.keys())
                 additional_args = args[i].expand_wildcard(phase_names=phase_names)
                 args.extend(additional_args)
+            elif hasattr(args[i], 'sublattice_index') and args[i].sublattice_index == '*':
+                # We need to resolve sublattice_index before species to ensure we get the correct set of components
```
richardotis, line 329: the correct set of species, right?

richardotis (approved):

Issue #503, opened 2023-12-13
Charged phases that cannot charge balance raise errors


bocklund:
### Discussed in https://github.com/pycalphad/pycalphad/discussions/502

<div type='discussions-op-text'>

<sup>Originally posted by **zliang1990** December 13, 2023</sup>
Hi Richard (@richardotis),

My name is Zhi Liang, Materials Design Engineer in QuesTek. When I recently attempted to run the attached V-O description (GitHub does not support .TDB so I zipped it) with Pycalphad, I noticed that the following error message will pop up if I only choose oxygen (and vacancy) as elements.

Upon debugging, it seems to do with the HALITE phase, whose sublattice model is V,V+2,V+3,VA : O-2,VA. When only choosing oxygen, first sublattice only allows VA while second sublattice allows O-2 and VA, which seemingly triggers the following error message. This issue can be a blocker if in a multicomponent system, certain phase only has sublattice model A,VA : O-2,VA, but the element selection only has element B and oxygen, and this error message will pop up as well (Sorry that I cannot provide our multicomponent database since it is proprietary).

Could you please help to see how we can resolve this?

Thanks!

Best,
Zhi

![image](https://github.com/pycalphad/pycalphad/assets/84593259/fce335ce-9037-4f6c-8a72-e112d55c2a3e)
[15YANG_V-O description.zip](https://github.com/pycalphad/pycalphad/files/13666266/15YANG_V-O.description.zip)</div>

bocklund:
From #502:

What's happening here is that when only vacancies and oxygen are entered as elements PyCalphad is treating CORUNDUM and HALITE as active phases (because there are active species in each sublattice).

For both phases, it's impossible to charge balance given these components (pure vacancy endmembers don't count, such as in the case for HALITE) and that gives array size errors because there are no feasible points. I'm creating an issue for this, since it's a bug. We should probably prevent phases that cannot charge balance from being included in the active set of phases.

richardotis:
This should be easy to fix with the new sampler, as it only samples the feasible subspace of points. For these cases it would find no solution, so it would be easy to raise an error (or return zero points) so the phase is automatically excluded from the active set.

richardotis:
This is now fixed in #432

nury12n:
I noticed the test (pycalphad/tests/test_work_space.py::test_issue_503_pure_vacancy_charge_balance) for this issue can occasionally fail.

These were some of the outputs I got when the test failed:
[...]

PR #561, opened 2024-08-08
TST: Revise ambiguous test description and conditions

richardotis:
- Fixes the test names and descriptions of two tests related to #503 
- For [`test_issue_503_suspend_phase_infeasible_internal_constraints`](https://github.com/pycalphad/pycalphad/pull/561/files#diff-31e23fd0f866a3d4f331e4bd2bfa4840923e1f047fb204497622e549367b0cbcR328-R329): The behavior we are testing for does not require that we set a mole fraction to 1, so we change that as well. There's still an issue with the minimizer non-deterministically failing near the edge of composition space, but this test does not capture that behavior consistently enough to be useful, so we side-step the issue for now.

For that issue we ended up adding two sets of fixes because there were really two issues:
1. In [`test_issue_503_charged_infeasible_subsystem`](https://github.com/pycalphad/pycalphad/pull/561/files#diff-fb922ec5bc8777718fa5d38feb4e4392df16cad9db8084449ede5308916e4b69L1018-L1019), we needed to suspend HALITE because there was no vanadium entered into the list of components to allow for charge-balancing the oxygen, but the phase still had a pure-vacancy endmember that was feasible in the sampler (but un-physical).
2. In [`test_issue_503_pure_vacancy_charge_balance`](https://github.com/pycalphad/pycalphad/pull/561/files#diff-31e23fd0f866a3d4f331e4bd2bfa4840923e1f047fb204497622e549367b0cbcL328-L329), we had ZR, O, and VA all entered, but the SPINEL phase is defined as stoichiometric with non-zero charge, so it can never be entered and will always have zero feasible points from the sampler. Contrary to the test name and description, the pure-vacancy sublattices in the SPINEL phase aren't playing a role here.

### DIFF
--- a/pycalphad/tests/test_equilibrium.py
+++ b/pycalphad/tests/test_equilibrium.py
@@ -1015,7 +1015,7 @@ def test_eq_charge_ndzro(load_database):
                                 6.61821340e-03, 1.00000000e+00, 1.39970285e-03, 9.98600297e-01], rtol=5e-4)
 
 @pytest.mark.solver
-def test_issue_503_charged_infeasible_subsystem():
+def test_issue_503_suspend_pure_vacancy_configuration():
     "equilibrium suspends a phase with a pure-vacancy endmember as the only feasible configuration (gh-503)"
     tdb = """
  ELEMENT /-   ELECTRON_GAS              0.0000E+00  0.0000E+00  0.0000E+00!
--- a/pycalphad/tests/test_equilibrium.py
+++ b/pycalphad/tests/test_equilibrium.py
@@ -1016,7 +1016,7 @@ def test_eq_charge_ndzro(load_database):
 
 @pytest.mark.solver
 def test_issue_503_charged_infeasible_subsystem():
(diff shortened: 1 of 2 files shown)


richardotis:
It may make sense to add `issue_503` back to the test name. I'm not all the way sure. We actually already have another test that purports to fix #503, but we ended up adding two sets of fixes because there were really two issues:
1. In `test_issue_503_charged_infeasible_subsystem`, we needed to suspend HALITE because there was no vanadium entered into the list of components to allow for charge-balancing the oxygen, but the phase still had a pure-vacancy endmember that was feasible in the sampler (but un-physical).
2. In `test_issue_503_pure_vacancy_charge_balance`, we had ZR, O, and VA all entered, but the SPINEL phase is defined as stoichiometric with non-zero charge, so it can never be entered and will always have zero feasible points from the sampler. Contrary to the test name and description, the pure-vacancy sublattices in the SPINEL phase aren't playing a role here.

I'm wondering now if the names and descriptions of the two tests got mixed up somehow.

bocklund (approved):

PR #581, opened 2024-11-13
High order reciprocal model

nury12n:
This pull request implements the composition dependent reciprocal parameter as defined in eq 5.99 in:
H.L. Lukas, S.G. Fries, B. Sundman, "Computational Thermodynamics - The Calphad Method, Ch 5. Models for the Gibbs Free Energy, 5.8 Modeling using sublattices, 5.8.1. Reciprocal solutions, 5.8.1.1. Excess Gibbs energy for the reciprocal solution" Cambridge University Press 2007. ISBN: 978-0-521-86811-2

For excess mixing in a reciprocal solution:
```math
^E G_m = y_A^0 y_B^0 y_C^1 L_{AB:C} + y_A^0 y_B^0 y_D^1 L_{AB:D} + y_A^0 y_C^1 y_D^1 L_{A:CD} + y_B^0 y_C^1 y_D^1 L_{B:CD} + y_A^0 y_B^0 y_C^1 y_D^1 L_{AB:CD}
```

The ternary terms are defined by with a Redlich-Kister polynomial, but the reciprocal parameter can be defined as the following the include composition dependency:
```math
L_{AB:CD} = ^0L_{AB:CD} + (y_A^0 - y_B^0) ^1L_{AB:CD} + (y_C^1 - y_D^1) ^2L_{AB:CD}
```
This is an example assessment that used these higher order parameters: C. Zhang, Y. Peng, P. Zhou, W. Zhang, Y. Du, Calphad 51 (2015) 104-110, [DOI: 10.1016/j.calphad.2015.09.001](https://doi.org/10.1016/j.calphad.2015.09.001). 

Unit tests are added for the following behaviors:
- Accuracy of mixing energy with respect to Thermo-Calc behavior
- Model will raise error if reciprocal parameter with an order > 2 is used (e.g. `PARAM G(PHASE,A,B:C,D;3)`)
- Model will raise error if reciprocal parameter with order > 0 is used for mixing on 3 or more sublattices (e.g. `PARAM G(PHASE,A,B:C,D:E,F;1)`)

### DIFF
--- a/pycalphad/model.py
+++ b/pycalphad/model.py
@@ -768,26 +768,18 @@ class Model(object):
                     mixing_term *= Mul(*comp_symbols)
                 # is this a higher-order interaction parameter?
                 if len(comps) == 2 and param['parameter_order'] > 0:
-                    subl_indices = []
-                    comps_list = []
-                    for subl_index_2, comps_2 in enumerate(param['constituent_array']):
-                        if len(comps_2) == 2:
-                            subl_indices.append(subl_index_2)
-                            comps_list.append(comps_2)
-                    if len(comps_list) > 1:
+                    subl_indices, comps_list = list(zip(*[[idx,c] for idx,c in enumerate(param['constituent_array']) if len(c) == 2]))
+                    if len(comps_list) > 2:
+                        raise ValueError("Composition dependent reciprocal parameter is not supported for more than 2 sublattices")
+                    elif len(comps_list) > 1:
                         # only add (yA - yB) if the sublattice corresponds to the parameter order
                         # this is based off the equation from HL Lukas, SG Fries, B Sundman
                         #   Computational Thermodynamics - The Calphad Method, 
                         #   Ch5 Models for the Gibbs Free Energy, Section 5.8.1. Reciprocal solutions
                         #   Equation 5.99
-                        # TODO: check if this generalizes for mixing on more than two sublattices
-                        #       Ex. L_A,B:C,D:E,F
-                        #           1st order -> (yA-yB)
-                        #           2nd order -> (yC-yD)
-                        #           3rd order -> (yE-yF)
-                        #       This may not be necessary since its highly unlikely a parameter
-                        #       of this order would be used, and it should be discouraged anyways
-                        if subl_indices.index(subl_index) == param['parameter_order'] - 1:
+                        # Testing with thermo-calc seems to be that L1 corresponds to the second
+                        # sublattice and L2 corresponds to the first sublattice
+                        if (1-subl_indices.index(subl_index)) == param['parameter_order'] - 1:
                             mixing_term *= (comp_symbols[0] - comp_symbols[1])
                     else:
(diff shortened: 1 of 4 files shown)


bocklund:
Rebasing with the new changes in develop should fix the tests

pycalphad/model.py
```diff
@@ -768,9 +768,25 @@ def redlich_kister_sum(self, phase, param_search, param_query):
...
+                        # only add (yA - yB) if the sublattice corresponds to the parameter order
+                        # this is based off the equation from HL Lukas, SG Fries, B Sundman
+                        #   Computational Thermodynamics - The Calphad Method, 
+                        #   Ch5 Models for the Gibbs Free Energy, Section 5.8.1. Reciprocal solutions
+                        #   Equation 5.99
+                        # Testing with thermo-calc seems to be that L1 corresponds to the second
+                        # sublattice and L2 corresponds to the first sublattice
+                        if (1-subl_indices.index(subl_index)) == param['parameter_order'] - 1:
+                            mixing_term *= (comp_symbols[0] - comp_symbols[1])
+                    else:
```
bocklund, line 786: My reading of 5.8.1.1 is that this would also apply to the case of more than 3 sublattices, can we test that in TC?

nury12n: My interpretation from the last two paragraphs would be for a model with 3+ sublattices, the reciprocal parameter is only defined for two composition dependent terms. So:
- $L_{AB:C:EF} = ^0 L_{AB:C:EF} + (y^0_A - y^0_B)* ^1 L_{AB:C:EF} + (y^2_E - y^2_F)* ^2 L_{AB:C:EF}$ 
- $L_{AB:CD:EF} = ^0 L_{AB:CD:EF}$ - composition dependent reciprocal parameter not supported since there's 3 composition dependent terms

Is this how you interpret it as well?

Now, I suppose equation 5.99 can be expanded to more sublattices as:
$$L_{AB:CD:EF} = ^0 L_{AB:CD:EF} + (y^0_A - y^0_B)* ^1 L_{AB:CD:EF} + (y^1_C - y^1_D)* ^2 L_{AB:CD:EF} + (y^2_E - y^2_F)* ^3 L_{AB:CD:EF}$$

But I think I remember this failing the parser in TC, and the last two paragraphs seems to favor not supporting this expansion to discourage models with a too high number of parameters.

nury12n: Behavior in TC for reciprocal parameters on more than 2 sublattices seems to be:
- $^0 L_{AB:CD:EF}$ - is stated to not be supported according to the database parser, but equilibrium will still give the expected value of $G_{excess} = y^0_A y^0_B y^1_C y^1_D y^2_E y^2_F * ^0 L_{AB:CD:EF}$
- $^n L_{AB:CD:EF}$ for $n>0$ - is not supported and equilibrium will fail with message: Illegal composition dependency
- $^n L_{AB:C:EF}$ - is supported

PR #595, opened 2025-03-04
IsolatedPhase fix and composition Jansson derivative default change

cjkunselman18:
IsolatedPhase was not constraining chemical potentials to the compset, but was allowing phase separation in miscibility gaps.

Currently, the default for composition Jansson derivatives is to define n-1 dependent components rather than just 1. Defining one dependent component seems more intuitive and applicable.

### DIFF
--- a/pycalphad/core/minimizer.pyx
+++ b/pycalphad/core/minimizer.pyx
@@ -747,12 +747,12 @@ cpdef fixed_component_differential(SystemSpecification spec, SystemState state, 
     equilibrium_soln[:] = 0
 
     # delta mole fractions must sum to zero; we have degrees of freedom to decide how to distribute
-    # for now, redistribute evenly over all other fixed components
+    # for now, specify one dependent component
     for i in range(spec.prescribed_mole_fraction_coefficients.shape[0]):
         if np.all(np.asarray(spec.prescribed_mole_fraction_coefficients[i]) == comparison_array):
             equilibrium_soln[num_stable_phases + num_fixed_phases + i] = 1
         else:
-            equilibrium_soln[num_stable_phases + num_fixed_phases + i] = -1/(num_fixed_mole_fraction_conditions)
+            equilibrium_soln[num_stable_phases + num_fixed_phases + i] = 0
     lstsq_check_infeasible(equilibrium_matrix, equilibrium_soln, equilibrium_soln)
     for i in range(spec.free_chemical_potential_indices.shape[0]):
         chempot_idx = spec.free_chemical_potential_indices[i]
--- a/pycalphad/property_framework/metaproperties.py
+++ b/pycalphad/property_framework/metaproperties.py
@@ -214,8 +214,9 @@ class IsolatedPhase:
                             chemical_potentials: npt.ArrayLike) -> float:
                 if self._compset is None:
                     return prop.compute_property([], cur_conds, chemical_potentials)
-                self.solver.solve([self._compset], cur_conds)
-                return prop.compute_property([self._compset], cur_conds, chemical_potentials)
+                res = self.solver.solve([self._compset], cur_conds)
+                new_chemical_potentials = res.chemical_potentials
+                return prop.compute_property([self._compset], cur_conds, new_chemical_potentials)
 
             @staticmethod
             def jansson_derivative(compsets, cur_conds, chemical_potentials, deltas):


bocklund:
I'm not sure what's going on with the Windows tests, but that code seems to be orthogonal to the code here and also occurring in #594

bocklund:
I came up with test for this:

```python
from importlib.resources import files
import pycalphad.tests.databases
from pycalphad import Database, Workspace, calculate, variables as v
from pycalphad.property_framework import IsolatedPhase
import numpy as np

temperature = 100 # K
dbf = Database(files(pycalphad.tests.databases).joinpath("pbsn.tdb"))
wks = Workspace(dbf, phases=["BCT_A5"], components=["PB", "SN", "VA"], conditions={v.T: temperature, v.P:101325, v.X("SN"): 0.5})
calc_res = calculate(dbf, ["PB", "SN", "VA"], ["BCT_A5"], T=temperature, pdens=1000)

global_GM = wks.get("GM")
global_chempots = np.asarray(wks.get("MU(*)"))
global_sitefracs = wks.get("Y(BCT_A5,0,*)")
isolated_GM = wks.get(IsolatedPhase("BCT_A5",wks=wks)("GM"))
isolated_chempots = np.asarray(wks.get(IsolatedPhase("BCT_A5",wks=wks)("MU(*)")))
isolated_chempots_manually_extracted = np.asarray([wks.get(IsolatedPhase("BCT_A5",wks=wks)(x)) for x in ["MU(PB)", "MU(SN)"]])
# isolated_sitefracs = wks.get(IsolatedPhase("BCT_A5",wks=wks)("Y(BCT_A5,0,*)"))  # currently fails
print("Global")
print(global_GM)
print(global_chempots)
print(global_sitefracs)
print("Isolated")
print(isolated_GM)
print(isolated_chempots)  # TODO: this is a bug, we should bet an array out
print(isolated_chempots_manually_extracted)  # TODO: this is also a bug, it"s the same as the global min hyperplane these look like chempots in the MG
# print(isolated_sitefracs)  # currently fails

plt.scatter(calc_res.X.sel(component="SN").squeeze(), calc_res.GM.squeeze(), s=2, label="GM", c='blue')
plt.scatter(wks.get(IsolatedPhase("BCT_A5",wks=wks)("X(SN)")), wks.get(IsolatedPhase("BCT_A5",wks=wks)("GM")), c='k', label="Isolated phase GM")
plt.scatter(wks.get("X(SN)"), wks.get("GM"), c='red', label="Global GM")
plt.plot([0, 1], np.asarray([wks.get(IsolatedPhase("BCT_A5",wks=wks)(x)) for x in ["MU(PB)", "MU(SN)"]]), c='k', ls='-', label="Isolated phase MU")
plt.plot([0, 1], np.asarray(wks.get("MU(*)")), c='red', label="MU", ls='--')
plt.xlabel("X(SN)")
plt.ylabel("GM")
plt.title(f"Pb-Sn T = {temperature} K")
plt.legend(loc="center left", bbox_to_anchor=(1.01, 0.5))
plt.savefig(f"Pb-Sn-GM@{temperature}K.png", dpi=150, bbox_inches="tight")
```

The test actually demonstrates two apparent bugs:
1. Getting MU() for IsolatedPhase gives the global min MU.
2. We either get errors or meaningless results for using wildcards in isolated phase: `wks.get(IsolatedPhase()("MU(*)"))` returns `0.0` and `wks.get(IsolatedPhase("BCT_A5",wks=wks)("Y(BCT_A5,0,*)"))` raises `ValueError: Y(BCT_A5,0,*) is not in list`

On current develop you get this:
Pb-Sn-GM@100K-broken

After apply changes so far up to 0911aef you get this (which is correct):
Pb-Sn-GM@100K

I committed a test to this branch for the chemical potentials fix. The second bug is still unaddressed and we also still need a test for the change to `fixed_component_differential`.

bocklund:
I've played a little bit with the isolated phase computed properties and managed to get this test to pass:

```python
@select_database("pbsn.tdb")
def test_wildcards_for_isolated_phases(load_database):
    """Using wildcards in wks.get for IsolatedPhase give the correct values (and not raise exceptions)"""
    dbf = load_database()
    temperature = 100 # K
    wks = Workspace(dbf, phases=["BCT_A5"], components=["PB", "SN", "VA"], conditions={v.T: temperature, v.P:101325, v.X("SN"): 0.6})

    out = IsolatedPhase("BCT_A5",wks=wks)("X(*)")
    print("out", out)
    isolated_X = np.asarray(wks.get(*IsolatedPhase("BCT_A5",wks=wks)("X(*)")))
    isolated_Y = np.asarray(wks.get(*IsolatedPhase("BCT_A5",wks=wks)("Y(*,*,*)")))
    isolated_chempots = np.asarray(wks.get(*IsolatedPhase("BCT_A5",wks=wks)("MU(*)")))
    assert_allclose(isolated_X, np.asarray([0.4, 0.6]))
    assert_allclose(isolated_Y, np.asarray([0.4, 0.6, 1.0]))
    assert_allclose(isolated_chempots, np.asarray([-2992.70848448, -5280.02439881]))
```

but I'm not very happy with the internal hacks I had to do to make this work, so I think we should treat that as out of scope for the purpose of this PR. I'll open another issue for that.

@cjkunselman18 can you give an example where the Jansson derivative fix is comes into play? I can help write the test if needed and then this is ready for review and merge.

cjkunselman18:
I would love help writing a test. For any system with three or more components, the fix ensures that Jansson derivatives with respect to an independent composition variable only define one dependent component. A simple way to test for this is ensuring that dG/dX(A) = mu(A) - mu(B) where B is the dependent component. Prior to the fix, for system A-B-C, dG/dX(A) = mu(A) - mu(B)/2 - mu(C)/2.

pycalphad/core/minimizer.pyx
```diff
@@ -747,12 +747,12 @@ cpdef fixed_component_differential(SystemSpecification spec, SystemState state,
...
 
     # delta mole fractions must sum to zero; we have degrees of freedom to decide how to distribute
-    # for now, redistribute evenly over all other fixed components
+    # for now, specify one dependent component
     for i in range(spec.prescribed_mole_fraction_coefficients.shape[0]):
         if np.all(np.asarray(spec.prescribed_mole_fraction_coefficients[i]) == comparison_array):
             equilibrium_soln[num_stable_phases + num_fixed_phases + i] = 1
         else:
-            equilibrium_soln[num_stable_phases + num_fixed_phases + i] = -1/(num_fixed_mole_fraction_conditions)
+            equilibrium_soln[num_stable_phases + num_fixed_phases + i] = 0
```
richardotis, line 755: This doesn't look quite right to me. Delta mole fractions must sum to zero. The revised sum here is 1. Where is the code that specifies a dependent component that has its RHS set to -1?

cjkunselman18: There are no explicit deltas for the dependent component, from my understanding. When you set up the Jansson derivative, you specify the deltas for all of the independent components, and the dependent component takes on the value necessary to make the sum 0.

bocklund: i checked that values in the added test are in agreement with TC

richardotis: @bocklund Was the example you checked with more components than 2?

bocklund: The Cr-Fe-V ternary case I added to the tests are in agreement with the TC values

PR #602, opened 2025-04-11
MAINT: Bump minimum Python version to 3.13, drop 3.9; test and dependency fixes

bocklund:
Dropping 3.9 in accordance with [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html)

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -66,7 +66,7 @@ jobs:
       - uses: actions/setup-python@v5
         name: Install Python
         with:
-          python-version: '3.9'
+          python-version: '3.13'
 
       - run: pip install build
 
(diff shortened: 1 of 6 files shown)


bocklund:
3.13 tests appear to show a regression

bocklund:
I don't know what's changed that this is manifesting only in our 3.13 tests, but the problem is

```python
sorted(np.asarray([0.000151, np.nan, 0.9998,])) != np.sort(np.asarray([0.000151, np.nan, 0.9998,]))
```
specifically
```python
sorted(array([9.99848292e-01, 1.51707990e-04, nan]))
```
becomes
```python
[np.float64(0.00015170798979249788), np.float64(nan), np.float64(0.9998482920120817)]
```

The array `eq.NP` seems to be ordered as expected with NaN last. Changing the test from `sorted` to `np.sort` fixes it.

bocklund:
Added `symengine<0.14` to runtime requirements as a mitigation for #602.

richardotis (approved):

PR #603, opened 2025-04-12
ENH: Add mass to variables

bocklund:
Add `Mass` (with alias as `B`) to variables, which enables `wks.get("B(*)")` or `wks.get("Mass(*)")`. With this feature, you can get:
- System mass, `v.B()`
- Mass of a component, `v.B("AL")`
- Mass of a component in a phase, `v.B("LIQUID", "AL")`
- Mass of a phase (though explicit construction of `v.B()` and setting the `phase_name` variable)
- Concrete components, when we go to do a calculation, must be defined in terms of pure elements, so allow parsing help by matching against pure elements with optional numbers so things like "KF" work with "K" and "F" elements without "K1F1"
- "H2O" == "OH2" and don't rename from user defined 
- mass conversion is matrix multiply; emit warning if one mass is zero and mass conditions are used. emit error if all element masses are zero and mass conditions are used. 

There's also no functionality here that would allow normalizing by mass (or normalizing the mass itself, e.g. `BM` to compute molar mass). Maybe that would come when we generalize `N` conditions beyond system `N`.


richardotis (commented):
Looking good! Just a few general comments:
1. Can mass be used as a condition here? Because of our linear combination of mole fraction support, I think this would be relatively straightforward.
2. Can you calculate Jansson derivatives involving the mass?

bocklund:
1. I haven't wired that up. There would be a gotcha that system mass likely wouldn't work because we currently require system moles = 1. My experience with the linear combination conditions (weight fractions) is that they are less stable than mole fractions. My preference would be to defer until we can do a linear combination of moles (N) rather than write it in mole fractions (my hypothesis is that that would be more stable)
2. Jansson derivatives don't yet work. I could take a look at getting support as a numerator. Getting support as a denominator is dependent on (1), of course.

richardotis:
I suggest getting the numerator derivative working, add a test, and then merge.

Issue #626, opened 2025-09-07
Grid/starting points bug for infeasible phases


bocklund:
Other flaky tests ([log here](https://github.com/bocklund/pycalphad/actions/runs/17196748193/job/48780969976)) appear to produce actual bugged behavior (`NP` not unity). I tried to run this test 100 times in a loop from the command line and it never failed on a fresh macOS environment with Python 3.13.3

```
_________ test_issue_503_suspend_phase_infeasible_internal_constraints _________

    def test_issue_503_suspend_phase_infeasible_internal_constraints():
        "Phases that cannot satisfy internal constraints are correctly suspended (gh-503)"
        TDB = """
        ELEMENT /-   ELECTRON_GAS              0.0000E+00  0.0000E+00  0.0000E+00!
        ELEMENT VA   VACUUM                    0.0000E+00  0.0000E+00  0.0000E+00!
        ELEMENT O    1/2_MOLE_O2(G)            1.5999E+01  4.3410E+03  1.0252E+02!
        ELEMENT ZR   BLANK                     0.0000E+00  0.0000E+00  0.0000E+00!
        SPECIES O-2                         O1/-2!
        SPECIES ZR+4                        ZR1/+4!
        PHASE SPINEL %  4 1 2 2 4 !
        CONSTITUENT SPINEL : ZR+4 : VA : VA : O-2 :  !
        PHASE GAS:G %  1  1.0  !
        CONSTITUENT GAS:G :O,ZR :  !
        """
        # SPINEL phase cannot charge balance, so even though it contains ZR, O, and VA, it must be suspended
        wks = Workspace(TDB, ['O', 'ZR', 'VA'], ['SPINEL', 'GAS'], {v.P: 1e5, v.X('O'): 0.5, v.T: 1000})
        print(wks.get_dict('X(*,*)'))
        print(wks.get_dict('NP(*)'))
        assert np.isnan(wks.get('NP(SPINEL)'))
>       np.testing.assert_almost_equal(wks.get('NP(GAS)'), 1.0)
E       AssertionError: 
E       Arrays are not almost equal to 7 decimals
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference among violations: 0.5
E       Max relative difference among violations: 0.5
E        ACTUAL: array(0.5)
E        DESIRED: array(1.)

[...]

PR #627, opened 2025-09-08
FIX: Fix `_compute_phase_values` adding a point with garbage data when the size of feasible `points` is zero.

bocklund:
This PR fixes the bug described in #626 where `_compute_phase_values` produces a point with bad data when there are no `points` from `_sample_phase_constitution` (i.e. `points.size == 0`). 

The cause of the bug is that `np.atleast_2d(points)` gives `points.shape == (1, 0)` in that case, which messes up the logic in the code that assumes `points.shape[0]` refers to the number of points, which it usually does except for when the `points` array is size zero and the sizes of the dimensions are meaningless. This causes calls to `phase_record.prop_2d`/`prop_parameters_2d`  and `phase_record.mass_obj_2d` to produce garbage data because the `dof` provided is malformed.

The fix implemented here is to only call into the phase record functions if there are a non-zero number of points and otherwise make empty, zero-sized `phase_output` and `phase_compositions` arrays that have the correct shapes (and to correct the shape of `points` as it is used later).

If a call to `calculate` produces no feasible points, each phase that produces no points will raise a warning and if there are no feasible points among all phaes a `ConditionError` is raised that prompts the user to check the active phases and set of conditions (and state variables).

### DIFF
--- a/pycalphad/core/calculate.py
+++ b/pycalphad/core/calculate.py
@@ -258,6 +258,7 @@ def _compute_phase_values(components, statevar_dict, str_phase_local_conditions,
     bc_statevars = np.ascontiguousarray([broadcast_to(x, points.shape[:-1]).reshape(-1) for x in statevars])
     param_symbols, parameter_array = extract_parameters(parameters)
     parameter_array_length = parameter_array.shape[0]
+    max_tieline_vertices = len(pure_elements)
     if points.size > 0:
         pts = points.reshape(-1, points.shape[-1])
         dof = np.ascontiguousarray(np.concatenate((bc_statevars.T, pts), axis=1))
@@ -275,7 +276,6 @@ def _compute_phase_values(components, statevar_dict, str_phase_local_conditions,
         for el_idx in range(len(pure_elements)):
             phase_record.mass_obj_2d(phase_compositions[:, el_idx], dof, el_idx)
 
-        max_tieline_vertices = len(pure_elements)
         if isinstance(phase_output, (float, int)):
             phase_output = broadcast_to(phase_output, points.shape[:-1])
         if isinstance(phase_compositions, (float, int)):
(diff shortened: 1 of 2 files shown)


bocklund:
I think raising if there are no points at all for any phase can make sense, as that means the set of conditions are totally infeasible. We could raise a `ConditionError` in `equilibrium`/`Workspace`. A warning might make less sense because it might be expected behavior and be needlessly noisy. I could be convinced otherwise.

pycalphad/core/calculate.py
```diff
@@ -256,39 +256,49 @@ def _compute_phase_values(components, statevar_dict, str_phase_local_conditions,
...
+        # The size of the points dimension will be hardcoded to zero (np.atleast_2d will have the new dimension be size 1 even if the input array is size zero).
+        points = points.reshape(points.shape[:-2] + (0, points.shape[-1],))  # np.atleast_2d always makes a (1, N) even if size is zero, here we just make it consistent
+        if parameter_array_length <= 1:
+            phase_output = np.empty(points.shape[:-1], dtype=np.float64)
+        else:
+            phase_output = np.empty(points.shape[:-1] + (parameter_array_length,), dtype=np.float64)
+        phase_compositions = np.empty(points.shape[:-1] + (len(pure_elements),), dtype=np.float64)
+
     if fake_points:
         output_shape = points.shape[:-2] + (max_tieline_vertices,)
```
nury12n, line 303: This branch here can be accessed before `max_tieline_vertices` is created. `max_tieline_vertices` is only created if the phase is active and has feasible points. However, for a phase that is active but has no feasible points and `fake_points` is `True` in `calculate`, then `max_tieline_vertices` will not be defined before going to this branch. I think this can be easily fixed if line 278 is moved before the `if points.size > 0:` branch

nury12n: This can be tested in `test_calculate.py::test_calculate_raises_if_no_feasible_points_exists` if `fake_points` is set to True, where it will raise an `UnboundLocalError`

PR #633, opened 2025-10-24
MAINT: Transition setup.py to pyproject.toml and use uv in CI

bocklund:
- transition`setup.py` to `pyproject.toml`. we need to retain `setup.py` for Python extensions, but can move all the metadata into `pyproject.toml`
  - this allows us to unify all the requirements into `pyproject.toml`. We therefore delete `requirements-dev.txt`
  - updated the PR template to simplify that requirements are only managed in `pyproject.toml`
- switch to `uv` in CI. `uv` simplifies the CI scripts and seems to cut off about 1 minute off CI on average
- manually checked that the `pip`-based workflow for installing still works (without `requirements-dev.txt`) and that we can turn off build isolation. I update the docs to reflect that


richardotis (approved):
Please fix the doc build and then it is good to merge after all CI passes.

PR #634, opened 2025-10-24
MAINT: Database object typing

bocklund:
Add typing information for Database objects.

We have to switch around how we do the documentation here to make all the tooling happy.
- language servers weren't detecting the types or the documentation properly. It seems like they expect attributes on `__init__`, but we use `__new__`.
- For sphinx, autodoc was detecting multiple definitions of our attributes, e.g. "elements". This fixes it and also improves how objects are seen by language servers
- for sphinx, there are apparently two sets of annotation documentation types allowed, either comment above `#: <doc>` or docstrings below `"""<doc>"""`. I think the comments above look much better, but the language server implementations (e.g. pylance) seems to like the docstring version better.

### DIFF
--- a/pycalphad/io/database.py
+++ b/pycalphad/io/database.py
@@ -63,21 +63,6 @@ class Database(object): #pylint: disable=R0902
     """
     Structured thermodynamic and/or kinetic data.
 
-    Attributes
-    ----------
-    elements : set[str]
-        Set of elements in database.
-    species : set[Species]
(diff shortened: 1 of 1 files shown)


nury12n (approved):

nury12n:
This seems to also help pylance recognize what types the `Database` attributes are, which is nice when I'm manually changing things in a `Database` instance.

Issue #638, opened 2025-11-03
Solid-liquid two-phase regions not plotted correctly by ternplot in V0.11.1


GeorgeZJT:
Hello pyhelphad dev team, this is a follow up on issue #624 

In V0.11.1, `ternplot` and `plot_ternary` don't plot the solid-liquid two-phase regions correctly. This happens both when using `ternplot `directly, and when using`TernaryStrategy`, `add_nodes_from_conditions()`, `do_map()`, and `plot_ternary()`, as in [the new documentation](https://pycalphad.org/docs/latest/examples/TernaryExamples.html)

Sometimes, adding a special starting point may generate the correct image, but the method is not reliable across all temperatures and pressures.

As such, I'm unable to recreate the resulting figure in the example.

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/6915278c-d782-4b9c-a103-5c0f837b714f" />

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/47ed9c91-2121-4d2e-843b-f93736af4613" />

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/84b08a7f-9a47-4aa1-94e0-3ad75326f45d" />

Environment:
pycalphad 0.11.0 and 0.11.1
windows 11
python 3.13.9

Code:
```
import matplotlib.pyplot as plt
from pycalphad import Database, ternplot
from pycalphad import variables as v

temperature = 1300

db_al_cu_y = Database('Al_Cu_Y_assessment_files.tdb')
comps = ['AL', 'CU', 'Y', 'VA']
phases = list(db_al_cu_y.phases.keys())
conds = {v.T: temperature, v.P:101325, v.X('AL'): (0,1,0.015), v.X('Y'): (0,1,0.015)}
ternplot(db_al_cu_y, comps, phases, conds, x=v.X('AL'), y=v.X('Y'), label_nodes = True)
plt.title(f'Al-Cu-Y at T={temperature}K on V0.11.1')
plt.savefig(f'T={temperature} V.11.0.png')
plt.show()
```

```
import matplotlib.pyplot as plt
from pycalphad import Database
from pycalphad import variables as v
from pycalphad.mapping import TernaryStrategy, plot_ternary

temperature = 1770

db_al_cu_y = Database('Al_Cu_Y_assessment_files.tdb')
comps = ['AL', 'CU', 'Y', 'VA']
phases = list(db_al_cu_y.phases.keys())
[...]

PR #639, opened 2025-11-11
FIX: global min check on invariant calculations in binary and ternary mapping

nury12n:
This addresses part of #638 to fix tie-lines not being computed out of a three-phase tie-triangle. This fix checks for degenerate composition sets when performing a global min check on computed invariants. Without this check, it was possible that a tie-triangle at global min was falsely rejected as a metastable equilibria.

This also seems to fix some of the troublesome binaries that were brought up in #564:
- [AuSn-13Don](https://phasediagrams.org/phase-diagram/AuSn-13Don.tdb)
- [BiLi-13Liu](https://phasediagrams.org/phase-diagram/BiLi-13Liu.tdb)
- [PdTi-11Guo](https://phasediagrams.org/phase-diagram/PdTi-11Guo.tdb)


bocklund:
Looks great, the 6 line change is punching above its weight class here in local testing. Is there a minimal test that can be written that gets at the specific tie-triangle behavior this fixes?

bocklund (approved):

PR #646, opened 2026-01-06
MAINT: Upgrade to pyparsing v3 pep-8 API

bocklund:
Upgraded via pyparsing tooling as suggested by the version 3.3 release logs. Release should be backwards compatible to pyparsing v3.0

Checklist

### DIFF
--- a/pycalphad/io/grammar.py
+++ b/pycalphad/io/grammar.py
@@ -4,11 +4,11 @@ from pyparsing import alphas, nums
 from pyparsing import Group, OneOrMore, Optional, Regex, Suppress, Word
 import re
 
-pos_neg_int_number = Word('+-' + nums).setParseAction(lambda t: [int(t[0])])  # '+3' or '-2' are examples
+pos_neg_int_number = Word('+-' + nums).set_parse_action(lambda t: [int(t[0])])  # '+3' or '-2' are examples
 # matching float w/ regex is ugly but is recommended by pyparsing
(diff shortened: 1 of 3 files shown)


bocklund:
Just docs builds are failing here. There was a new major version of sphinx released recently which seems to have potentially broken builds for other folks according to their issue tracker. Part of the traceback is as follows:

```python
      File "/Users/bocklund1/src/pycalphad/.venv/lib/python3.13/site-packages/sphinx/util/inspect.py", line 547, in object_description
        s = repr(obj)
      File "/Users/bocklund1/src/pycalphad/pycalphad/core/workspace.py", line 217, in __repr__
        return repr(TypedField.__get__(self, obj))
                    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
      File "/Users/bocklund1/src/pycalphad/pycalphad/core/workspace.py", line 127, in __get__
        setattr(obj, self.private_name, default_value)
        ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AttributeError: 'NoneType' object has no attribute '_calc_opts' and no __dict__ for setting new attributes
```

Unclear if this a bug in PyCalphad or a bug in sphinx (seems to get triggered by `DictProxy` objects). Either way, it seems orthogonal and we can defer fixing docs builds for now.

nury12n (approved):
Since it's mainly just pyparsing API changes and the tests are passing, I don't see any issue here. I did a quick double check in `grammer.py` and `tdb.py` to make sure there wasn't any calls to the pyparsing API was overlooked and it looks like the changes made here cover everything.

Issue #230, opened 2019-07-17
DOC: Add FAQ item for failed calculations because temperature is outside of Piecewise expression


bocklund:

bocklund:
Closing because this is a duplicate of #229

PR #656, opened 2026-03-12
FIX: Cache StateVariable objects to mitigate memory leak

bocklund:
Checklist

As described in [ESPEI#230](https://github.com/PhasesResearchLab/ESPEI/issues/230), the PyCalphad `StateVariable` class subclasses SymEngine's `Symbol` class which causes a memory leak because of memory cycles that can't be garbage collected with the underlying SymEngine library. It's not clear whether that will be fixed in the near future. I proposed a mitigation for that to cache Symbol objects in our least recently used cache that is implemented here.

The current behavior is that whenever we create a `StateVariable` (a `Symbol` subclass) we create a new object. For example, `v.Y(x, y, z) is v.Y(x, y, z)` evaluates to `False` while `v.Y(x, y, z) == v.Y(x, y, z)` evaluates to `True`. creates a new object for `x`, `y`, and `z`. The idea here is to re-use objects rather than re-create them by caching on the arguments to `__new__` such that `v.Y(x, y, z) is v.Y(x, y, z)` evaluate to `True` by looking in the cache.


bocklund:
At this point, the fix is "well known" and I've tested it on and off when necessary over the years (ESPEI #230 is nearly 4 years old at this point), but I have always been unhappy with how much of a hack it feels like. I don't like the idea of committing to having a cache here as a design decision and the semantics of having instances of `StateVariable` objects becoming quasi-global variables via the cache. Specifically how that might break existing assumptions in subtle ways or causes users to write code in ways that use this as an assumption that we might break in the future if upstream is ever fixed. 

Nonetheless, a strength of PyCalphad is the ability to do high-throughput simulations, deploy on HPC, have long-running processes (like ESPEI) and that feels undermined by having to deal with memory issues.

I am not in a hurry to merge this and would be happy to take feedback from PyCalphad maintainers and users (cc @richardotis @nury12n).

bocklund:
@nury12n could you take a look at this PR? Ideally if we're merging it, I'd like to get it in before the Calphad meeting release

pycalphad/variables.py
```diff
@@ -375,6 +380,35 @@ def __getitem__(self, new_units: str) -> "StateVariable":
...
+        reduce_result = self.__reduce__()
+        _, args = reduce_result[0], reduce_result[1]
+
+        # Create new object bypassing the cache
+        newobj = wrapped_new(cls, *args)
+
+        # Run __init__ to set up instance attributes
+        newobj.__init__(*args)
+
+        # Copy any instance-level attribute overrides (like display_units)
```
nury12n, line 405: Aren't `display_units` and `implementation_units` implemented as class attributes in `StateVariable` and not instance attributes, so shouldn't they be shared across all instances of a class? If the intent is that they can change among instances, then why are they defined as class attributes? I say this also to refer to the unit tests also testing that these attributes are not shared when copied.

nury12n: Nevermind, class attributes only seemed to be shared if they only have operations applied to them (i.e. append, insert, etc.). But they can be overwritten (especially for immutable types), which seems to be the intended behavior here.

nury12n:
I'm okay with the changes here and running several downstream codes didn't lead to any issues as well. While it is a bit odd for the state variables to take on some sort of global status using the cache, I always felt like the typical usage of state variables treated them as such. I would think any assumptions that is broken with the cache would be minor edge cases. If we're really worried about possible breaking behavior either with this pull request or in a future pull request where the memory leak is solved without having to use the cache, I think we could add something to the documentation to clarify.

pycalphad/variables.py
bocklund: Yes, you're right. For future readers, `display_units` is a class variable, but `__getitem__` creates a copy of the object and sets `display_unit` as an instance variable on that copy, which takes priority over the class variable when Python follows the usual Method Resolution Order.

```python
>>> from pycalphad import variables as v
>>> t = v.T["celsius"]
>>> print(t.display_units, t.__class__.display_units)
celsius kelvin
```

the former is the instance variable and the latter the class variable

PR #657, opened 2026-03-29
Some changes in examples to contribute with issue #321

disouzam:
This pull request tries to close some items described in issue #321, namely:

- [x] Add v.N: 1 as a condition for all equilibrium calls and N=1 for all calculate calls.
- [x] Activity example: Use v.R instead of hardcoded 8.315
- [x] Activity example: change the title of the plot to "1023 K" to match the data

Besides that, some lines have trailing whitespace trimmed to keep files "cleaner".

I'm not sure about the current practice with Jupyter notebooks but, in the commits that I pushed here, I only changed the code but didn't commit changes in output. I can do that as well if necessary.

Issue #321, opened 2021-04-07
DOC: regenerate examples and small fixes


bocklund:
Examples could be regenerated and some low hanging fruit issues could be tackled (few lines of changes maximum for each item):

- [ ] Add `v.N: 1` as a condition for all `equilibrium` calls and `N=1` for all `calculate` calls.
- [ ] Make sure xarray output is using the newer rich HTML output, which is what users will see
- [ ] Binary phase diagram examples: when creating a figure in the notebook, use a higher DPI (maybe 150 or 200?) instead of increasing the `figsize`
- [ ] "Calculating Energy Surfaces of Binary Systems" - do the calculate step for all phases in one call and use xarray `calc_res.where(calc_res.Phase == phase_name)` syntax for masking in the loop
- [ ] "NP vs. T" - use the broadcast_like functionality of xarray, similar to the cookbook example for "NP vs. T" [here] 
- [ ] Prefer xarray indexing and broadcasting to numpy indexing wherever possible (see, for example the ["NP vs. T" cookbook example](https://github.com/pycalphad/cookbook/blob/master/PhaseCompositions/PhaseCompositions.ipynb))
- [ ] Change the reference state of the convex hull enthalpy of formation example to be at 298.15 K.
- [ ] Activity example: Use `v.R` instead of hardcoded 8.315
- [ ] Activity example: change the title of the plot to "1023 K" to match the data

Related: it would be good to be running the docs notebooks as part of our CI to catch any docs regressions. The website build could regenerate the docs on-the-fly in each build, but it probably isn't necessary.

disouzam:
Hello, @bocklund :

I'm new to xarray and I didn't search yet all the places where it is used in pycalphad. 

Looking at the second item of this issue (Make sure xarray output is using the newer rich HTML output, which is what users will see), do you mean the output shown in the picture below (taken from examples I'm following from PyData Prague 2020 talk using marimo's notebook)?:

<img width="1174" height="627" alt="Image" src="https://github.com/user-attachments/assets/ff94cbcd-d6c8-4e57-a904-81b37df92b6d" />

bocklund:
No, it's slightly different. For example, in the "Exploring `calculate` and `equilibrium` xarray Datasets" notebook, changing `print(calc_result)` to just `calc_result` in the first calculation cell gives this interactive HTML element 

[...]

PR #658, opened 2026-03-29
Some changes in examples to contribute with issue #321 - Part 2

disouzam:
This pull request is a continuation of #657 (I can merge them together if it makes easy for you to review). My goal here was to address, at least partially, the items listed below:

- [x] Binary phase diagram examples: when creating a figure in the notebook, use a higher DPI (maybe 150 or 200?) instead of increasing the figsize
    - Though it is not fully clear to me, I set a fixed dpi to all figures in example notebooks: there were 3 scenarios of figure creation: 1) Just figsize set and no dpi, 2) Just dpi set and no figsize and 3) Neither parameter was passed.
    - Set all dpi sizes to 300 (but can also set to the 200 or smaller dpi as 150). I decided to set to an even higher dpi size to have figures that could be inspect in greater details or even printed in large sizes.

I have also added two other commits that could help with the maintenance of example notebooks:
- In one of them (edae32e), I have added explicit parameter names in calls to methods. IMHO, this improves readability, specially for newbies like myself that needs some guidance to familiarize with the code base
- And in the other commit (e0db06c), I have removed empty cells from all example notebooks that had them.


bocklund:
Thanks for opening these! Feel free to keep these changes for these together. They seem easy enough to review at this stage. 

I'm generally okay with more consistent DPI and size settings. If it helps, I tend to think about sizes more for scaling plot area vs. text size and information density - smaller figure sizes have less tick marks and labels and tends to be easier to read in for presentation slides, for example. I find 150 dpi to be sufficient for most uses, but I'm open to whatever looks nice :)

On the other hand, I would suggest _not_ using explicit keyword arguments over positional arguments to methods like `calculate`, `np.unique`, `set_xlabel`, `set_ylabel`,  `set_title`, and so on. I think this is a bit of an anti-pattern and not something usually seen "in the wild". It could be confusing to someone used to seeing `np.unique()` used with positional arguments to see one with keyword arguments as an implication that the function is being used differently than what they might be used to.

disouzam:
Thanks, @bocklund !

I will merge two PRs then and make changes you suggested. 

Are you open to other changes in examples? For example, I was thinking on moving databases to a separate folder for better separation between notebooks and data. Another change I thought was to rename examples by putting a numbered prefix to follow the order they are presented in documentation website.

bocklund:
Some reorganization is okay with me if it helps make them easier to follow and contribute to. I don't really think of them as ordered, but we do tend to go through them in order when we do workshops, see here for example: https://github.com/materialsgenomefoundation/2025-calphad-workshop/tree/main/pycalphad

disouzam:
> Some reorganization is okay with me if it helps make them easier to follow and contribute to. I don't really think of them as ordered, but we do tend to go through them in order when we do workshops, see here for example: https://github.com/materialsgenomefoundation/2025-calphad-workshop/tree/main/pycalphad

It is exactly what I was thinking. More specifically, the list in pycalphad.org suggest a sequence (maybe randomly selected, I don't know) and it would be good (at least for me) to make it easier to follow (as files are sorted in a different order in VS Code, for example, and don't follow the exact same order).

image

In any case, I will make a proposal (it is not that difficult) and you decide later if proceed with it or abandon.

Regarding your comments, @bocklund , I will close this pull request and consolidate all changes in the first one (#657) .

PR #667, opened 2026-05-01
Qkto extrapolations

jorgepazsoldanpalma:
Checklist
Have added changes to PyCalphad to minimize the major parsing errors when attempting to open the MSD-TC database. The things that have been changed are the addition of a if statement that takes care of the negative value in QKTO phases when they exhibhit different extrapolation schemes. Additionally, I have added a term that handles the composition dependent reciprocal interaction parameter observed in both MSD-TC for fluorides and chlorides


bocklund (changes requested):
Thanks! 

- [x] Would you be able to add tests for new behavior?
- [ ] Can you cleanup/remove the trivial whitespace changes and commented out code/print statements?
- [ ] Please also remove the `requirements-dev.txt` as we use `pyproject.toml` to manage all dependencies now.

Could you also describe more what changes are happening here and why the existing code was wrong for QKTO, especially if you can point to a paper/textbook description. Not required to merge, the code, but it would be helpful to add a relevant comment in the Model or database parser, where appropriate, that points out any subtleties that were missing.

jorgepazsoldanpalma:
I believe I have addressed the issues. Please let me know if there is anything else you need from me!

jorgepazsoldanpalma:
I unfortunately could not find a manuscript that directly talks about what was missed in the parser aside from what I added in the test section in the current push. If you are fine with the equation I am referencing and think it is enough then I will also add it to the parser as well. If not I can try digging for more appropriate examples.

bocklund:
Thanks for the changes, I made a few minor cleanups. I'll take a look at the paper and see if I get it. If it's clear, I think this is good. 

`atol=10.0` compared to Thermochimica is really big. Is this a minimizer issue? I'm wondering if an energy test with given site fractions would be more appropriate to show that the model is implemented correctly and giving the same value as we expect.

jorgepazsoldanpalma:
Hi Brandon, 

That is my mistake. The actual value we are comparing to is from FactSage itself and it is more close to 4 J than 10. I just put 10 but feel free to lower the atol to 5.

bocklund:
No worries, I can tighten the tolerance and change the comment. The question is still the same though - typically I'd expect that we can reproduce the energy with the same site fractions to within like less than 1e-5 J. I'm trying to get at the separate issues of 1) whether the model is implemented correctly, and 2) whether the equilibrium solver gets the right answer

bocklund (approved):
I took a closer look and made a couple changes. I think understand what's going on now and refactored how you approached `model_mqmqa.py` to be more explicit about the reciprocal `A != B and  X != Y` case in line with the other types of mixing.

This PR looks good to me and I'm ready to merge it on my end. Would you mind looking at my changes to see if you agree with them @jorgepazsoldanpalma? Is there anything the reciprocal parameter is missing to be general to what is available in other software tools?

jorgepazsoldanpalma:
Hi Brandon, 
Everything looks great to me except the cycler import. Is that something new that you are adding?

bocklund:
Silly VS Code auto import, thanks for catching!

pycalphad/tests/test_equilibrium.py
```diff
@@ -5,6 +5,7 @@
 
 import warnings
 import os
+from cycler import V
```
bocklund:
```suggestion
```

PR #671, opened 2026-05-15
added zip equilibrium

daniel-sintef:
Hey Everyone,

Disclaimer, the code was written by AI, though I have been carefully testing & validating it. (This text of this PR is completely written by hand.) 

I have a use case where I want to do 'high throughput' CALPHAD computations over the Al-Cu-Mg-Zn-..... simplex for a grid of temperatures and compositions. My understanding is that the current 'equilibrium' code is not very well suited to this because it runs a cartesian product over compositions & temperatures. I had a coding agent create a version of equilibrium version that 'zips' the conditions before running. 

As far as I can tell the 'zip equilibrium' with a massive speed bump compared to the built in method. I am guessing it probably runs with some combinatorial factor. For example on a 4 element, fairly coarse grid I get a 22x speedup over the built in method. There are some deviations but they are, as far as I can tell, very small ~10^-12 - ~10^-7 in composition, so in line with numerical error. 

While I would hestitate to make a pull request using 'AI slop' the speedup is so large that I thought it would be worth it. Especially given that the error seems fairly small (and for many use cases, like mine quite acceptable). That said I am very happy to pitch in and try to run more validation. I have included a version of the validation script I have been using so far which will confirm the large speedup and low error. 

It would be great if someone could help me confirm this is fine. I wanted to merge this in more smoothly, but am finding the 'workspace' logic a bit hard to parse. 

Thanks!

Daniel 

Unit tests have been added, below is the script I used to 'manually' verify the results. You should run it in the 'examples' directory or adjust the tdb path

```
from itertools import product
from pathlib import Path
from time import perf_counter

import numpy as np
import pandas as pd

from pycalphad import Database, equilibrium, variables as v, zip_equilibrium


dbf = Database(Path("pycalphad/tests/databases/COST507.tdb"))
components = ["AL", "MG", "SI", "CU", "VA"]
elements = components[:-1]
solutes = elements[1:]
phases = list(dbf.phases.keys())
temperatures = [1000.0, 1250.0, 1500.0]
pressure = 101325.0
density = 4
eps = 1e-4
phase_amount_cutoff = 1e-8


def composition_grid(component_count, density):
    rows = []
    for counts in product(range(density + 1), repeat=component_count):
        if sum(counts) == density:
            row = np.asarray(counts, dtype=float) / density
            row = np.clip(row, eps, 1.0)
            rows.append(row / row.sum())
    return np.asarray(rows)


def phase_values(eq_result):
[...]

PR #672, opened 2026-05-16
Add a high-throughput calculation example

daniel-sintef:
This creates an example that illustrates how to efficiently compute a grid of equlibrium states, and bypasses the need for https://github.com/pycalphad/pycalphad/pull/671


bocklund (approved):
Thanks @daniel-sintef, this is great! I was thinking this would be a good example based when I was writing up my comment in #671. 

Since `Workspace` is the recommended way to do this, I simplified the example to only use `Workspace` so we don't confuse people (or LLMs) who are reading the docs later and trying to understand what the recommended pattern is. 

I also added to the grid example with a random sampling approach. Being able to perform calculations over arbitrary composition lists is a major strength of the `Workspace` approach that doesn't have an equivalent in `equilibrium`.

daniel-sintef:
Awesome! thanks so much for your help on this

Issue #662, opened 2026-04-01
TernaryStrategy.do_map appears to loop indefinitely for Al-Cu-Y at T > 2260 K


GeorgeZJT:
Hello pycalphad dev team, this is another follow up on issue #624.

For [this example](https://pycalphad.org/docs/latest/examples/TernaryExamples.html), when the temperature is set above 2260K, `TernaryStrategy.do_map()` appears to loop indefinitely, and no plot is produced.

Thermodynamically, the bug happens at higher temperature where only one solid phase plus liquid phase are stable.

Environment:
pycalphad 0.11.1 and the latest development version
windows 11
python 3.13.12

Code:
```
import numpy as np
import matplotlib.pyplot as plt
import pycalphad
from pycalphad import Database
from pycalphad import variables as v
from pycalphad.mapping import TernaryStrategy, plot_ternary
from packaging.version import parse

new_version = parse(pycalphad.__version__) >= parse("0.11.1")

temperature = 2260

db_al_cu_y = Database('Al_Cu_Y_assessment_files.tdb')
comps = ['AL', 'CU', 'Y', 'VA']
phases = list(db_al_cu_y.phases.keys())
conds = {v.T: temperature, v.P:101325, v.X('AL'): (0,1,0.015), v.X('Y'): (0,1,0.015)}
strat = TernaryStrategy(db_al_cu_y, comps, phases, conds)
# Add automatic starting points by searching along the axes
if new_version:
    strat.generate_automatic_starting_points()
else:
    strat.initialize()
# Add custom starting point near middle of phase diagram
strat.add_nodes_from_conditions({v.T: temperature, v.P: 101325, v.X('AL'): 0.33, v.X('Y'): 0.33})
strat.do_map()
plot_ternary(strat, x=v.X('AL'), y=v.X('Y'), label_nodes=True)
plt.title(f'Al-Cu-Y at T={temperature}K')
plt.savefig(f'T={temperature}')
plt.show()
```

bocklund:
@nury12n can you look into this?

PR #676, opened 2026-05-20
Fix: circular phase boundaries causing infinite loop in ternary mapping

nury12n:
This should address #662. This is a case where you have two-phase equilibrium between solution phase and an intermetallic (in #662, it is the liquid and AlCuY phase), where the liquid phase boundary forms a loop. Before, mapping would only stop once it finds a node or reached an axis limit, and thus would infinitely step around the phase boundary loop.

Changes
- Added check to see if a last point of the ZPF line approaches the first point and ends the ZPF line when this occurs
- Added `max_iter` argument in `Strategy.do_map()`. By default, this is set to -1, which would iterate the mapping code until all stopping conditions are met. But one can set it to a positive finite value which would stop the mapping code pre-maturely to help with debugging in the future.
- When plotting tielines, if the axis variable in x and y varies by a small amount (< 1e-3), then this will plot a point at the average composition. 
  - [ ] This might need some changing to allow the user to disable or set a different threshold if the user is mapping within a small composition range.

Example
```python
from pycalphad.mapping import TernaryStrategy, plot_ternary

dbf = Database('Al-Cu-Y.tdb')
comps = ['AL', 'CU', 'Y', 'VA']

temperature = 2260
conds = {v.T: temperature, v.P:101325, v.X('AL'): (0,1,0.015), v.X('Y'): (0,1,0.015)}
strat = TernaryStrategy(dbf, comps, phases, conds)

strat.add_nodes_from_conditions({v.T: temperature, v.P: 101325, v.X('AL'): 0.33, v.X('Y'): 0.33})
strat.do_map()
plot_ternary(strat, x=v.X('AL'), y=v.X('Y'), label_nodes=True)
plt.title(f'Al-Cu-Y at T={temperature}K')
plt.savefig(f'T={temperature}')
plt.show()
```

T=2260


pycalphad/mapping/plotting.py
```diff
@@ -155,6 +155,8 @@ def plot_tielines(ax, strategy: Union[BinaryStrategy, TernaryStrategy], x: v.Sta
             x, y, p = single_phase_data.x, single_phase_data.y, single_phase_data.phase
             if not all((single_phase_data.y == 0) | (single_phase_data.y == np.nan)):
                 ax.plot(single_phase_data.x, single_phase_data.y, color=phase_colors[p], lw=1, solid_capstyle="butt")
+            if label_end_points and (np.amax(x) - np.amin(x)) < 1e-3 and (np.amax(y) - np.amin(y)) < 1e-3:
```
bocklund: what do you think about having these be relative rather than absolute differences? I'm okay with hard coding them, but could also consider a module level constant or something so it's slightly configurable?

pycalphad/tests/test_mapping_strategy.py
```diff
@@ -579,3 +579,34 @@ def test_issue_638_degenerate_cs_binary(load_database):
...
+    comps = ['AL', 'CU', 'Y', 'VA']
+    phases = list(dbf.phases.keys())
+    conds = {v.T: T, v.P:101325, v.X('AL'): (0,1,0.015), v.X('Y'): (0,1,0.015)}
+    strat = TernaryStrategy(dbf, comps, phases, conds)
+    strat.add_nodes_from_conditions({v.T: T, v.P: 101325, v.X('AL'): 0.33, v.X('Y'): 0.33})
+
+    # this is pretty much the loop in strat.do_map()
+    # the conditions we set here should be 90 iterations
+    # so mapping does not finish by 1000 iterations, then something would be wrong
+    strat.do_map(200)
```
bocklund: 1000 or 200?

bocklund (approved):
Thanks, @nury12n!

PR #677, opened 2026-05-25
ENH: Use dynamic axis labels from StateVariable attributes in eqplot

adityamoolya:
The _plot_labels dict in eqplot.py was hardcoded to only T and P, so passing any other state variable (like N) to _axis_label() would raise a KeyError. The same issue existed in _map_coord_to_variable().

Since every StateVariable subclass already defines display_name and display_units, I replaced the hardcoded dict with dynamic label generation using those attributes. Also updated _map_coord_to_variable() to build its lookup table dynamically from pycalphad.variables instead of listing variables manually.

Other small additions:

MassFraction support in _axis_label() for consistency with MoleFraction
9 unit tests for both functions
Note: axis labels change from abbreviated units (K, Pa) to the display_units already on the classes (kelvin, pascal). Happy to adjust if a different format is preferred.

### DIFF
--- a/pycalphad/plot/binary/plot.py
+++ b/pycalphad/plot/binary/plot.py
@@ -6,7 +6,7 @@ as `equilibrium`.
 
 import pycalphad.variables as v
 import matplotlib.pyplot as plt
-from pycalphad.plot.eqplot import _axis_label
+from pycalphad.mapping.plotting import get_label
 from pycalphad.plot.utils import phase_legend
 from .map import map_binary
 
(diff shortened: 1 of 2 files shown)


bocklund:
Thank you for the PR! I'll be happy to review it.

I am curious -  how are you using `eqplot` today? I was just considering marking it as deprecated and scheduling it for removal since the mapping is so good now. Is there some code you need help migrating or a use case for `eqplot` that I'm overlooking?

FYI, I've also been working on a branch to get better unit support in mapping and hope to have that up this week.

adityamoolya:
> so good now. Is there some code you need help migrating or a use case for `eqplot` that I'm overlooking?

Thanks for the quick response!

I was reading through the source code to get familiar with pycalphad and came across the TODO,  since _axis_label is also used by binary/plot.py, it seemed worth fixing even beyond eqplot itself.

If eqplot is heading toward deprecation, happy to drop the _map_coord_to_variable changes and keep it scoped to just _axis_label. And if there's anything I can help with on the mapping unit support branch, I'd be glad to pick something up.

bocklund (changes requested):
Thanks for putting this together.

The `x`, `y`, and `z` arguments to `eqplot` are documented to be `StateVariable` objects, I think it could be sufficient to use the `pycalphad.mapping.plotting.get_label()` instead of `_axis_label` and `_map_coord_to_variable`. `_axis_label` is also used by `pycalphad.plot.binary.plot.plot_boundaries()` (and I think that should also work with `get_label()`).

`get_label()` also has tests associated already, so this could maybe be a net removal of code :)

adityamoolya:
Oh nice, I hadn't noticed get_label() in the mapping module, that already handles everything. I'll wire that up in eqplot.py and binary/plot.py and push the updated changes.

bocklund (approved):
Excellent @adityamoolya, thank you so much!

PR #678, opened 2026-05-25
ENH/FIX: Unit handling for mapping

bocklund:
Checklist

This PR:
1. Adds unit support to mapping by doing JIT conversion. Updates the charged phases example to show use of Celsius unit rather than doing the conversion by-hand for mapping.
2. We refactored the unit conversion API to pull the functionality out of `Workspace`, fixing a bug where the context contained all the phases, not just the active ones requested in the property.


bocklund:
~~Probably could use a short isopleth test and test temperatures / axis limits~~ Done

pycalphad/mapping/primitives.py
```diff
@@ -411,7 +429,7 @@ def get_var_list(self, var : v.StateVariable):
         Gets variable along ZPF line and returns list
         The variables will decipher between local and global variables
         """
-        return np.array([p.get_property(var) for p in self.points])
+        return np.array([to_display_units(p.get_property(var), p.stable_composition_sets, var) for p in self.points])
```
nury12n, line 432: It seems like `Point.get_property` will always return in SI units while `ZPFLine.get_var_list` return in display units. Do you think this could cause possible confusion in the future?

I don't think this should prevent merging since `ZPFLine.get_var_list` can of course return in SI units and the mapping code as it is right now assumes that `Point.get_property` returns in SI units. Perhaps, I'll just add a comment in the docstrings for these functions to state their expected behavior

pycalphad/tests/test_mapping_strategy.py
```diff
@@ -185,6 +189,7 @@ def test_unary_strategy(load_database):
     """
     Tests that strategy works on unary system
     The strategy needs to maintain certain array shapes for site fractions, composition,
+    The strategy needs to maintain certain array shapes for site fractions, composition,
```
nury12n: This line is repeated

pycalphad/mapping/strategy/strategy_data.py
```diff
@@ -98,8 +99,10 @@ def get_invariant_data_from_tieline_strategy(strategy, x: v.StateVariable, y: v.
             node_phases = node.stable_phases_with_multiplicity
             data = []
             for p in node_phases:
-                x_data = node.get_property(_get_phase_specific_variable(p, x))
-                y_data = node.get_property(_get_phase_specific_variable(p, y))
+                x_var = _get_phase_specific_variable(p, x)
```
nury12n, line 102: Want to confirm as a sanity check: In both the previous and current changes, the getting node data in the `IsoplethStrategy` will scale with the computed phase fraction on the node and getting invariant data in the tieline strategies will not. For the `IsoplethStrategy`, this makes sense since the mapping plane is slicing across the node traversing the plane will lead to different phase fractions across the node. Therefore, points on the node found during mapping will correspond to a 2 unique sets of phase fractions. For the tieline strategies, the node is in the mapping plane, so all three points on the node corresponds to where one of phase fractions is 1.

I'm not sure if I originally had a reason for not accounting for phase fraction in `get_invariant_data_from_tieline_strategy` or if I overlooked it by assuming all properties passed here would not depend on phase fraction (I can't really think of a case where one would need to plot a property that depends on phase fraction in both the isopleth and tieline strategies). But I think it would be worth adding the `_composition_sets_with_phase_fractions` here and set the phase fraction to 1 for each phase when iterating through the node phases since that would what each point on the node represent

pycalphad/mapping/strategy/strategy_data.py
nury12n, line 144: Similar concern here where each phase boundary on a `ZPFLine` in the tieline strategies represent where the phase fraction of said phase is 1, although during mapping, one is fixed to 0.

Although, for the `IsoplethStrategy`, this becomes tricky since the phase boundary of a `ZPFLine` is where the phase fraction of the fixed phase is 0. So if one were to plot a property that depended on the phase fraction, it could result in 0 across the entire `ZPFLine`

pycalphad/mapping/primitives.py
bocklund: I thought about pushing this down to `Point.get_property`, but most usage of that method in the code are in mapping internals that assume implementation units. Like, if we put a `to_display_units` flag int `Point.get_property` it's also going to be just these functions. I feel like that's YAGNI right now - we can always add it, but can't take it away.

pycalphad/mapping/strategy/strategy_data.py
bocklund: Hmm. I'm struggling to think of a valid use case for any computed quantity scaled by the phase fraction in an isopleth.



Reminded me of this diagram as a use case. For an isopleth, something where the phase fractions are scaled to one, in Hillert's terminology, would not be a true phase diagram because of the overlapping phase fields. For example, this is from a binary and a slightly different

image

nury12n (approved):

pycalphad/mapping/primitives.py
```diff
@@ -402,6 +417,9 @@ def num_free_phases(self):
     def append(self, point: Point):
         self.points.append(point)
 
+    def __repr__(self):
```
nury12n, line 420: `__repr__` and `__str__` seems to not be covered by the unit tests. I think this can be added to `test_mapping_strategy.py::test_primitive_representation`

PR #680, opened 2026-05-29
DOC: Add example for custom mapping plots with shaded regions 

nury12n:
Added to the Mapping API example to show how the data in `BinaryStrategy` and `TernaryStrategy` can be used to customize plotting behavior.


bocklund:
Failing test will be fixed by #681

bocklund (approved):
Awesome! Thank you @nury12n

PR #679, opened 2026-05-25
FIX: Fix constraint violation in Hit & Run sampler

bocklund:
Issue #675 exposed a case where the Hit & Run sampler returned extra points that violated a sublattice mass balance constraint. The root cause is that it was using `np.linalg.pinv` that gave a solution outside the box, which was then clipped inside the box in a way that violated the constraint.

Here we switch to `scipy.optimize.lst_linear` which allows passing lower/upper bounds and gives solutions, if available, inside the box. If solutions are not found, it returns `None` which raises - raising should be the correct behavior because anything that gets to this point in the H&R sampler should be able to be made feasible.

Isolated benchmarking shows that using `lst_linear` is about a 2-3x performance loss over `pinv` (20 µs -> 50 µs). With caching of points, this should be relatively inconsequential and better than using `scipy.optimize.linprog` which has about 450 µs (calls to `sample` are about 1500 µs currently, so this is a relevant amount).

Includes a test isolated from #675.

Issue #116, opened 2017-08-03
Calculate gives the wrong result if pressure and temperature are not both specified.


bocklund:
In the binary example, we see the following code produce this plot 

``` python
from pycalphad import calculate
from pycalphad.plot.utils import phase_legend
import numpy as np

legend_handles, colorlist = phase_legend(my_phases_nbre)

fig = plt.figure(figsize=(9,6))
ax = fig.gca()
for name in my_phases_nbre:
    result = calculate(db_nbre, ['NB', 'RE'], name, T=2800, output='GM')
    ax.scatter(result.X.sel(component='RE'), result.GM,
               marker='.', s=5, color=colorlist[name.upper()])
ax.set_xlim((0, 1))
ax.legend(handles=legend_handles, loc='center left', bbox_to_anchor=(1, 0.6))
plt.show()
```

![image](https://user-images.githubusercontent.com/7681751/28926636-642598d2-7836-11e7-9e6a-d434cae7abbd.png)

But changing the temperature to 300 or 1000000 gives the same plot (and the same values from calculate). If you specify pressure in calculate e.g. `result = calculate(db_nbre, ['NB', 'RE'], name, T=2800, P=101325, output='GM')` the resulting plot is
![image](https://user-images.githubusercontent.com/7681751/28926687-8666f7a6-7836-11e7-925f-739b60c8274a.png)

Further changes to temperature produce expected results. The energy surface plotted when P is not specified gives a similar result to T>=6000 (piecewise max in TDB). I assume this is a bug and not the intended result?

amjokisaari:
Please make this throw an error if something isn't specified, instead of silently calculating the wrong thing, as suggested by @richardotis.

richardotis:
In https://github.com/pycalphad/pycalphad/blob/develop/pycalphad/core/calculate.py#L428
The assigned values to `statevar_dict` need to be wrapped with a call to `unpack_condition()`
That should fix it.

richardotis:
That does not fix the issue. Still investigating.

PR #681, opened 2026-05-29
FIX: raise `ConditionError` in `calculate()` when there are missing state variables

bocklund:
Checklist

In `calculate()` state variables are passed as keyword arguments. It's currently possible to call `calculate()` without providing all the state variables that are used by `Model` instances, and this creates mismatches in the expected vs. actual shapes of arguements passed to the phase record callable functions. 

This PR fixes the flaky test introduced by #679 that exhibited this behavior by not including `T` as a state variable argument. It also includes a check (with a corresponding test) that the state variables passed to `calculate()` are a superset of those required and recorded by the `PhaseRecordFactory` and raise if any required variables are not present. 

There's also an interesting historical note that #116 was basically asking for this check to exist, and at the time we provided default T and P, but those defaults got removed by #200 and the introduced in the fix for #116 didn't include a set of conditions that exercised this issue.


nury12n (approved):
LGTM

Issue #695, opened 2026-06-16
Add support for extensive amount of moles conditions, `N(i)` support


bocklund:
- Drop hard requirement and default `N = 1` conditions. Generalize to non-unity `N` conditions
- Allow for extensive amount of components to be mixed with other conditions
- Implement like we implement the linear combination of conditions for future `B`, `B(i)` support.

PR #699, opened 2026-06-18
ENH: Expanded conditions support and Components support

### LINKED ISSUE #692: General support for non-pure element component basis with `v.Component`
Be able to provide a non-pure element component basis to `Workspace`, e.g. `Workspace(..., components=["AL2O3", "SIO2", "O2"])`. Conditions are required to be in the same component basis. The property framework can either be in the passed component basis or in the pure element basis.

Specific things that can be broken out to new issues if needs that are blocked by this one if the work needs to be deferred:
- [ ] Ensure Components in Jansson derivatives work
- [ ] Components can be specified for any set of conditions or outputs that take a component.

bocklund:
**This is a draft PR.** 

This PR:
- Removes the requirement to have a total `N` condition and only support for `N=1`. This is a breaking change because we remove the default `N=1` condition. Any legacy code that did not explicitly include `v.N: 1` conditions will break. 
  - Consequently, `v.N` is removed as a "state variable" in the sense of being built into phase records and composition sets. `N` is never an input to a `Model` callable and only current potentials `P`, `T` and any new user-defined state variables that are part of `Model` objects make sense to be in there. We always derive `N` from the moles of elements computed by callables.
- Adds support for `N(i)` conditions
- Adds support to take in a set of `v.Component` objects (or strings that parse into components) to define a new component basis. Similar to the `DEFINE_COMPONENTS` feature found in other software.

The new component basis is intended to be completely general for the conditions and the property framework.

This would close #692 and #695.

--- 

I realized after some discussions, re-reading how we do linear combination conditions and drafting up `N(i)` conditions (which we implement a sum of all elements) that user-defined components are just a change of basis in that requires a linear combination of pure element conditions in the same way that we do explicit linear combination conditions, but general to `v.W`, `v.B`, `v.X(Component)` (and the `W(Component)`, `B(Component)`, `N(Component)` variants), the explicit linear combination conditions of those, as well as `MU(Component)`. 

We're actually not far from supporting that and I threw an LLM at the last part, which introduced the ability to provide a set of `v.Component` objects to `Workspace` and the functions it calls that construct the new components basis and store it in the `PhaseRecordFactory`. I am not completely happy with the implementation yet, but I think the implementation is correct at first glance and the external API changes (i.e. the only change being that we now support providing non-pure element `v.Component` objects to `components` arguments) should be stable such that we can test this branch as I evaluate the code.

Before merging, I'd like to
- Rewrite/simplify the implementation of how we track and use the basis
- As a stretch goal, I'd like to unify better how we handle explicit linear combination conditions and generalize those beyond the current `v.X` conditions. Maybe they get implemented as a condition-local change of basis?

Adding support for more conditions greatly increases the different ways parts of the code can be composed and cause breakages. I think this will need a good deal of manual testing to try to uncover unforeseen edge and corner cases. I'm particularly concerned about how support for more types of conditions affects starting point selection.

Extending Component support to `mapping` is not in scope for this PR, but I expect that we can work on it on top of this branch since the external-facing API changes relevant to mapping should be stable with respect to the implementation details that may change. 

I expect this to be a key feature for pseudobinary and pseudoternary support. Although as these changes are simply a linear combination of conditions, it will not resolve issues with convergence in the solver when the solutions are in subspaces of the component space. The thermodynamically correct thing to do is to set the chemical potentials of one of the components, but observations in issues elsewhere indicate that may be flaky.

### DIFF
--- a/pycalphad/codegen/phase_record_factory.py
+++ b/pycalphad/codegen/phase_record_factory.py
@@ -4,6 +4,7 @@ from pycalphad.core.utils import get_pure_elements, unpack_species, \
     extract_parameters, get_state_variables
 from pycalphad.core.phase_rec import PhaseRecord
 from pycalphad.core.constraints import build_constraints
+from pycalphad.core.errors import ConditionError
 from itertools import repeat
 from functools import lru_cache
 import numpy as np
@@ -14,6 +15,7 @@ class PhaseRecordFactory(object):
         self.pure_elements = get_pure_elements(dbf, comps)
         self.nonvacant_elements = sorted([x for x in self.pure_elements if x != 'VA'])
         self.molar_masses = np.array([dbf.refstates[x]['mass'] for x in self.nonvacant_elements], dtype='float')
+        self._build_component_basis(dbf, comps)
         parameters = parameters if parameters is not None else {}
         self.models = models
         self.state_variables = sorted(get_state_variables(models=models, conds=state_variables), key=str)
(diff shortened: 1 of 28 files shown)


bocklund:
Closing because this will get superseded elsewhere. This work continues in https://github.com/bocklund/pycalphad/commits/define-components/

PR #700, opened 2026-06-19
ENH: MQMQA Bragg-Williams (B/H) excess mixing parameters (closes #403)

### LINKED ISSUE #403: MQMQA support: `B`-type excess mixing parameters
MQMQA models in the DAT format can have mixing code `Q`, `G`, or `B` that correspond to different excess equations. Right now `Q` and `G` are implemented in `ModelMQMQA.excess_energy`, but mixing code `B` is not yet supported.

gabrielbusta:
Closes #403.

## Plain-language description

Adds the Bragg-Williams excess mixing codes `B` and `H` to
`ModelMQMQA.excess_mixing_energy`. `G`/`Q` were already supported and `R` was added in
#667; `B`/`H` previously raised `ValueError`.

For cation mixing the term is

```
g_ex = L · X*_{A/X}^(1+p) · X*_{B/X}^(1+q) / (X*_{A/X} + X*_{B/X})^(1+p+q)
```

with the anion-mixing analogue. `B` and `H` are identical, and the contribution is `g`
with no quadruplet wrapper, matching Thermochimica's SUBG routine
([CompExcessGibbsEnergySUBG.f90 L586-L632](https://github.com/ORNL-CEES/thermochimica/blob/0c35c8d7d1cf2084b4e2ca5d6608f7dcdf60adad/src/gem/CompExcessGibbsEnergySUBG.f90#L586-L632)).

Validated against compiled Thermochimica on one-line `G -> B` edits of two existing test
fixtures (added as `*-Bvariant.dat`): the cation arm (`MQMQA-tern-tests`) and anion arm
(`Shishin` slag) match to better than 5e-6 and 2e-7. Unit tests also pin the term's value,
`B == H`, and the `L=0` limit.

A `B`/`H` parameter that carries an additional third-constituent term raises
`NotImplementedError` (Thermochimica applies no such factor and we had no validated form);
ordinary multicomponent systems are unaffected. This is a starting point; happy to revise
if you have a preferred formulation or that term already in progress.

## Checklist
- [x] Plain language description of the changes.
- [x] No dependency changes.


bocklund:
Thanks for these. Do you know what is the difference between `B` and `H`? Do you have a paper/book reference that has either (or, preferably, both) of these cases written out? 

I don't think it's necessarily a blocker for this PR, but I do wonder if ternary mixing is theoretically defined for these models, but just not in Thermochimica. 

Since the changes in copied databases are rather small, I'd like to see our tests make the changes programmatically, rather than making copies of the database files that are mostly the same in the test suite.

I also think the tests in `test_energy.py` could be removed entirely and replaced with a test like `test_MQMQA_bragg_williams_anion_oracle_vs_thermochimica` that is doing a code-to-code comparison. Rolling our own implementation in `test_energy` to test against our own implementation in `model_mqmqa` doesn't seem that valuable of a test for correctness as much a test for regressions against the test implementation. 

In general, `check_energy`-type tests are preferred as those isolate the `Model` implementation from global minimization issues

bocklund:
This comment in #403 says B and H terms should use Y (site equivalent fractions) https://github.com/pycalphad/pycalphad/issues/403#issuecomment-1126043363. I think that answers my previous question. Wondering if the X* terms used in this implementation are the same for the issue tested or if the docs for FactSage might be wrong (as suggested by the linked comment). We need to resolve this before it can be merged. Testing against FactSage would be ideal.

PR #703, opened 2026-06-29
ENH: Add read/write support of ThermML database format

flotang-gtt:
The ThermML database format in Version 0.1 (current) is a superset of
TDB databases and ChemSage.dat databases, including but not limited to
metadata and phase models.

Supported phases:
* CEF structure: sublattices, site constituents, per-phase species
* Endmember, Interactions
* IHJ magnetism.
* Order/disorder models
* MQM / SUBQ: cation/anion split, chemical groups, endmembers,
  quadruplets, binary interactions, pair fraction expansions, asymmetric
  ternary interactions, reciprocal pair fractions
* ternary non-Muggiano extrapolations

Implementation Gaps
* Xiong magnetic model (untested, expressable in ThermML)
* Wildcard support as in https://doi.org/10.1016/j.calphad.2025.102849
  (pseudo-binary implementation in higher order systems)
* EBEF, 4SL formalism
* MQ,MF,DQ,DF,VS,VO,VA,VC TDB commands for volume, compressibility,
  thermal expansion, diffusion, mobility, kinetics are not
  aligned/reconciled with the MolarVolume, ThermalExpansion,
  Compressibility, BulkModulusDerivative etc from ThermML
* VISC, ELRS, THCD, SIGM, XI property model extensions
* Two-State liquid model
* 3rd generation unary parameters (GD, THETA, NT)
* ThermML richer metadata is not carried through pycalphad
* Explicit ternary interpolation overrides other than Muggianu are not
  supported by pycalphad. Warnings are issued for those files.
* MQMQ reciprocal excess parameters, asymmetric ternary interactions are
  currently not suppoted by pycalphad.

Testing / Validation
* TDB >> ThermML >> pycalphad >> TDB roundtrips provably exact for CEF +
  magnetic + order/disorder systems.
* ChemSage.dat >> ThermML >> pycalphad >> ChemSage.dat roundtrips for
  quasichemical phases that do not hit pycalphad implementation gaps.
* A fair amount of warnings on lossy conversions/ingestion problems is
  present.


bocklund:
High level thoughts before a detailed read through:

1. As there are now multiple `.xml` variants out there now, this conflicts with and will break if people currently use something like pycalphad-xml via the plugin system. We probably need a slightly more advanced registry mechanism that will try to match on both file extension and metadata. 
2. How stable do you consider ThermML? Since our release cadence is only every ~6 months or so, I wonder if it would be better to distribute this as a plugin like pycalphad-xml. If we also resolve (1), then I wouldn't be opposed to adding pycalphad-thermml as a dependency to pycalphad so users get it automatically and by default. That would allow the ThermML reader and write to be maintained and updated separately from the PyCalphad release cycle.

flotang-gtt:
Good points. I hesitated on this PR initially due to similar concerns around ThermML's stability, but we have stabilized most, if not all of it.

We are already releasing supporting software (e.g., Calphad Optimizer) on a roughly annual release cycle - some other internal tools have higher format churn, but this should not be a big issue. Future schema expansions such as crystallographic metadata, viscosity parameterizations, or other physical properties should be structured as additions to a static "core" schema, meaning they won't break or impact this component. ThermML is just about to hit the biggest common denominator of the roundtrippable thermodynamic databases.

As for file extensions and metadata matching, ThermML includes enough data to handle this. We are still deciding on a dedicated .thermml extension internally, but since it's standard XML, runtime introspection is necessary for proper parsing anyway and makes tooling a tiny bit simpler.

bocklund:
I've been thinking more about this. I want this to be available to users, but I haven't come up with a system that I want to commit to that de-conflicts the different possible `.xml` extension formats and gives a friendly experience.

Right now, I think this would require having special handing for parsing XML, resolving XML formatting bugs, then dispatching on to a registered format based on the declared schema. 

@flotang-gtt if you're supportive, I'd be happy to create and a http://github.com/pycalphad/pycalphad-thermml plugin repo that can be maintained under the pycalphad org, cherry-pick these commits, hook up PyPI publishing, etc.. I can add you and any others as maintainers to that repo as well.

flotang-gtt:
OK, thanks for your thoughts on this.

There are indeed many options of how this can be integrated.The pycalphad-thermml plugin concept sounds absolutely right, and I would be happy to maintain/help out with that.

Regarding the issue of two concurrent XML schemata: Wouldn't reasonably ThermML superseed the existing variant? I was under the impression that that format was mostly used as an intermediary, somewhat temporary solution and does not have longterm backing. I may be wrong, obviously, but still, resolving into a single (ThermML based) schema sounds like much desirable: From the XML side the required transformations of a file are quite small and easily bridged. Maybe there is more dependent infra that I am not aware of?

To make it nice in terms of UX and DX, I suggest to let pycalphad-thermml provide the necessary branching of XML reading code - for the most part, I would make sure that ingestion of the legacy XML format is possible, whereas emitted XML would be confined to ThermML. pycalphad-xml would then be sunset, or resolved/merged into pycalphad-thermml. Migratory issues would only arise if downstream consumers are specifically targetting the legacy xml code - do you know if this?  Then, the simple input file routing can stay in pycalphad and overall complexity does not rise unnecessarily.

Happy to fill me in if my picture is overly simple, or get to work on the legacy-xml transformation rules.

pycalphad/tests/test_thermml.py
```diff
@@ -0,0 +1,893 @@
...
+    assert dbf.refstates["SN"]["H298"] == pytest.approx(1.0)
+    assert dbf.refstates["SN"]["S298"] == pytest.approx(2.0)
+
+
+def test_thermml_extension_autodetect():
+    "A .xml filename is auto-detected as the ThermML format."
+    path = files(pycalphad.tests.databases).joinpath("Pb-Sn.xml")
+    dbf = Database(str(path))  # no explicit fmt=
+    assert "LIQUID" in dbf.phases
+
```
bocklund:
```suggestion
```

Redundant with `@select_database("Pb-Sn.xml")` tests

pycalphad/io/thermml.py
```diff
@@ -0,0 +1,1500 @@
...
+
+# Volume-model properties. TODO how these map/align to TDB-based V0, ...
+_VOLUME_PROPERTY_TYPES = frozenset(
+    {
+        "MolarVolume",
+        "ThermalExpansion",
+        "Compressibility",
+        "BulkModulusDerivative",
+    }
+)
```
bocklund: `Model` objects currently support volume model types by Lu, Selleby, Sundman:

```
    'V0',     # Molar volume at STP
    'VA',     # Integrated thermal expansivity
    'VC',     # High-pressure fitting parameter
    'VK',     # Isothermal compressibility
```

`Database` objects themselves allow `parameter_type` of each parameter entry be anything (whether or not those mean anything to any particular model). The boundary is typically at serialization, e.g. TDBs don't read unknown parameter types.

pycalphad/io/thermml.py
```diff
@@ -0,0 +1,1500 @@
...
+# Reuse the ChemSage-DAT species renaming so MQMQA species names match the .dat reader.
+from pycalphad.io.cs_dat import rename_element_charge
+
+
+# ThermML XML namespace. The schema major version is encoded in the namespace
+# (the ``.../thermml/v0`` segment), so a major bump implies a new namespace.
+TM_NS = "http://calphad.org/thermml/v0"
+XSI_NS = "http://www.w3.org/2001/XMLSchema-instance"
+_TM = "{%s}" % TM_NS
+_XSI_TYPE = "{%s}type" % XSI_NS
```
bocklund:
```suggestion
_TM = f"{{{TM_NS}}}"
_XSI_TYPE = f"{{{XSI_NS}}}type"
```

pycalphad/io/thermml.py
```diff
@@ -0,0 +1,1500 @@
...
+        "BulkModulusDerivative",
+    }
+)
+
+
+# ---------------------------------------------------------------------------
+# Small XML helpers
+# ---------------------------------------------------------------------------
+
+
```
bocklund:
```suggestion
```

Issue #690, opened 2026-06-15
Equilibrium calculation fails for SOLID phase in the Al2O3-Fe2O3 subsystem of Fe-Al-O


TaarygooSan:
# [Bug] Equilibrium calculation fails for SOLID phase in the Al2O3-Fe2O3 subsystem of Fe-Al-O

## Problem Description

I am encountering convergence failures when calculating the energy curve of the SOLID phase in the Fe-Al-O system. Specifically, the equilibrium solver returns `nan` for certain compositions in the Al2O3-Fe2O3 subsystem.

In contrast, the same approach works perfectly for the Fe2O3-FeO and FeO-Al2O3 subsystems, producing complete energy curves.

## Energy Curves of Other Subsystems (for reference)

**Fe2O3-FeO system — SOLID phase energy curve (works fine):**

<img width="690" height="490" alt="Image" src="https://github.com/user-attachments/assets/1929617a-f3e6-49df-ad37-b68baff59ad4" />

**FeO-Al2O3 system — SOLID phase energy curve (works fine):**

<img width="689" height="490" alt="Image" src="https://github.com/user-attachments/assets/b4b3197f-3166-4869-81d4-8ac8449a4ce3" />

**Fe2O3-Al2O3 system — SOLID phase energy curve (fails at certain compositions):**

<img width="690" height="490" alt="Image" src="https://github.com/user-attachments/assets/47843bd6-2cec-4ae7-abd0-3700b61c3dc0" />

As shown above, the calculation breaks at intermediate compositions where the equilibrium result becomes `nan`.

## Database Used

```xml
<?xml version="1.0"?>
<?xml-model href="../../../.venv/Lib/site-packages/pycalphad_xml/database.rng" schematypens="http://relaxng.org/ns/structure/1.0" type="application/xml"?>
<Database version="0">
    <ChemicalElement id="FE" mass="55.845" reference_phase="BLANK" H298="0.0" S298="0.0" />
    <ChemicalElement id="AL" mass="26.982" reference_phase="BLANK" H298="0.0" S298="0.0" />
    <ChemicalElement id="O" mass="15.999" reference_phase="BLANK" H298="0.0" S298="0.0" />
    <Species id="FEO">
        <ChemicalElement refid="FE" ratio="1" />
        <ChemicalElement refid="O" ratio="1" />
    </Species>
    <Species id="FEO1.5">
        <ChemicalElement refid="FE" ratio="1" />
[...]

PR #710, opened 2026-07-26
FIX: Fix charged phases adding linearly dependent constraints

bocklund:
Checklist

Phase internal constraints, such as charge balance, add additional rows to the phase matrix in the minimizer. In some cases, where phases are charge balanced without any charge degrees of freedom (e.g. stoichiometric (Ti+2)(O-2) or (Ca+2,Mg+2)(O-2)), the charge constraint is a trivial multiple of the sublattice site fraction sum constraint and the phase matrix becomes linearly dependent and produces garbage data when inverted.

This causes convergence issues and divide by zero issues as the Newton steps in the minimizer get polluted. This basically prevented charged point calculations from working (about 5% error rate in some of my testing on dense grids in a binary) and prevent charged phase diagrams from plotting correctly in most examples on https://phasediagrams.org.

This PR:

1. Adds explict tests that charge constraints are not added when a phase has no charge degrees of freedom _and_ is charge neutral (no DOF, charge infeasible phases are still linearly dependent, but are added to ensure the starting point code properly filters out the infeasible points).
2. Adds an equilibrium test that a simple two phase stoichiometric charge case converges. These conditions failed for me consistently, but might be machine/architecture-dependent.
3. Adds corresponding `Model` code to handle the charge constraint case and makes the tests pass.
4. Adds `tool.uv` `cache-keys` so that `uv sync`, `uv run pytest`, etc. rebuild the code when Cython or the build script (`setup.py`) changes so users don't need to run `python setup.py build_ext --inplace` if they are using uv.
5. Fixes a bug introduced in the last release where mapping plotting helpers don't catch the `variable.phase_name = "*"` condition and produced labels of `X(*,O)`, etc. for mole and weight fraction variables.

It is possible that other phase internal constraints devised by users could lead to linearly dependent phase matrix rows. For now we do not handle those cases. It is possible that we could detect poorly conditioned phase matrices in the minimizer and switch to a pseudo-inverse type method that could be more stable as a more general workaround, but pseudo-inverse methods are slower and may be less exact than solving the problem at the root. I'm considering more general solutions out of scope.

### DIFF
--- a/pycalphad/tests/test_equilibrium.py
+++ b/pycalphad/tests/test_equilibrium.py
@@ -1131,3 +1131,24 @@ def test_MQMQA_reciprocal_interaction_parameter_Dixon(load_database):
     print('Y', eq.Y.values.squeeze())
     print('Phase', eq.Phase.values.squeeze())
     assert_allclose(eq.GM.values.squeeze(), -229356.0, atol=5)  # FactSage result
+
+@pytest.mark.solver
+@select_database("TiO-17Yan.tdb")
+def test_charged_stoichiometric_phases_converge(load_database):
+    """Two-phase equilibrium between stoichiometric charged compounds converges (gh-712)"""
+    # The charge balance constraint of a stoichiometric ionic compound is a linear
+    # combination of its site fraction balance constraints. Adding it made the phase
+    # matrix singular, so the minimizer inverted it into garbage and never converged.
+    # The charged phases here are identical to the uncharged TI3O2NC/TIO_ALPHANC pair
+    # except for the charges of their constituents, so both must give the same answer.
+    dbf = load_database()
+    comps = ["TI", "O", "VA"]
+    conds = {v.P: 1e5, v.N: 1, v.T: 620.0, v.X("O"): 0.41}
+
+    eq = equilibrium(dbf, comps, ["TI3O2", "TIO_ALPHA"], conds)
+
+    assert np.all(np.isfinite(eq.GM.values))
+    assert_allclose(eq.GM.values, -241451.22198839)
+    stable_amounts = eq.NP.values.squeeze()
+    assert_allclose(np.sort(stable_amounts[~np.isnan(stable_amounts)]), [0.1, 0.9], atol=1e-8)
+    assert_allclose(eq.MU.values.squeeze(), [-486633.66127366, -71070.20485795])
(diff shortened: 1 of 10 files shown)


bocklund:
Some before/after phase diagrams:

| phasediagrams.org | this PR |
|--------|--------|
|  | image> |
|   | image |
|  | image | 

Cu-O and Dy-O are perfect now. Ti-O has some issues near the transition to gas phase, but I think that is likely a mapping-related issue that is orthogonal to these changes. A dense grid of stepping in `X(O)=0.01` and `T=20` produce no points that fail to converge or raise exceptions.

bocklund:
~~Before merging, I need to double check that this also fixes https://github.com/pycalphad/pycalphad/issues/690. In that case, I would expect the constraint would be kept, but the solution (given the conditions) could also produce linearly dependent phase matrices in the limit of zero FeO site fractions. I'm not sure if minimum site fraction limits are enough (and I assume if there are convergence issues that they are not). Potential solution in that case is to detect rank-deficient phase matrix and fall back to a pseudo-inverse that is slower, but won't blow up.~~

that issue doesn't actually use charged phases

pycalphad/tests/test_equilibrium.py
```diff
@@ -1131,3 +1131,24 @@ def test_MQMQA_reciprocal_interaction_parameter_Dixon(load_database):
...
     assert_allclose(eq.GM.values.squeeze(), -229356.0, atol=5)  # FactSage result
+
+@pytest.mark.solver
+@select_database("TiO-17Yan.tdb")
+def test_charged_stoichiometric_phases_converge(load_database):
+    """Two-phase equilibrium between stoichiometric charged compounds converges (gh-712)"""
+    # The charge balance constraint of a stoichiometric ionic compound is a linear
+    # combination of its site fraction balance constraints. Adding it made the phase
+    # matrix singular, so the minimizer inverted it into garbage and never converged.
+    # The charged phases here are identical to the uncharged TI3O2NC/TIO_ALPHANC pair
```
nury12n: I don't get this last part of the comment. Is there a test with TI3O2NC/TIO_ALPHANC that this test is supposed to be comparing against? I couldn't find any database in the test suite that defined these two phases.

bocklund: I had manually added "non-charged" (NC) versions of these phases when i was initially testing to try to isolate and verify that it that the charge constraint was the root cause and this slipped in

nury12n (commented):
model changes and solver options look good on my end

PR #715, opened 2026-08-15
FIX: add moles Hessian for ionic liquid special case, tune phase amount convergence criteria and fix site fraction convergence criteria

### LINKED ISSUE #652: How to expose max_iterations, fixes convergence issues for our "larger" system
I've been working on a cement database with @wahab2604 (using ESPEI to fit) but we've always had convergence issues. 
The system is fairly big with several sections in the CaO-Al2O3-SiO2-MgO-Fe-O system modeled including solid solutions and melt, and some stochiometric phases outside of that section. So far we've relied on Pandat to fit these sections.

However, we've noticed we can greatly improve our successful convergence by increasing the numbers on these lines:
https://github.com/pycalphad/pycalphad/blob/f374077e0209bbcf4610498e821657672d245263/pycalphad/core/hyperplane.pyx#L208
https://github.com/pycalphad/pycalphad/blob/f374077e0209bbcf4610498e821657672d245263/pycalphad/core/solver.py#L168
I think there's something to be done about the convergence criterion here for bigger systems, but for now I just want to expose these parameters so I can play with them at runtime to see what matters (note changing pdens does not help).

I tried to expose the `max_iterations` parameter by threading it up the calling functions but when I get to `recompute()` inside `Workspace` I'm stuck figuring out how best to expose it. The call I want to pass it into is here:
https://github.com/pycalphad/pycalphad/blob/f374077e0209bbcf4610498e821657672d245263/pycalphad/core/workspace.py#L362

I thought about following the same way that `"pdens"` is exposed via a options dict, but I cannot add `"max_iterations"` to `grid_opts` as it gets spread into `calculate()` here:
https://github.com/pycalphad/pycalphad/blob/f374077e0209bbcf4610498e821657672d245263/pycalphad/core/workspace.py#L361
So I'd need to either
* modify calculate to accept the parameter, but its only used in `starting_point()` so I don't want to start changing unrelated functions. 
* Filter it out of the grid_opts before passing it to calculate
* Provide the parameter separately to the constructor of the Workspace.

How do you want these magic constants passed up/down? I will probably try to expose any I find while I play with convergence.

### LINKED ISSUE #704: ZeroDivisionError in write_row_fixed_mole_fraction when total system amount transiently reaches 0 (fixed mole-fraction equilibrium on an ionic-liquid database)
Hello,
Thanks for pycalphad, i love thermodynamics.
I found a potential bug in the program, see issue below.
Regards,
Ruud

## Summary

`equilibrium` with fixed mole-fraction conditions raises an uncaught
`ZeroDivisionError` for a CaO-MgO-Al2O3-SiO2 (CMAS) two-sublattice ionic-liquid
database. The crash is in `write_row_fixed_mole_fraction`
(`pycalphad/core/minimizer.pyx`): the rows there divide by `current_system_amount`,
which transiently becomes `0.0` during the solve, and only `prefactor == 0.0` is
guarded — not `current_system_amount == 0.0`.

This also breaks any downstream consumer that computes a fixed-composition
equilibrium — e.g. ESPEI's `estimate_hyperplane` (ZPF error) hits it on every
tie-line, so ZPF-based liquid re-assessment of such a database can't run.

## Minimal reproducer

Uses a CMAS ionic-liquid TDB (attached: `cao_mgo_al2o3_sio2.tdb`; the 2-sublattice
ionic liquid `(AL+3,CA+2,MG+2)(ALO2-1,O-2,SIO4-4,VA,SIO2)` with ~33 phases).

```python

db = Database("cao_mgo_al2o3_sio2.tdb")
comps = ["AL", "CA", "MG", "SI", "O", "VA"]
phases = list(db.phases.keys())

# near-anorthite composition; T at the (expected) liquidus
conds = {v.T: 1826.15, v.P: 101325, v.N: 1,
         v.X('AL'): 0.15385, v.X('CA'): 0.07693,
         v.X('MG'): 1e-6,    v.X('SI'): 0.15384}

equilibrium(db, comps, phases, conds)   # -> ZeroDivisionError
```

Reproduces with `X(MG) = 1e-6` and with `X(MG) = 0.0`. (Not composition-universal:
some fixed-composition points on this database solve fine, others crash — consistent
with a transient degenerate state during the iteration rather than a bad input.)

## Expected behaviour

Either converge, or fail gracefully (non-convergence / `NaN` result), as happens
for other fixed-composition points — not raise an uncaught `ZeroDivisionError`.

## Actual behaviour (traceback tail)

```
  File "pycalphad/core/eqsolver.pyx", line 242, in _solve_eq_at_conditions
  File "pycalphad/core/solver.py", line 168, in solve
    converged = spec.run_loop(state, 1000)
  File "pycalphad/core/minimizer.pyx", line 838, in solve_state
  File "pycalphad/core/minimizer.pyx", line 689, in construct_equilibrium_system
  File "pycalphad/core/minimizer.pyx", line 268, in fill_equilibrium_system
  File "pycalphad/core/minimizer.pyx", line 145, in write_row_fixed_mole_fraction
ZeroDivisionError: float division
```

## Root cause

`write_row_fixed_mole_fraction` (`minimizer.pyx`) divides by `current_system_amount`
in several rows, e.g.:

```cython
out_row[...] += prefactor * (phase_amt[idx]/current_system_amount) * ...   # line ~145
out_row[...] += prefactor * (1./current_system_amount) * ...               # line ~156
```

The function guards `if prefactor == 0.0: return` but not `current_system_amount`.
When the total system amount transiently reaches `0.0` during `run_loop`, these
divisions raise. (Because `current_system_amount` is C `double`, the divide raises
`ZeroDivisionError` rather than yielding `inf`.)

## Suggested fix

Add the analogous early-return guard at the top of the function (mirrors the
existing `prefactor == 0.0` guard):

```cython
if prefactor == 0.0:
    return
if current_system_amount == 0.0:
    return
```

(Or clamp / skip the affected rows.) A quick check on whether the sibling
`write_row_*` helpers need the same guard would be worthwhile.

## Environment

- pycalphad 0.11.2
- numpy 1.26.4, scipy 1.16.3
- Python 3.11.11
- macOS (arm64)

## Notes

- Discovered while running ESPEI ZPF (`estimate_hyperplane` builds a fixed-mole-
  fraction `Workspace` equilibrium); the crash is entirely inside pycalphad, as the
  standalone reproducer above shows.
- Happy to attach the TDB and a full log.

bocklund:
Checklist

When working with a metal-only flavor of a 2 sublattice ionic liquid model, I was getting some spurious convergence failures and divide by zero errors in the minimizer. Convergence failures were caused by an oscillation of site fractions near convergence. This PR:

1. Fixes ionic liquid convergence by introducing the moles Hessian into the phase matrix 
2. loosens `ALLOWED_DELTA_PHASE_AMT` from 1e-10 to 1e-8 (fixes #652)
3. fixes a convergence correctness bug where only the last composition set was used in the site fraction convergence check

I expect this should also close #704, but I need the database to reproduce it.

Overall, I would expect to see a small performance bump from loosening the `ALLOWED_DELTA_PHASE_AMT`. The moles Hessian is relatively cheap to evaluate (compared to energy-based functions) because the expressions are very simple (especially the derivatives).

### Minimizer details

The root cause of this unstable behavior is that the moles of an ionic liquid are non-linear in site fractions due to the $P$ and $Q$ variables introduced. Currently PyCalphad's minimizer follows Sundman, Lu, and Ohtani, *Computational Materials Science* 101 (2015) 127-137 [doi: 10.1016/j.commatsci.2015.01.029](http://doi.org/10.1016/j.commatsci.2015.01.029). In that paper, Eq.37 defines the Lagriangian for a single phase with respect to site fractions:
$$\frac{\partial L}{\partial y^\alpha_{is}}  = \aleph^\alpha \frac{\partial G^\alpha_M}{\partial y^\alpha_{is}}  	- \aleph^\alpha \sum_{A} \mu_A \frac{\partial M^\alpha_A}{\partial y^\alpha_{is}}	-\eta^\alpha_s 	+ \lambda^\alpha \frac{\partial Q^\alpha}{\partial y^\alpha_{is}}$$

Since $\partial G^\alpha_M$ is non-linear, it is linearized by a first order Taylor expansion
$$\frac{\partial G^\alpha_M}{\partial y^\alpha_{is}} =  \left( \frac{\partial G^\alpha_M}{\partial y^\alpha_{is}} \right)	+ \left(\frac{\partial^2 G^\alpha_M}{\partial y^\alpha_{is} \partial T}\right) dT	+ \left(\frac{\partial^2 G^\alpha_M}{\partial y^\alpha_{is} \partial P}\right) dP	+ \sum_t \sum_j \left(\frac{\partial^2 G^\alpha_M}{\partial y^\alpha_{is} \partial y^\alpha_{jt}}\right) dy^\alpha_{jt}$$
Other terms are treated as linear. 

However, the two sublattice ionic liquid model is a special case because the introduction of variable site ratios for the cation ($P$) and anion ($Q$) sublattices introduces a non-linearity in the moles.
We should similarly linearize it with a first order Taylor expansion
$$\frac{\partial M^\alpha_A}{\partial y^\alpha_{is}} = 	\left( \frac{\partial M^\alpha_A}{\partial y^\alpha_{is}} \right)	+ \left(\frac{\partial^2 M^\alpha_A}{\partial y^\alpha_{is} \partial T}\right) dT	+ \left(\frac{\partial^2 M^\alpha_A}{\partial y^\alpha_{is} \partial P}\right) dP	+ \sum_t \sum_j \left(\frac{\partial^2 M^\alpha_A}{\partial y^\alpha_{is} \partial y^\alpha_{jt}}\right) dy^\alpha_{jt}$$
For regular CEF-style models (including the MQMQA) the moles are linear in site fractions and the second derivatives become zero. 

If we insert these back into the Lagrangian, convert to finite differences, remove the charge constraint because the 2 sublattice ionic liquids handles it implicitly by the $P$ and $Q$ auxiliary variables, and note that $\frac{\partial^2 M^\alpha_A}{\partial y^\alpha_{is} \partial T} = 0$ and $\frac{\partial^2 M^\alpha_A}{\partial y^\alpha_{is} \partial P} = 0$ (the code is general and handles this if any moles in the future are not linear in T and P), then we get (analogous to Sundman Eq. 39)

$$\sum_t \sum_j \left(\frac{\partial^2 G^\alpha_M}{\partial y^\alpha_{is} \partial y^\alpha_{jt}}		- \sum_{A} \mu_A \frac{\partial^2 M^\alpha_A}{\partial y^\alpha_{is} \partial y^\alpha_{jt}} \right) \Delta y^\alpha_{jt}-\frac{\eta^\alpha_s}{\aleph^\alpha} = \sum_{A} \mu_A 	\left( \frac{\partial M^\alpha_A}{\partial y^\alpha_{is}} \right)-\frac{\partial G^\alpha_M}{\partial y^\alpha_{is}}-\left(\frac{\partial^2 G^\alpha_M}{\partial y^\alpha_{is} \partial T}\right) \Delta T-\left(\frac{\partial^2 G^\alpha_M}{\partial y^\alpha_{is} \partial P}\right) \Delta P$$

This does introduce a dependence on chemical potentials in the phase matrix and we use the exiting ones from the previous iteration. A few tests are included that failed before these fixes. 

Fun fact that this functionality used to be in the code, but got removed in #360 with the justification that there was no mass Hessian contribution in Sundman's paper. Hopefully the comment and test are a little stickier this time :)

### Loosen `ALLOWED_DELTA_PHASE_AMT` tolerance

I was investigating #652 simultaneously. As  I mentioned there, when some of the phases have very small amounts of some of the species, e.g. nitrogen. The phase matrices to be poorly conditioned and there's some inherent noise in the numerics when we invert the phase matrix. This noise is pushes up against the convergence tolerances in the minimizer. I tuned the ALLOWED_DELTA_PHASE_AMT in the solver from 1e-10 to be 1e-8 and it makes the clinker database provided in #652 reliably converge in ~33 iterations. I added that to the test suite. **I will note that the middle set of conditions does require a higher pdens to converge correctly.** Hopefully we can eliminate this at some point in the future. But I think it's a good stress test to keep in the test suite for global min and starting point issues if/when we can fix this later.

### State variable convergence criteria fix

This is a smaller change. The `state.largest_y_change` convergence criteria was getting reset every composition set, such that only the last composition set would be included in the convergence criteria. This could create some scenarios where this convergence criteria falsely passes. 

There is a pre-existing implied decision (that is not changed here) that site fraction convergence of metastable composition sets are included in the convergence criteria. This is the only convergence criteria that is influenced by metastable composition sets. This is the most correct, but could be sub-optimal for performance if the minimizer needs more iterations to ensure that metastable phases are properly converged. On the other hand, this also could prevent us from converging to a local minimum if tightly converging a metastable composition set would make it stable. For comparison, OpenCalphad seems to only consider the driving force of metastable composition sets in its convergence criteria. 

At first, I did not have a specific test that verified this issue, however when I made the `ALLOWED_DELTA_PHASE_AMT` change, the `test_eq_charge_alfeo` test started failing. The Al-Fe-O test case   was added because it's sensitive to careful convergence. It turns out that the tight phase amount tolerance was masking an issue where the `state.largest_y_change` was not well converged in the entire system because it was being controlled by only the GAS phase (the last composition set). When we loosened `ALLOWED_DELTA_PHASE_AMT`, the site fractions were not fully converged in the stable phases and caused the test to fail. Fixing `state.largest_y_change` to correctly apply to all phases fixes the convergence.


bocklund:
Tests here on Windows look a bit flaky and are fixed by #716. Going to close this and let that one supersede this one.

PR #716, opened 2026-08-18
Refactor convergence criteria

bocklund:
Checklist
* [x] Include a plain language description of your changes.
* [x] Any dependency changes are reflected in the `pyproject.toml`.

This PR closes #641, #652, and #704. It contains the following changes:

1. Fixes ionic liquid convergence by introducing the moles Hessian into the phase matrix (see detailed notes in #715)
2. loosens `ALLOWED_DELTA_PHASE_AMT` from 1e-10 to 1e-8 (again, see notes in #715)
3. fixes a convergence correctness bug where only the last composition set was used in the site fraction convergence check and only allows stable phases to contribute to `largest_y_change` 
4. replace the sum-based implementation of the mass-residual to be per component condition (with better handling for linear combination conditions)
5. fixes a bug in the `solver.py` with bad assumptions about the shape of the conditions that would create zero-size arrays if there were no mole conditions

---

On investigating #641, I found that the main reason we weren't in agreement was early exits from Scheil simulation due to convergence failures. 
- I tracked one convergence failure down to being due to some components being dilute, which caused the mass residual tolerance (which was implemented as a sum) to be too small for good convergence. With some LLM assistance, I was able to code up something that treats the mass residual of all components in a more general way that avoids issues with using a sum to compute the residual and is more general for cases like weight fraction and linear combination of conditions (for which more changes are coming down the pipes). Those changes fixed `test_eq_issue641_trace_composition_conditions_converge`, but left `test_eq_issue641_unstable_compset_jitter_does_not_block_convergence` unfixed. At this point, the Scheil simulations were progressing a little further, but still below 90% solidified.
- The second fix was to only let `largest_y_change` be contributed to only by phases in the active set (non-zero amount or fixed). This fix alone wouldn't have fixed both tests, but in combination with the first set of changes, does resolve both tests.
- This also fixes a bug in the `solver.py` with bad assumptions about the shape of the conditions.

As noted in #641, we get much better convergence in the dilute limits, better than Thermo-Calc.

I also asked an LLM to instrument the test suite to track solver iterations to convergence and track cases where the solver fails to converge given the current test suite. In general, ionic liquid is small improvements to those special cases, and we get further improvement here, only a regression in `test_ternary_strategy`. 

| test | develop | fix-ionic-liquid | refactor-convergence |
|---|---:|---:|---:|
| `test_binplot[Al-Mg_Zhong.tdb]` | 8 | 8 | 8 |
[...]

PR #719, opened 2026-08-30
FIX: Fix charged sampling species order

bocklund:
Checklist

This PR:
1. Fixes an issue with charged phases where `Species` should be sorted by their escaped names in `calculate` to match the sorting of `SiteFraction` variables. This was affecting relative sorting of containing vacancies and charged vanadium, e.g. `'V+3' < 'VA'` but `'VA' < 'V_POS3'`. This bug affects all charged phases (ionic liquids and standard charged phases)
2. Fixes some solver bugs that led to divide by zero because floating point issues sometimes caused negative step sizes and because single phase composition sets could have had zero amount after removal and cause a divide by zero. Invariants about non-negative step sizes and minimum phase amounts are now explicitly enforced. 
3. Handle an edge case to remove pure vacancy phases in `filter_phases`. Also removes phases that cannot charge balance.


bocklund:
the changes made here to to the phase amount pinning seemed to break the new test in CI from #716 that was passing locally. https://github.com/pycalphad/pycalphad/pull/719/commits/1633007b03ed8389ca424e71bc157a4312687090 makes some changes to just remove composition sets that want to become unstable in `advance_state`. Not sure if I like this yet, but seems to fix the issue in CI. Running larger set of regression tests and will report back.

bocklund:
the `advance_state` composition set removal is a regression on some tests locally. reverting that and marking the badly behaving test as xfail so we keep it around

pycalphad/core/calculate.py
```diff
@@ -103,7 +102,10 @@ def _sample_phase_constitution(model, sampler, fixed_grid, pdens, phase_local_co
             constant_site_ratios = False
     species_charge = []
     for sublattice in range(len(model.constituents)):
-        for species in sorted(model.constituents[sublattice]):
+        # Points columns are ordered following model.site_fractions, which sorts by the
+        # string of the SiteFraction, that uses species escaped_name sorting.
+        # This may differ from species.name sorting, particularly when species are charged.
+        for species in sorted(model.constituents[sublattice], key=lambda s: s.escaped_name):
```
nury12n: is there any reason we cannot change `__lt__` in `Species` to use `escaped_name` instead of `name`? Are there other parts of the code where we need to sort `Species` solely by their `name` attribute?

bocklund: Yeah, I think we can. Should prevent us from making this mistake elsewhere (e.g. in ESPEI / kawin). I don't think there's anywhere in the code that would need name sorting rather than escaped_name sorting.

I went ahead and made that change and made anywhere that uses species sorting to just use bare `sorted` rather than `key=str` or `key=lambda s: s.escaped_name`.

bocklund: Nice catch

pycalphad/tests/test_calculate.py
nury12n: I'm a little confuse with `SPINEL` defining 4 sublattices but only defining constituents for the first 3 sublattices. The code still goes through `can_charge_balance` and correctly filters assuming the phase has only 3 sublattices. But I'm not sure if it's intended behavior for a database or model to load correctly when the phase and constituents definition don't match.

nury12n (approved):

pycalphad/tests/test_calculate.py
bocklund: Yeah, somehow it doesn't break things. Parameter filters probably get weird or things could be numerically incorrect. I raise a warning in Model now and fixed two instances.

PR #720, opened 2026-09-02
ENH: Read and write the DIFFUSION and ZEROVOLUME_SPECIES TDB commands

sorkins:
The TDB grammar already recognized both commands but discarded them. They are now kept, so a kinetic database round-trips and downstream code can read ferromagnetic diffusion coefficients.

- `DIFFUSION` is parsed with a pyparsing grammar covering the four models of the DICTRA extensions (`NONE`, `DILUTE`, `SIMPLE`, `MAGNETIC`) and stored per phase as `Phase.model_hints['diffusion']`, a `DiffusionModel` named tuple: the model, the `ALPHA` and `ALPHA2` coefficients as `(species or None, value)` pairs (`None` for the bare form that applies to every species), and the constituent array of the `DILUTE` and `SIMPLE` forms. `alpha_for(species)` and `alpha2_for(species)` resolve a species-specific value over a bare one. A plain tuple keeps the phase hashable and picklable; `Model` ignores the key like any other hint it does not consume.
- `DIFFUSION` lines precede the `PHASE` lines in Thermo-Calc's files, so the parsed models are queued and attached once every phase is read, as `TYPE_DEFINITION` is. A command naming a phase the file never defines is dropped, since mobility databases are often distributed as an appendix to a thermodynamic database and carry lines for phases defined only there. A command outside the grammar is skipped with a warning rather than failing the parse.
- `Database.zerovolume_species` holds the names listed by `ZEROVOLUME_SPECIES`. The command is file-level and its usual argument is `VA`, hence a set of names on the `Database` rather than a `Species` or refdata attribute. `__setstate__` supplies the default so a `Database` pickled by an older pycalphad still unpickles complete.
- `write_tdb` emits `ZEROVOLUME_SPECIES` and one `DIFFUSION` line per phase carrying the hint, after the `TYPE_DEFINITION` block and before the first `PHASE`.

Tests cover all four models, coefficient resolution for a named and a bare species, the phase and species names the `PHASE` and `CONSTITUENT` commands accept, write/read and pickle round-trips, `Model` construction with the hint present, the undefined-phase and malformed-command cases, and the shipped databases: crfe_bcc_magnetic.tdb gains `ZEROVOLUME_SPECIES VA !` and `DIFFUSION MAGNETIC BCC_A2 ALPHA=0.3 !` (inert for the energies it tests), and diffusion.tdb's commented-out `ZEROVOLUME_SPECIES VA !` is enabled.

Checklist


sorkins:
@bocklund can you approve CI workflows to run? Happy to fix any issues that happen in the actual CI process.

bocklund (commented):
I am not super familiar with these, so I looked at the DICTRA help in the database module on Thermo-Calc console mode:

```
 DICTRA EXTENSIONS
  
        DICTRA Extensions to the Database Definition File Syntax
        ********************************************************
  
 Software packages for simulation of diffusional phase transformations,
 such as DICTRA, needs both thermodynamic data and kinetic data (i.e.,
 diffusivities or mobilities). Naturally, the handling and storage of
 kinetic data will also benefit from the use of some kind of database
 management. Thus, the TDB database definition file syntax has been
 extended to incorporate some new keywords needed for storing kinetic
 data.
  
  
 =======
 PARAMETER [special GES parameter name] [low temp. limit]
    [expression 1]; [upper temp. limit 1] Y
    [expression 2]; [upper temp. limit 2] Y
    [expression 3]; [upper temp. limit 2] Y
       .......... ; ..... Y
    [expression n-1]; [upper temp. limit n-1] Y
    [expression n]; [upper temp. limit n] N {Ref. Index} !
  
 The keyword PARAMETER allows to enter all types of normal GES parameters
 for thermodynamic data (as previously presented in Section 6.3.15), as
 well as five special extensions suitable for kinetic data used in the
 DICTRA software. Valid extensions to special GES parameter names are:
  MQ ==> Activation enthalpy for mobility equation.
  MF==> Pre-exponential factor for mobility equation.
  DQ==> Activation enthalpy for diffusivity equation.
  DF==> Pre-exponential factor for diffusivity equation.
  VS==> Volume counted per mole of volume carrying species.
  
 Examples:
  PARAMETER MQ(BCC,FE:VA) 298.15 1000+200*T+...; 6000 N !
  PARAMETER MF(BCC,CO:VA) 298.15 1000+200*T+...; 6000 N !
  PARAMETER DQ(FCC,FE:VA) 298.15 1043+...; 6000 N 10 !
  PARAMETER DF(FCC:CR:C ) 298.15 1000+200*T+...; 6000 N 10 !
  PARAMETER VS(FCC) 298.15 1000+200*T+...; 6000 N 11 !
  
  
 =======
 DIFFUSION [model keyword] [phase name] [additional parameter(s)] !
  
 The keyword DIFFUSION specifies what type of diffusion model to use for
 a phase if the default model is not desired. The default model calculates
 the full diffusion matrix. A diffusivity is calculated from the different
 mobilities and the thermodynamic factors. The former ones are calculated
 as:
  M = exp [SUM(MF/RT)] exp [SUM[MQ/RT)] / RT
 where SUM stands for a weighted summation of the different MF's and MQ's
 plus possibly a Redlich-Kister term. Valid model keyword are:
  NONE       no diffusion in this phase
  DILUTE     constitution list of dependent species in each sublattice
               must be given as an additional parameter. Only the diagonal
               terms in the diffusion matrix are calculated.
                 D = exp [SUM(DF/RT)] exp [SUM(DQ/RT)]
  SIMPLE     constitution list of dependent species in each sublattice
               must be given as additional parameter. Only the diagonal
               terms in the diffusion matrix are calculated.
                 D = SUM(DF) + SUM(DQ)
  MAGNETIC   the so-called ALPHA and ALPHA2 parameters must be given as
               additional parameters. ALPHA is for the substitutional
               magnetic model and ALPHA2 for the interstitial one.
               By appending an "&" (the and sign) and a species name
               after the alpha keyword one can supply individual values
               for the different species. The full diffusion matrix is
               calculated.
  
 Examples:
  DIFFUSION NONE SIGMA !
  DIFFUSION DILUTE CEMENTITE : FE : C : !
  DIFFUSION MAGNETIC BCC_A2 ALPHA=0.3 ALPHA2&C=1.8 ALPHA2&N=0.6 !
  
  
 =======
 ZERO_VOLUME_SPECIES [list of species] !
  
 In the DICTRA software, the assumption that the volume is carried by
 the substitutional elements only is applied. The interstitial elements
 are assumed to have zero molar volumes. The keyword ZERO_VOLUME_SPECIES
 uses a list of which species are to be considered as zero volume ones
 for an argument.
  
 Example:
  ZERO_VOLUME_SPECIES VA C N !
```

I'm not sure yet how I feel about putting these as additional Database attributes. 
1. If we take the definition of `DIFFUSION` commands here as exhaustive, I think we're safe to write a pyparsing grammar. These look to be phase-based, so I lean towards making adding them as additional metadata on `Phase` objects. 
2. Maybe zero volume species might go as some additional attribute on Species (if actually allowed on species) in the refdata (if allowed only on pure elements)? I have an old DICTRA PDF manual that talks about a breaking change reducing the support for species in preference of elements, but I'm not sure how up to date that is

What do you think? Also @nury12n feel free to add comments.

sorkins:
@bocklund Definitely makes sense to me as a phase-specific attribute, I've attempted an initial refactor to that, let me know if anything doesn't look like what you meant. I'm also not well-versed in the zero volume stuff, and our internal expert is on PTO for a while so can't ask right now unfortunately. Would be happy to fix any issues / rework in the future as needed.

sorkins:
@bocklund @nury12n quick reminder if any more thoughts on this I could be addressing/thinking on.

bocklund:
I think the main thing for me is understanding what the use cases are before we change the API surface (e.g. adding attributes/methods) to stable core objects like `Database`. 

1. I'm good with approach of the `DIFFUSION` because there's no new API surface since we can re-use `model_hints`. 
2. I don't quite understand the use case of `ZERO_VOLUME_SPECIES`. Since expands the API surface, before we merge, I want to understand 
    a. whether commercial codes currently make use of this (do we _need_ roundtrip/writer support?). how does it show up in the models/equations?
    b. how does this differ from a `VS = 0` parameter in implementation?
    c. if we do need explicit `ZERO_VOLUME_SPECIES` are they actually corresponding to `SPECIES` or to `ELEMENT`? Do they belong as a separate collection on `Database` objects (as implemented) or attributes on `Species` objects (or an attribute on (currently non-existent) `Element` objects).
3. It's also worth a pass through to clean up some spurious comments and stuff, I can do this as I review.

sorkins:
I looked into this further and have dropped ZEROVOLUME_SPECIES since nothing in pycalphad would consume it, there's no reason to parse the data for now. Whichever algorithm wants it, kawin for instance, is the one better placed to implement the usage and to define the representation it needs, the way PRISMA and DICTRA have defined their use of zero volume species. So looks like we can defer how we would ideally want this kind of data held in pycalphad.

PR #723, opened 2026-09-17
ENH: Improve Database parsing coverage 

bocklund:
Checklist

This PR:
1. Adds parsing support for `VB` volume parameters. These are not currently used in any models implementations.
2. Adds support for multi-temperature Einstein parameters (`LNTHETA1`/`THETAF1`, ..., `LNTHETA5`/`THETAF5`). Only indices 1-5 are supported in commercial softwares, so we don't accept parsing anything other than those in the parser
    - Add model support for these as well. Models support arbitrary integer-indexed LNTHETA/THETAF parameters.
    - Einstein energies using LNTHETA parameters are per mole of atoms and so we need to multiply by the number of atoms to get the per mole formula. Legacy `THETA` parameters give contributions of per mole of formula (incorrectly with the physics, confirmed with Thermo-Calc), which is why legacy databases using THETA in the wild give sublattice site ratios that sum to unity.
3. Add more general support for lexing all `TDB_PHASE_DESCRIPTIONS`. As a policy: we explicitly handle known `AMEND_PHASE_DESCRIPTION` keywords, ignore keywords that don't matter, and raise for keywords that are known, but not supported yet pointing people to open issues if they need the features. As part of this we
    - Change the if statements to handle only one keyword per `_process_typedef` call
    - Support properly expanded names of `TDB_PHASE_DESCRIPTIONS`, rather than hard-coded cases in `_process_typedef`
    - Create an explicit ignore list containing `COMPOSITION_SETS`, `MAJOR_CONSTITUENT`, `FRACTION_LIMITS` which are hints that are not used by PyCalphad
    - Add support for `STATUS_BITS 02200800`, which is the legacy version of `NEVER_DISORDER`


pycalphad/model.py
```diff
@@ -1184,28 +1184,52 @@ def twostate_energy(self, dbe):
...
+
+        # LNTHETA parameters are per-mole-atoms that we convert to per-mole formula
+        num_atoms = self._site_ratio_normalization
+        # we can't preform arbitrary RK sum because
+        # 1. the normal missing parameter convention of theta_i = 0 is invalid
+        # 2. each LNTHETA parameter is matched to a THETAF parameter
+        # so we dynamically accept any integer-indexed LNTHETA parameter
+        lntheta_query = (
+                (where('phase_name') == phase.name) & \
+                (where('parameter_type').matches("LNTHETA[0-9]+")) & \
```
nury12n, line 1223: I know in general cases, `LNTHETA` and `THETAF` would be limited to 1-5 from the tdb parser, but would `LNTHETA[0-9]+` allow for a user to add a term other than 1-5 to a `Database` object (outside of parsing a tdb file) and have it still work in `Model`? Is that intended behavior?

bocklund: Yeah, I wanted to restrict database I/O to what is commercially supported, but I didn't see a reason to restrict `Model` if people wanted to experiment with more temperatures

nury12n (approved):

